PR #5, opened 2017-09-01
[WIP] Adds GUI dialogue when user tries to open file in sd-journalist

joshuathayer:
Opening this as an example of how to open QT Dialogues. In this case, we alert the user when they try to open a file in the sd-journalist VM (which, I believe, should only be used to download submissions from the SD Journalist interface, not for general-use computering).

### DIFF
--- a/dom0/sd-journalist-files.sls
+++ b/dom0/sd-journalist-files.sls
@@ -23,9 +23,9 @@
     - group: root
     - mode: 755
 
-/usr/local/bin/do-not-open:
+/usr/local/bin/do-not-open-here:
   file.managed:
-    - source: salt://sd/sd-journalist/do-not-open
+    - source: salt://sd/sd-journalist/do-not-open-here
     - user: root
     - group: root
     - mode: 755
--- a/sd-journalist/do-not-open-here
+++ b/sd-journalist/do-not-open-here
@@ -1,11 +1,16 @@
 #!/usr/bin/python
 
 import sys
-from PyQt4 import Qt
+from PyQt4 import QtGui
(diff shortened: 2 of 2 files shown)


joshuathayer:
Just so I remember for later, this suggests another path for KDE mimeapps.list: https://bbs.archlinux.org/viewtopic.php?id=202636

redshiftzero:
Hmm, so I symlinked the existing `mimeapps.list` in the `sd-journalist` VM for Dolphin:

`ln -s ~/.config/mimeapps.list ~/.local/share/applications/mimeapps.list`

but Dolphin isn't respecting this. There is a file containing  `/usr/share/applications/mimeinfo.cache` that has the same filetypes as accessed via `keditfiletype` - these are _not_ honoring the `mimeapps.list` and I suspect is where Dolphin is looking. Not sure how `/usr/share/applications/mimeinfo.cache` is getting populated. 

Gonna try just removing Dolphin and installing Nautilus via salt instead since this has been a bit of a rabbit hole and we don't even like Dolphin 🙃 ...

redshiftzero:
Writing my approach out so someone can report any obvious issues: In the branch `salt-config-customize-sd-journalist`, I'm attempting to do this package config via salt. Since this customization of software packages needs to be done in the template, I used `qvm-clone` to make a new template (template `sd-journalist-whonix-ws`), then set `sd-journalist-whonix-ws` as the template for the `sd-journalist` AppVM, and finally added the custom configuration in `sd-journalist-template.sls` (to install Nautilus and uninstall Dolphin).

Issue #3, opened 2017-09-01
Wonky dir names in SVS Sources dir


joshuathayer:
The source directory structure seems to include a needless root directory based on the name of the file that's been downloaded from SD. I *think* we can just discard another dir level and files will end up in their properly named source directories.

redshiftzero:
Ah, this is probably not an issue in the SecureDrop workstation and is instead likely the issue on the SecureDrop server side I just filed here:  https://github.com/freedomofpress/securedrop/issues/2262

To verify this is the case, if there is only a wonky directory name when you download _all submissions from a single source_ (i.e. check one box on the journalist interface as described in the STR in https://github.com/freedomofpress/securedrop/issues/2262), then it's the server side issue. If so, I would probably avoid making customizations on the workstation side for the wonky folder structure and instead unpack what the server generates, as the customization would need to be removed here after the issue is fixed upstream.

redshiftzero:
Good news, [the server-side issue](https://github.com/freedomofpress/securedrop/issues/2262) was [fixed](https://github.com/freedomofpress/securedrop/pull/2336) by @kushaldas - closing this! If you provision your staging VMs again on latest develop, this extra folder will no longer be there.

PR #6, opened 2017-09-01
Discard another directory when saving decrypted submissions

joshuathayer:
This should address #3. From manual testing in many download permutations from the SD Journalist interface this seems to drop files into source-named directories in the `~/Sources` dir in `sd-svs`.


kushaldas:
Tested, and the PR worked as suggested. :+1:  Though I can see merge conflict, which needs to be fixed first.

redshiftzero:
Hey @kushaldas - can you restamp this one for merge?

Issue #2, opened 2017-09-01
Small submissions fail to process


joshuathayer:
Very short messages seem to fail to process through to the SVS Sources dir.

After some investigation, this seems to be an issue with the tarball that gets sent from the `dispVM` to `sd-svs`.

redshiftzero:
Possibly dupe of #8?

redshiftzero:
I just tested this a bunch on latest `master` using both small documents and messages and wasn't able to trigger any errors, so I'm closing this :innocent:

PR #7, opened 2017-09-01
[WIP] Make sure to close tars before using them

joshuathayer:
This hasn't been end-to-end tested yet, but should address bug #2. We were not properly closing the python Tar object after creating it on the disp VM, which in certain instances seemed to lead to some submissions not making to to `sd-svs`. This takes care of that bug, plus adds proper closing to other places where we use python's `tarfile`. I'll update this once it's been tested on a freshly built installation.

### DIFF
--- a/decrypt/decrypt-sd-submission
+++ b/decrypt/decrypt-sd-submission
@@ -18,7 +18,7 @@ tmpdir = tempfile.mkdtemp()
 # we get given a tarball of everything the user has downloaded.
 # first step, extract that archive
 with tarfile.open(input) as tar:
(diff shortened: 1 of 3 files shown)


redshiftzero:
On `master` I was unable to get submissions appearing in `sd-svs` but a fresh rebuild on the PR fixes. I tested the "all-" zip file flow (downloaded by logging into the Journalist Interface, clicking **select all**, and then **Download**). I then verify that the submissions appear in the `~/Sources` directory in `sd-svs`. Merging!

joshuathayer:
Woohoo! 🥇

PR #9, opened 2017-09-05
Support *gpg single file downloads

### LINKED ISSUE #8: Support *.gpg single files downloaded from the SecureDrop server
The `sd-process-download` script running in the `sd-journalist` AppVM currently expects the downloaded file to be a `*.zip`, which may not always be the case. SecureDrop downloads can be `*.zip` files or `*.gpg` files (`*.gpg` files are downloaded from the server by clicking the name of the individual submission in the individual source view). We should add support for these `*.gpg` single file downloads.

redshiftzero:
## Description

Closes #8.

This PR:
  * Extends the "Process SecureDrop Download" handler to the `application/pgp-encrypted` MimeType
  * Extends the `move-to-svs` script to also move `*.gpg` files to the SVS AppVM

## Testing

1. In `dom0` as a regular user in the directory with the `securedrop-workstation` repo code copied:

```
./reset.sh
```

2. Checkout this branch in your `work` AppVM and copy it to `dom0`.

3. In `dom0` as a regular user in the directory with the `securedrop-workstation` repo code copied:

```
./run.sh
```

4. When complete, start the `sd-journalist` AppVM, go to the journalist interface and download a single file by clicking this button:

howtodownloadgpg

5. Verify that the _default_ handler is "Process SecureDrop Download" and that a DispVM starts.

### DIFF
--- a/decrypt/decrypt-sd-submission
+++ b/decrypt/decrypt-sd-submission
@@ -22,8 +22,8 @@ with tarfile.open(input) as tar:
   # given a malicious tarball
   tar.extractall(tmpdir)
 
-# everything in the archive should be zips, as created by SD
-# let's unzip those here
+# some files may be zips, some may be .gpg
+# let's unzip the zip files here to produce .gpg files
 zips = glob.glob(tmpdir + "/*zip")
 for z in zips:
   with zipfile.ZipFile(z) as zf:
--- a/sd-journalist/move-to-svs
+++ b/sd-journalist/move-to-svs
@@ -9,9 +9,11 @@ import subprocess
 
 TB_DL_LOC="/home/user/.tb/tor-browser/Browser/Downloads"
 
+# Files downloaded from SecureDrop can end in .zip or .gpg
 zips = glob.glob(TB_DL_LOC + "/*.zip")
+gpgs = glob.glob(TB_DL_LOC + "/*.gpg")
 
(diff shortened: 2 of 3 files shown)


decrypt/decrypt-sd-submission
```diff
@@ -22,8 +22,8 @@ with tarfile.open(input) as tar:
   # given a malicious tarball
   tar.extractall(tmpdir)
 
-# everything in the archive should be zips, as created by SD
-# let's unzip those here
+# some files may be zips, some may be .gpg
+# let's unzip the zip files here to produce .gpg files
```
joshuathayer: There's a problem here... when we encounter zip files, we unzip in into `tmpdir + "/extracted/"`, which creates that directory if needed. Then, on line 54, we create a new tarball from the files in`tmpdir + "/extracted/"`.

The "bare" gpg files don't make it into .`.../extracted/` (and indeed `.../extracted/` isn't created if we're only dealing with bare gpg files) so the tarball creation on line 54 fails.

I suppose we could make the `.../extracted` dir by hand and move the bare gpg files into place there?

joshuathayer: It's sort of a drag because the only indication of the source name we have is the name of the `.gpg` file... for the example i'm working with now, that's `2-underhanded warrior-doc.gz-1`

redshiftzero: Good call @joshuathayer, I'll add a block to handle this

redshiftzero:
Closing in favor of #78

PR #28, opened 2017-09-19
Makefile and README update

joshuathayer:
This moves away from the `run.py`-based workflow, in favor of a Makefile which is able to clone, build, clean, and run tests on the installation. It also includes a small unrelated feature: it provides a desktop notification when decryption fails in the disp vm... that was a git mistake, but we may as well get it in master at some point.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,90 @@
+DEVVM=work
+DEVDIR=/home/user/projects/qubes-sd # important: no trailing slash
+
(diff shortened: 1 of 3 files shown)


redshiftzero:
Changed the base branch here to `master` since I believe that was your intention - but I pushed your existing branch to the shared remote (as `backup-makefile-and-fixes`) just in case I am incorrect :innocent:

PR #32, opened 2017-11-08
Fix #31 use sd-dispvm to open files from sd-svs

### LINKED ISSUE #31: dispvm does not correctly have its netvm set to none
Apparently the new sd-dispvm has its netvm set to sys-firewall. It should not have network access at all.

kushaldas:


joshuathayer:
Splended. Tested this and it works appropriately. Thanks!

PR #40, opened 2017-12-07
Give the user feedback during processing

joshuathayer:
Lots of changes to support sending progress back to the sd-journalist during processing. This is used to show progress to the user, but can also be used for integration testing.


conorsch:
@joshuathayer Note that #39 fights with this one a bit. We can merge #39 first and retarget this branch, or I can start QAing these changes on hardware immediately, and rebase #39 on top of this one once it's merged.

emkll:
I am seeing errors while doing `make all` with both this branch and master. I have since removed `securedrop-workstation` folder from `dom0` and re-copied it over. I still get the following error:
```
local:
  Data failed to compile:
-----------
  Detected conflicting IDs, SLS IDs need to be globally unique.
  The conflicting ID is 'sed -i 'li$tag:sd-decrypt-vm sd-gpg allow' /etc/qubes-rpc/policy/qubes.Gpg' and is found in SLS 'base:sd-dispvm' and SLS 'base:sd-decrypt'
```

joshuathayer:
@emkll , nope, that looks like a real bug in the sd-gpg provisioning code. I'll file an issue and fix it!

joshuathayer:
@emkll this looks like an issue in the provisioning code that's been fixed, but you may need to manually fix a broken file that the bug caused.

Take a look in `/etc/qubes-rpc/policy/qubes.Gpg` on `dom0`. I suspect you'll see this line repeated twice at the head of the file:

    $tag:sd-decrypt-vm sd-gpg allow

remove both instances of that line. Rerunning the provisioning scripts should then work- if you look back in that file, one instance of that line should appear, and won't incorrectly be added multiple times on subsequent provisioning.

emkll:
Thanks @joshuathayer. Local `/etc/qubes-rpc/policy/qubes.Gpg` in `dom0`  had 1 instance of that line, removed it. Ran `make clean` and `make all`, still get the same error. Now, the policy is not added to the file on subsequent runs:
```
cat /etc/qubes-rpc/policy/qubes.Gpg
$anyvm  $anyvm  ask
```

conorsch:
Now that #47 is merged, proceeding with review here. @emkll, the issue you encountered can be resolved by taking the steps described in #50.

conorsch:
Confirmed that the feedback overlay window indeed hangs. Debugging a bit interactively, turns out that `cat`ing the named pipe in `sd-journalist` allows the processing to proceed, and the subsequent decryption process continues once the pipe has been unblocked. Seems to me that #43 might be the culprit, there.

@joshuathayer Given how much work we have accrued in this PR, I vote for immediate merge, given that all the scaffolding we need is largely present. We can open a separate issue to debug the problem described above (again, may be related to #43, but I'll file a separate one to investigate and confirm or disprove).

joshuathayer:
Per @conorsch and @redshiftzero, we're going to merge this not-yet-quite-ready branch into master, since we don't think anyone expects master to be quite stable yet and it'll simplify some other merging and testing we need to do.

PR #41, opened 2017-12-07
Adds "validate" Makefile target

### LINKED ISSUE #34: The `make all` command should fail if `config.json` is missing
The `make all` target for building the VMs _requires_ a `config.json` file. If that file is missing, the VM creation cannot succeed, and will fail with rather cryptic error output.

Let's check for existence of that file and error out if it's missing, _before_ attempting to provision VMs without it.

conorsch:
Simply checking for existence of the JSON config file, which is
mandatory for creating SecureDrop Workstation VMs. Without the
`validate` check, the VM management logic will run, then fail loudly
with a rather cryptic message about the missing JSON config file.

Closes #34.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 DEVVM=work
 DEVDIR=/home/user/projects/securedrop-workstation # important: no trailing slash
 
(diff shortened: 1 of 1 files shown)


conorsch:
Conflicts resolved, ready for review.

Issue #54, opened 2018-01-29
Remove securedrop-workstation dir on dom0 as part of `make clone`


joshuathayer:
Currently, `make clone` on dom0 copies the files on the developer's work VM into `~/securedrop-workstation` on dom0. This brings over any changes the developer has been working on, but any *removed* files in the repo on the work VM will still exist in the copy on dom0. We've seen this cause hard-to-diagnose bugs.

We could address this by removing `~/securedrop-workstation` on dom0 as part of the `make clone` process. Thus, the now-empty directory would only be populated by files present in the developer's work VM. A potential downside is that any files changed by the user directly on dom0 would be overwritten. But, that's a workflow we discourage, anyway: all changes should be made in a branch in the work VM.

conorsch:
Resolved by #57.

PR #57, opened 2018-01-30
empty securedrop-workstation dir during make-clone

joshuathayer:
See #54- removes `~/sd-journalist/*` on dom0 during `make clone`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,26 @@
 DEVVM=work
 DEVDIR=/home/user/projects/securedrop-workstation # important: no trailing slash
+HOST=$(shell hostname)
 
-all: validate clean sd-whonix sd-svs sd-gpg sd-journalist sd-decrypt \
+assert-dom0:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,14 +1,26 @@
 DEVVM=work
 DEVDIR=/home/user/projects/securedrop-workstation # important: no trailing slash
+HOST=$(shell hostname)
 
-all: validate clean sd-whonix sd-svs sd-gpg sd-journalist sd-decrypt \
+assert-dom0:
+ifneq ($(HOST),dom0)
+	@echo "     ------ securedrop-workstation's makefile must be used only on dom0! ------"
+	exit 1
+endif
```
joshuathayer: Since the new "remove sd-journalist" step involves an `rm -rf ...`, I wanted to put at least a bit of logic to try to avoid removing the files in the wrong place or on the wrong VM. Thus this step, and using it in various recipes.

Issue #48, opened 2018-01-09
Migrate to fedora 26 templates


joshuathayer:
Fedora 25 was EOLed and Qubes has made Fedora 26 templates available. We should move any VMs off of 25 to 26.

conorsch:
Announcement by Qubes here: https://www.qubes-os.org/news/2018/01/06/fedora-26-upgrade/ Running these commands is not onerous, but we'll need tests to confirm that F26 is active in the relevant VMs, and fail is F25 is found.

conorsch:
Near as I can figure, the Qubes Admin API doesn't expose the base OS underlying a given app VM in a concise manner. One _can_ easily retrieve the template on which a VM is based, but that's still not the answer we want to write a pass/fail test for F25. 

Looks like we'll have to resort to parsing the contents of `/etc/os-release` on each VM and validating that against a list of permitted platforms.

conorsch:
Performed the manual commands [described in the Qubes docs](https://www.qubes-os.org/doc/template/fedora/upgrade-25-to-26/#qubes-40-instructions) before @redshiftzero helpfully pointed out _there's a better way_: https://github.com/QubesOS/qubes-issues/issues/3499 Will ask reviewers to use that method when the PR goes up.

PR #62, opened 2018-01-30
Lints all Python files via flake8

conorsch:
Cosmetic changes throughout, guided by `flake8`. The vast majority of changes were whitespace-related—pretty standard for Python projects—although in at least one case, the linter caught a genuine error in code, in which the test name `test_decrypt_sd_submission_desktop` was used twice, causing clobbering. That's resolved now.

Note that it's not easy to run `make flake8` in the development environment: you'll need to run `sudo dnf install python2-flake8` in the Work VM in order to use it. The goal is at least to get coverage in CI working, so that we have flake8 validation running prior to every PR merge.

Partial completion toward #56.

~~Marking this PR as "WIP" until I confirm CI is working as expected.~~ **Done, CI is passing.**

PR #63, opened 2018-01-31
Preliminary tests for platform (App VM base OS)

conorsch:
**This PR depends on #62, so review and merge that first.** Without the changes in #62, CI will never pass here.

Partial progress toward #48.

Trying to write pass/fail tests for supported platforms for all VMs.
Aiming for a hard fail on EOL Fedora 25, in support of migration to Fedora
26, but focusing on the test structure for now.

The hardcoded copypasta in the form of the VM list is not DRY, and
suggests to me we should consider a more declarative format for the test
assumptions, e.g. a single YAML file describing the VM layout and
attributes, which all tests can import once and reuse.

These tests are currently passing, and intended to. As we migrate to F26 as
described in #48, we should see these tests start failing.

### DIFF
--- a/tests/test_vms_platform.py
+++ b/tests/test_vms_platform.py
@@ -1,0 +1,55 @@
+import unittest
+
+from qubesadmin import Qubes
+
+
+SUPPORTED_PLATFORMS = [
+    "Fedora 25 (Twenty Five)",
+    "Debian GNU/Linux 8 (jessie)",
+]
+
+WANTED_VMS = [
+    "sd-decrypt",
+    "sd-gpg",
+    "sd-journalist",
+    "sd-svs",
+    "sd-svs-disp",
+    "sd-whonix",
+]
+
+
+class SD_VM_Platform_Tests(unittest.TestCase):
+    def setUp(self):
+        self.app = Qubes()
+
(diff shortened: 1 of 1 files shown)


tests/test_vms_platform.py
```diff
@@ -0,0 +1,55 @@
+import unittest
+
+from qubesadmin import Qubes
+
+
+SUPPORTED_PLATFORMS = [
+    "Fedora 25 (Twenty Five)",
```
kushaldas: One question to @joshuathayer, do we want to update to Fedora 26 for the vms?

joshuathayer: Yes, I think we should. There's an issue open for it (#48). What do you think?

conorsch: Yes, #48 is tracking the migration to F26—the tests are just an outline for tests for that migration. These tests pass, but when we resolve #48, we'll need to update these same tests, since they _should_ start failing under F26.

conorsch:
Rebasing on top of latest master, to include changes from #62 (specifically the CI config).

conorsch:
CI passed, approvals preserved, merging!

PR #64, opened 2018-01-31
Keep feedback UI from hanging.

### LINKED ISSUE #59: User feedback feature hangs in sd-journalist
During review of #40, was able to reproduce reliably the following:

1. Use sd-journalist VM to log into SecureDrop Journalist Interface
2. Download submission, choose the "Process SecureDrop submission" handler.
3. Observe that the "SecureDrop message processing" window displays... but hangs.
4. Open shell on `sd-journalist`, run `cat sdfifo`, and watch the processing continue, since it was unblocked. 
5. Run `cat sdfifo` multiple times, to step through the rest of the process.

Based on the above, sounds like the `sd-journalist` script that monitors the named pipe (`sdfifo`) is not working correctly. Note that #43 _sounds_ related, but here it appears the named pipe is indeed created correctly.

joshuathayer:
This closes #59.

The Python script which runs the UI for the feedback process reads a stream of updates off a FIFO (and messages are placed on the FIFO via a custom qrexec service triggered by events on remote SD VMs). 

The script had a relative path to the FIFO, which worked when testing the script by hand directly on sd-journalist. But when triggered automatically by Tor Browser upon downloading SD submissions, the UI script could not find the FIFO to read from. 

This fixes that problem by using an absolute path to the FIFO.

### DIFF
--- a/sd-journalist/sd-process-display
+++ b/sd-journalist/sd-process-display
@@ -104,7 +104,7 @@ if __name__ == '__main__':
     app = QApplication([])
     d = SDDialog()
 
-    reader = pipereader.PipeReader("sdfifo", poller_cb)
+    reader = pipereader.PipeReader("/home/user/sdfifo", poller_cb)
     t = threading.Thread(target=reader.read)
 
     t.start()


joshuathayer:
To test this, you'll need to change in to this branch on your work VM, `make clone` on dom0, then `make sd-journalist`. Then use TorBrowser on sd-journalist as normal to download some submissions. Submission process should commence and the UI should update as it progressing.

conorsch:
Wow, thanks for debugging, @joshuathayer. Looks like a remarkably simple fix.

conorsch:
Rebased on top of latest master to get CI passing. Will test locally to confirm resolution.

conorsch:
Confirmed that the decryption flow completes without manual nudging:

sd-source-interface-feedback-window-not-closing

PR #67, opened 2018-02-04
Updates tests to expect Fedora 26 (instead of Fedora 26)

### LINKED ISSUE #48: Migrate to fedora 26 templates

conorsch:
Closes #48.

Forces tests to fail if Fedora 25 is used anywhere. There are two places we need to check:

* inside the running SD VMs (by inspecting the contents of `/etc/os-release`)
* in the Qubes dom0 settings for new DispVMs (which may not be running and therefore won't be caught by the above check)

There's a long circuitous route for upgrading the templates in the Qubes docs, but when reviewing please use the Makefile target. I wrote the Makefile target _after_ I upgraded my VMs, but it should cover all your bases. If not, please comment.

### Testing
1. Checkout this feature branch in the work VM.
2. In dom0, run `make clone`
3. In dom0, run `make all`
4. In dom0, run `make test`
5. **Confirm tests fail due to Fedora 25 VMs.**
6. In dom0, run `make update-fedora-templates`
7. In dom0, re-run `make all`.
8. In dom0, re-run `make test`.
9. **Confirm tests pass.**

I also ran the integration tests and confirmed those were passing. Will update the documentation on the integration tests separately.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,8 @@ flake8: ## Lints all Python files with flake8
 	@flake8 .
 
 update-fedora-templates: assert-dom0 ## Upgrade Fedora 25 to Fedora 26 templates
-	sudo qubes-dom0-update
+	sudo qubes-dom0-update qubes-template-fedora-26
+	sudo qubes-prefs default_template fedora-26
 	sudo qubesctl state.sls qvm.default-dispvm
 	qubes-prefs default_dispvm fedora-26-dvm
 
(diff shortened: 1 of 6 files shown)


conorsch:
You're completely correct, @joshuathayer. Definitely needs the `qubes-template-fedora-26` addition on the dom0 update command. As for updating the default template for normal AppVMs, that I did manually (which is why the Makefile target is so shoddy). It _should_ be possible to run a Salt action similar to what's use for the DispVM update, but that's just a guess.

joshuathayer:
OK cool, glad I'm understanding this correctly. I'll take a look now at changing the default template via salt, or maybe specifying the template to use in the individual salt states.

PR #68, opened 2018-02-06
Installs Fedora 26 template and tells VMs to use it

joshuathayer:
This should fix the issue on @conorsch 's #67 PR, by installing Fedora 26 and using it as the template VM for the various sd-workstation salt configurations.

Tests pass, following the process in #67!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,6 +120,7 @@ flake8: ## Lints all Python files with flake8
 
 update-fedora-templates: assert-dom0 ## Upgrade Fedora 25 to Fedora 26 templates
 	sudo qubes-dom0-update qubes-template-fedora-26
+	sudo qubes-prefs default_template fedora-26
 	sudo qubesctl state.sls qvm.default-dispvm
 	qubes-prefs default_dispvm fedora-26-dvm
 
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,7 @@ flake8: ## Lints all Python files with flake8
 	@flake8 .
 
 update-fedora-templates: assert-dom0 ## Upgrade Fedora 25 to Fedora 26 templates
-	sudo qubes-dom0-update
+	sudo qubes-dom0-update qubes-template-fedora-26
 	sudo qubesctl state.sls qvm.default-dispvm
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -119,7 +119,7 @@ flake8: ## Lints all Python files with flake8
 	@flake8 .
 
 update-fedora-templates: assert-dom0 ## Upgrade Fedora 25 to Fedora 26 templates
-	sudo qubes-dom0-update
+	sudo qubes-dom0-update qubes-template-fedora-26
 	sudo qubesctl state.sls qvm.default-dispvm
 	qubes-prefs default_dispvm fedora-26-dvm
```
conorsch: Shouldn't we also add `qubes-prefs default_template fedora-26` here? I believe that would remove the requirement to hardcode the platform in the sls files, although having the sls files explicit is a Good Thing™ and I support keeping it.

conorsch: And, if so, then a `sudo qubesctl state.sls qvm.SOMETHING` seems warranted, as well, but `qvm.default-template` throws an error for me.

joshuathayer: To the first comment, yes probably! I couldn't find the right command to set the default template when I made this PR, but I'll try `qubes-prefs default_template fedora-26` and maybe open a new PR if that seems to work OK.

I don't quite understand your second comment!

joshuathayer: Actually I'm going to just add the `default_template` bit to this PR and merge it :)

conorsch: Adding the `default_template` change to this PR would be great, but during manual testing, I could not figure out what CLI command to use in dom0 in order to set the default VM template. It's clearly configurable via the GUI, but not sure about CLI yet. 

> I don't quite understand your second comment!

Rephrasing: In order to change the default _disposable_ VM template, two CLI commands were necessary from dom0: the first using `qubes-prefs` and the second using `qubesctl`. Only running one of those commands was not sufficient to update the default dispVM template. Therefore, we will need two commands to update the default AppVM template, as well—but I struggle to identify the correct Salt incantation.

PR #71, opened 2018-03-02
Adds AGPLv3 license

### LINKED ISSUE #70: Repository is missing a LICENSE file
This repository is missing a LICENSE file. I suggest that we use AGPLv3, purely for consistency with the main SecureDrop codebase.

https://github.com/freedomofpress/securedrop/blob/develop/LICENSE

joshuathayer:
The project was missing a license. This adds the AGPLv3 license, making it consistent with other FPF projects. Closes #70.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,663 @@
+Copyright (C) 2013 Aaron Swartz and James Dolan
+
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
+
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,663 @@
+Copyright (C) 2013 Aaron Swartz and James Dolan
```
conorsch: Nit: we should remove this line.

joshuathayer: Great catch! Will do.

PR #76, opened 2018-04-03
Add detailed rationale (pros/cons of the Qubes OS approach)

eloquence:
Per discussion in multiple venues, this PR exists so we have more of a shared reference point on why we're pursuing this project, and can validate these reasons iteratively. Let me know if I'm missing any key points or misunderstanding parts of the architecture. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,28 @@ This project aims to make journalists' experience working with SecureDrop less o
 
 This project is under active development, has known bugs and shortcomings, and is not ready for end users. This README is geared towards interested technical users and developers.
 
+### Detailed rationale
+
+SecureDrop's security model is proven, but it has major drawbacks:
+
+- Journalists must access a separate, airgapped device to even validate that a submission is relevant. This is very onerous, and may reduce the reliance on SecureDrop overall.
+
+- The complexity of the setup and the usage procedures create operational security risks. For example, journalists may accidentally boot up the wrong device using the SVS USB drive, breaking the airgap.
+
+- Once a document has been decrypted on the SVS, a journalist is more or less on their own right now. Work on the submission and the collaboration with other journalists are "not our department". Yet, security failures are likely during this stage. It's difficult to address this fundamental issue with the current workflow, since the SVS is essentially a dead end as far as SecureDrop is concerned.
+
+The Qubes OS approach addresses this at multiple levels:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,6 +6,28 @@ This project aims to make journalists' experience working with SecureDrop less o
 
 This project is under active development, has known bugs and shortcomings, and is not ready for end users. This README is geared towards interested technical users and developers.
 
+### Detailed rationale
+
+SecureDrop's security model is proven, but it has major drawbacks:
```
conorsch: Perhaps link to https://docs.securedrop.org/en/stable/threat_model/threat_model.html ? It's a bit dated, but it's what we have documented public. As a newcomer, I'd wonder immediately what we mean by "SecureDrop's security model," and that's the best answer available in public docs.

README.md
```diff
@@ -6,6 +6,28 @@ This project aims to make journalists' experience working with SecureDrop less o
 
 This project is under active development, has known bugs and shortcomings, and is not ready for end users. This README is geared towards interested technical users and developers.
 
+### Detailed rationale
+
+SecureDrop's [architecture](https://docs.securedrop.org/en/latest/overview.html#infrastructure) and [threat model](https://docs.securedrop.org/en/stable/threat_model/threat_model.html) are proven, but the current approach also has major drawbacks:
+
+- Journalists must access a separate, airgapped device to even validate that a submission is relevant. This is very onerous, and may reduce the reliance on SecureDrop overall.
+
+- The complexity of the setup and the usage procedures create operational security risks. For example, journalists may accidentally boot up the wrong device using the SVS USB drive, breaking the airgap.
```
redshiftzero: We can add: the cumbersome submission workflow that must be followed even for spam submissions may cause journalists to develop workarounds that reduce the overall security of the SecureDrop platform.

README.md
```diff
@@ -6,6 +6,28 @@ This project aims to make journalists' experience working with SecureDrop less o
...
+
+- Once a document has been decrypted on the SVS, a journalist is more or less on their own right now. Work on the submission and the collaboration with other journalists are "not our department". Yet, security failures are likely during this stage. It's difficult to address this fundamental issue with the current workflow, since the SVS is essentially a dead end as far as SecureDrop is concerned.
```
redshiftzero: All these points are valid and well-put. An additional point that is important with respect to the Qubes Workstation: 

- Timely software updates are an important component of any secure system. Unfortunately, the process for applying security updates to the SVS is difficult, and as such the SVS can lag behind in security patches. This creates additional vectors of attack, especially since the journalist is by construction opening unknown files in the workstation as the private submission key.

PR #77, opened 2018-04-04
Make the feedback processes smarter in the face of a blocking socket

### LINKED ISSUE #65: Keep processing from hanging when sd-journalist FIFO is not attended to
In the happiest path, we have a UI process in sd-journalist which is reading from a FIFO to update the user feedback interface during submission processing. 

If for some reason the UI process crashes or otherwise stops reading from the pipe, remote processes trying to write to the pipe via our custom qrexec service hang forever (since writing to a FIFO is blocking until someone consumes reads it). This was why we saw the process hang in #59.

The fix in #59 was to address the inability for the UI process to read from the pipe, but we should do a more general fix to keep the remote calling process from hanging when nothing is reading the pipe. We don't want to be in a situation where multiple submission-handling processes are operating in parallel, which we could see if users "try again" after a process gets hung.

A potential approach would be to put a timeout in the write in `sd-journalist/sd-process-feedback`

joshuathayer:
This closes #65. If the monitoring UI dies (user closes the window, perhaps), the feedback FIFO loses its reader. In a naive implementation, any processes trying to write to that pipe will block. This has the unhappy effect of stopping all SD submission processing. 

This changes how we write to that FIFO. Rather than using the shell's `echo` built-in, we use a Python script to open a nonblocking file handle to the FIFO. If the open fails, we silently move on, which allows submission processing to continue. If the open works, we write to the FIFO, which will update the user interface.

### DIFF
--- a/sd-journalist/sd-process-feedback
+++ b/sd-journalist/sd-process-feedback
@@ -1,7 +1,15 @@
-#!/bin/sh
+#!/usr/bin/python
 
-# `read` reads exactly one line from stdin as a literal string.
-read arg
+import errno
+import posix
+import sys
+import os
 
-# our process on the other end will read at most BUFSIZE bytes
-echo $arg > /home/user/sdfifo
+arg = sys.stdin.readline()
+
+try:
+    fifo = os.open('/home/user/sdfifo', posix.O_WRONLY | posix.O_NONBLOCK)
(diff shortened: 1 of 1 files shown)


redshiftzero:
Thanks for the PR @joshuathayer and for testing @kushaldas!

PR #78, opened 2018-04-05
Make single-file gpg and msg downloads work

### LINKED ISSUE #8: Support *.gpg single files downloaded from the SecureDrop server

joshuathayer:
Closes #8 

This extends the approach Jen took in #9. Extending the workstation to handle single file downloads only required updating the decryption phase (aside from updating the MIME registry in sd-journalist to make our own scripts handle `.gpg` files).

A confounding problem is that TorBrowser seems to want to add numeric suffixes to the names of "downloaded" files if they've already been downloaded in the past- even if the file was not saved to disk but was handled by a different application (as is the case here- our scripts handle downloaded submissions). So, our code needs to expect and handle numeric suffixes on .gpg files. We strip the number off the filename, so we don't end up with many copies of the same file, named slightly differently, on `sd-svs`. I handle that process in `sd-decrypt`, but a strong case could be made that it belongs in `sd-journalist`, I suppose.

The "numeric suffix" problem doesn't turn up when downloading multiple files from SD, since it generates a new filename for each bundle. 

Another oddity is that when downloading multiple files from securedrop, it puts each file in its own directory in the generated tarball. Our processing does not remove that directory level, so if (over time) a journalist downloads both single files and a bundle which contains those same files, they'll end up with a tree that might look like-

    Sources/monistic_convention/1-monistic_convention-doc
    Sources/monistic_convention/2-monistic_convention-msg
    Sources/monistic_convention/1_2017-09-22/1-monistic_convention-doc
    Sources/monistic_convention/2_2017-09-22/2-monistic_convention-msg
   
i.e., the same files are in two different places. I am not sure why SD adds that extra directory layer, but we should probably just remove that level in our processing.

### DIFF
--- a/sd-decrypt/decrypt-sd-submission
+++ b/sd-decrypt/decrypt-sd-submission
@@ -8,7 +8,8 @@ import zipfile
 import glob
 import subprocess
 import shutil
-
+import re
+import gzip
 
 def send_progress(msg):
     p = subprocess.Popen(["qrexec-client-vm", "sd-journalist",
@@ -51,6 +52,17 @@ for z in zips:
         zf.extractall(tmpdir + "/extracted/")
     os.unlink(z)
 
+gpgs = glob.glob(tmpdir + "/*gpg")
+for g in gpgs:
+    match = re.search('\d-(.*)-(msg|doc).*$', g)
+    source = match.group(1)
+    target = "{}/extracted/{}".format(tmpdir, source)
+
+    if not os.path.exists(target):
+        os.makedirs(target)
+
+    os.rename(g, "{}/{}".format(target, os.path.basename(g)))
+
 send_progress("SUBMISSION_FILES_EXTRACTED")
 
 # great, we should be left with a directory tree filled with files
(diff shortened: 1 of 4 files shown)


sd-decrypt/decrypt-sd-submission
```diff
@@ -51,6 +52,17 @@ for z in zips:
         zf.extractall(tmpdir + "/extracted/")
     os.unlink(z)
 
+gpgs = glob.glob(tmpdir + "/*gpg")
```
kushaldas: we should use `os.path.join` function call here. I can see other parts also where we have used string concatenations, we should use the join call everywhere.

sd-decrypt/decrypt-sd-submission
```diff
@@ -51,6 +52,17 @@ for z in zips:
...
+for g in gpgs:
+    match = re.search('\d-(.*)-(msg|doc).*$', g)
+    source = match.group(1)
+    target = "{}/extracted/{}".format(tmpdir, source)
+
+    if not os.path.exists(target):
+        os.makedirs(target)
+
+    os.rename(g, "{}/{}".format(target, os.path.basename(g)))
```
kushaldas: It would be nicer to read if we have `os.path.join` here too.

joshuathayer: 👍

kushaldas:
@joshuathayer Thank you, I will test the branch today.

PR #97, opened 2018-05-30
Add Data Flow Diagram

### LINKED ISSUE #94: Formal threat modeling
I think we're at the point where we need to do more formal threat modeling (i.e. what is the diff between the existing architecture and the new integrated Qubes workstation)

emkll:
Fixes #94 
Describes all components and data flows in the current implementation of the SecureDrop workstation, using the draw.io tool.


eloquence:
@joshuathayer Can you take another look and merge this if it's good to go from your perspective?

joshuathayer:
👍 shipit!

Issue #114, opened 2018-07-23
Use Fedora 28 templates


emkll:
Fedora 26 is now end-of-life [0], AppVMs that use Fedora 26 template should be updated to a more recent release. We should use Fedora 28 as 27 will be EOL 1 month after 29 release which is is tentatively scheduled for later this year [1].

[0] : https://fedoraproject.org/wiki/End_of_life
[1] : https://fedoraproject.org/wiki/Releases/29/Schedule

conorsch:
See here for upgrade instructions from Qubes team: https://www.qubes-os.org/doc/template/fedora/upgrade-27-to-28/

conorsch:
Used the [detailed instructions](https://www.qubes-os.org/doc/template/fedora/upgrade-27-to-28/#detailed-upgrading-the-standard-fedora-27-template-to-fedora-28-1) to upgrade Fedora 26 -> Fedora 27, will run through the same process for 27 -> 28 and report back.

The config tests will need to be updated for F28, as we did in #67.

emkll:
I recommend running `sudo qubes-dom0-update qubes-template-fedora-28` to pull the template directly from the qubes repos

PR #115, opened 2018-07-23
Automate Securedrop Workstation Qubes template building

### LINKED ISSUE #109: Automate the creation of grsecurity template using Qubes Builder
Through the work in #104 we learned that it is feasible to run grsecurity-enabled VMs as template HVMs in Qubes OS. The memory overhead of HVMs  (as opposed to PVH-based VMs) is a bit higher but not prohibitively so. The next step is to automate the creation of grsecurity template HVMs using [Qubes Builder](https://www.qubes-os.org/doc/qubes-builder/). This will result in RPM packages that can be installed in `dom0`.

Since we don't fully understand what's involved yet, we have timeboxed this task to 16 hours for the 7/11-7/25 sprint. If it's not completed by the end of the timebox, @emkll (who will likely continue to work on this) will post an update here.

Note that this still gets us only part of the way to transitioning to grsecurity VMs (esp. for the `sd-svs` and `sd-journalist` VMs); we will still have to figure out the distribution/update strategy we want to use for these packages.

emkll:
`build-workstation-template` in builder automatically builds a Qubes SecureDrop workstation template .rpm to be installed in dom0.

Installing this template will provide a usable grsecurity-hardened template. 

Fixes #109 

### Test plan:
1. create a fedora-28 vm for building purposes
2. increase the disk size to at least 15GB (as the build uses over 10GB)
3. run `build-workstation-template`
4. copy the rpm generated in `/home/user/src/securedrop-workstation/builder/qubes-builder/qubes-src/linux-template-builder/rpm/` to dom0
5. `sudo rpm -i <file>.rpm` (this takes a few minutes)
6. `qvm-create --template grsec-workstation test-grsec-kernels --class AppVM --property virt_mode=hvm --property kernel='' --label green`
7. open a terminal window in `test-grsec-kernels` and ensure the grsec kernel is running

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ For more information on the integration tests, run `test_integration --help`.
 1. Create a fedora-28 AppVM for building
 2. Increase the disk size to at least 15GB (as the build uses over 10GB)
 3. Import the QubesOS master key and the GPG key used to sign tags (see https://www.qubes-os.org/security/verifying-signatures/)
-4. Run `build-workstation-template` in the `builder` directory
+4. Run `make template` in the top-level of this repository.
 5. Copy the rpm generated in `/home/user/src/securedrop-workstation/builder/qubes-builder/qubes-src/linux-template-builder/rpm/` to `dom0`
 6. Install the template in `dom0` : `sudo rpm -i <file>.rpm` (this takes a few minutes)
 7. Create a new VM based on this template:
--- a/README.md
+++ b/README.md
@@ -167,6 +167,19 @@ and run tests with
 
 For more information on the integration tests, run `test_integration --help`.
 
+## Building the templates
+
+1. Create a fedora-28 AppVM for building
+2. Increase the disk size to at least 15GB (as the build uses over 10GB)
+3. Import the QubesOS master key and the GPG key used to sign tags (see https://www.qubes-os.org/security/verifying-signatures/)
+4. Run `build-workstation-template` in the `builder` directory
+5. Copy the rpm generated in `/home/user/src/securedrop-workstation/builder/qubes-builder/qubes-src/linux-template-builder/rpm/` to `dom0`
+6. Install the template in `dom0` : `sudo rpm -i <file>.rpm` (this takes a few minutes)
+7. Create a new VM based on this template:
+```
+qvm-create --template grsec-workstation test-grsec-kernels --class AppVM --property virt_mode=hvm --property kernel='' --label green
+```
+
 ## Threat model
 
 This section outlines the threat model for the SecureDrop workstation, and should complement [SecureDrop's threat model](https://docs.securedrop.org/en/stable/threat_model/threat_model.html). This document is always a work in progress, if you have any questions or comments, please open an issue on [GitHub](https://github.com/freedomofpress/securedrop-workstation) or send an email to [securedrop@freedom.press](mailto:securedrop@freedom.press).
--- a/builder/build-workstation-template
+++ b/builder/build-workstation-template
@@ -10,6 +10,7 @@ repo_root="$(git rev-parse --show-toplevel)"
 build_dir="${repo_root}/builder"
 qubes_builder_dir="${build_dir}/qubes-builder"
 sd_template_dir="${build_dir}/qubes-template-securedrop-workstation"
+gpg_homedir="${repo_root}/builder/qubes-builder/keyrings/git"
 
 # Chroot may exist due to previous failed build; let's clean up responsibly.
 if [[ -d "${qubes_builder_dir}" ]]
@@ -32,6 +33,15 @@ cp ${sd_template_dir}/securedrop-workstation.conf builder.conf
 
 # Install dependencies from Qubes builder logic.
 make install-deps
+
+# Get sources for only the builder component, which will get Qubes dev keys and initialize the keyring
+make COMPONENTS="builder" get-sources
+
+# Add signing key to the keyring
+gpg --homedir ${gpg_homedir} --keyserver pool.sks-keyservers.net --recv-key AF775782949D263DAABB3387AAFB3575FAC82745 || exit 1;
+echo 'AF775782949D263DAABB3387AAFB3575FAC82745:6:' | gpg --homedir ${gpg_homedir} --import-ownertrust;
+
+# Get all sources
 make get-sources
 
 # Build it!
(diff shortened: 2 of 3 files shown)


builder/build-workstation-template
```diff
@@ -17,33 +16,18 @@ sd_template_dir="${build_dir}/qubes-template-securedrop-workstation"
...
-./setup # Appears this is required, worth confirming
+# Provision securedrop-workstation config so that we get the correct sources via `make get-sources`.
+cp ${sd_template_dir}/securedrop-workstation.conf builder.conf
 
 # Install dependencies from Qubes builder logic.
 make install-deps
-
-# These *should* be pulled in automatically, but currently aren't.
-# May be worth a PR targeting upstream.
-sudo dnf install -y perl-Digest-MD5 perl-Digest-SHA
```
conorsch: Removing this line causes failure for me! Can you try rebooting your VM and seeing if it's really not necessary? You may have the packages sticking around from a previous invocation of this line.

builder/build-workstation-template
```diff
@@ -0,0 +1,33 @@
...
+set -o pipefail
+
+repo_root="$(git rev-parse --show-toplevel)"
+build_dir="${repo_root}/builder"
+qubes_builder_dir="${build_dir}/qubes-builder"
+sd_template_dir="${build_dir}/qubes-template-securedrop-workstation"
+
+# Heavy-handed, but ensures build is clean, and also sidesteps idempotence
+# issues with the git clone/checkout commands.
+rm -rf "${qubes_builder_dir}" "${sd_template_dir}"
```
conorsch: The `rm -rf` here doesn't provide sufficient cleanup. If a build fails, and the chroot is provisioned, then the `rm -rf` will nuke all files _other_ than the chroot (which is root-owned). We should run `make -C "${qubes_builder_dir}" clean-chroot` or similar in order to tear down the chroot responsibly, _then_ proceed with removing the dir.

conorsch:
@emkll Please try again. I pushed up a commit addressing the lingering chroot issue. If you can reproduce the perl dependency failure, feel free to push up a fix. We'll squash the script down prior to merge once we're satisfied with it.

builder/build-workstation-template
emkll: I've retested this in a clean VM, and the dependency is not required. I have recently upgraded to fedora-28 templates. Given that f26 is end of life (see https://github.com/freedomofpress/securedrop-workstation/issues/114), I think we can safely leave this out of the script for the time being.

conorsch: > recently upgraded to fedora-28 templates

Ah, thanks for clarifying. Alright, let's punt to #114, then—if anyone reports this error, we'll know their templates are out of date. :slightly_smiling_face:

conorsch:
Squashed commits, since script is looking stable. Still waiting on a full end-to-end build to report success before formally approving.

kushaldas:
Without using split-gpg, I got the following.

```
-> Updating sources for template-securedrop-workstation...
--> Fetching from https://github.com/freedomofpress/qubes-template-securedrop-workstation master...
--> Verifying tags...
No valid signed tag found!
---> One of invalid tag:
object 9e037c4c8c304d6e898559c53480948a90c50a89
type commit
tag v0.0.1
tagger mickael e <mickael@freedom.press> 1532385694 -0400

v0.0.1
gpg: Signature made Mon 23 Jul 2018 06:41:39 PM EDT
gpg:                using RSA key 49F3ED89F8597EF7
gpg: Can't check signature: public key not found
make: *** [Makefile:194: template-securedrop-workstation.get-sources] Error 1
```

emkll:
@kushaldas the error you are seeing is because the QubesOS master public key is not signed/trusted. You need to import it and set the trust to `ultimate` per the instructions here: https://www.qubes-os.org/security/verifying-signatures/

kushaldas:
I already have that.

```
$ gpg2 --edit-key 0x427F11FD0FAA4B080123F01CDDFA1A3E36879494
gpg (GnuPG) 2.2.8; Copyright (C) 2018 Free Software Foundation, Inc.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

gpg: starting migration from earlier GnuPG versions
gpg: porting secret keys from '/home/user/.gnupg/secring.gpg' to gpg-agent
gpg: migration succeeded

pub  rsa4096/DDFA1A3E36879494
     created: 2010-04-01  expires: never       usage: SC  
     trust: ultimate      validity: ultimate
[ultimate] (1). Qubes Master Signing Key

```

conorsch:
On a fresh Qubes installation, the GPG bootstrapping for verifying the build components is problematic enough that we need docs dedicated to it. While I _was_ able to finagle a working verification on a previous Qubes incarnation, I am now bumping into problems identical to what @kushaldas reports.

Will step through the use of `split-gpg` in a dedicated vault VM and report back on whether that resolves.

conorsch:
Unable to resolve the verification failures. Let's pair tomorrow, @emkll, and tag-team some docs updates. As I recall, the verification proceeded without error when we had no tag at all to verify. If that remains the case, I'm inclined to punt for now, and open a separate issue dedicated to sorting out the correct signing keys. As it stands, even if we sort out the docs for tag v0.0.1, we'll still need to circle back and update the signing procedure as soon as a key other than your personal key is used to update the template.

emkll:
@conorsch @kushaldas this should be fixed in the latest changes pushed to this branch in cd2dcc707d419e400a783180af0a5a6b3a3bd9ee : 

The builder uses a keyring that is imported when the builder sources are pulled in, and can be found in `qubes-builder/keyrings/git/`

conorsch:
@emkll I shall take these new changes for a spin! :mag:

conorsch:
So far so good! Got past the failing verification steps. Proceeding with full end-to-end build, will report results.

README.md
```diff
@@ -167,6 +167,19 @@ and run tests with
 
 For more information on the integration tests, run `test_integration --help`.
 
+## Building the templates
+
+1. Create a fedora-28 AppVM for building
+2. Increase the disk size to at least 15GB (as the build uses over 10GB)
```
conorsch: I've been struggling to resize my root volume; even when it works, and dom0 does not emit a cryptic error, I still run out of disk space. What I've been doing is resizing the _private_ volume for the build VM, which is both more reliable in terms of the resizing operation's output, and properly accommodates the filesystem-hungry build process.

Noting this rather than requesting changes because it seems more research required; we'll finetune the docs as we use these features more regularly during template maintenance.

README.md
```diff
@@ -167,6 +167,19 @@ and run tests with
...
+3. Import the QubesOS master key and the GPG key used to sign tags (see https://www.qubes-os.org/security/verifying-signatures/)
```
conorsch: We don't actually need this step anymore, given recent script changes, but it's OK to document for a while.

Issue #51, opened 2018-01-29
Create VM provisioning utility script


conorsch:
Errors in the Salt provisioning flow (e.g. `make all`) often do not percolate up to cause the `make` invocation to exit non-zero. This means that it can _look_ like the testing VMs were created successfully, but some actually lack required files to operate as part of the SD workstation environment.

One such example occurred during review of #47, when configuring the `sd-decrypt`:

```
Summary for local
-----------
Succeeded: 26 (changed=5)
Failed: 0
-----------
Total states run: 26
Total run time: 5.960 s
sd-decrypt: ERROR (exception [('/srv/salt/_tops/base/sd-dispvm.top', '/tmp/tmpzqv1xs/srv/salt/_tops/sd-dispvm.top', "[Errno 2] No such file or directory: '/srv/salt/_tops/base/sd-dispvm.top'"), ('/srv/salt/_tops/base/sd-dispvm-files.top', '/tmp/tmpzqv1xs/srv/salt/_tops/base/sd-dispvm-files.top', "[Errno 2] No such file or directory: '/srv/salt/_tops/base/sd-dispvm-files.top'")])
sudo qubesctl top.enable sd-svs-disp
```

Note that I transcribed the error output above by hand, due to lack of copy/paste support from dom0, so there may be typos. The problem isn't so much the particular error, more that the `sd-decrypt` VM was not fully configured, and yet `make all` exited zero, implying no problems.

redshiftzero:
Indeed, currently `make all` returns 1 only in a subset of possible failure scenarios, e,g, if `/srv/salt/sd` is already populated it will return 1. I hit the same issue as you in #72: VMs were created, but some critical files were _not_ created, but `make all` returned 0. 

While the Makefile is definitely super handy during development, I think what we'd eventually want for provisioning in `dom0` is a Python script that can do e.g. the following:

```
securedrop-vm-manager create all
securedrop-vm-manager delete all
securedrop-vm-manager create <vm-name>
securedrop-vm-manager delete <vm-name>
```

The target user here would be the admin but of course this would be useful for development as well (we'd want a `--verbose, -v` flag that would spit out all output for debugging issues).

[...]

PR #118, opened 2018-07-25
Uses Fedora 28 templates for all Fedora VMs

### LINKED ISSUE #114: Use Fedora 28 templates

conorsch:
Fedora 26 is now EOL. Rather than switch to 27, let's jump to 28 and buy ourselves a bit more breathing room on the lifecycle.

Closes #114.

### Testing

1. Pull the latest changes from this branch into dom0 via the `qvm-run` tar command (see README).
2. Run `make update-fedora-templates`.
3. Run `make all`.
4. Run `make test`.
5. Confirm all tests pass.

I'll note that the tests still permit Debian Jessie VMs, so we should definitely update that to force Stretch for any Debian hosts used in the config. Deferring that to a separate changeset, though.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,10 +126,7 @@ flake8: ## Lints all Python files with flake8
 		| xargs flake8
 
 update-fedora-templates: assert-dom0 ## Upgrade to Fedora 28 templates
-	sudo qubes-dom0-update qubes-template-fedora-28
-	sudo qubes-prefs default_template fedora-28
-	sudo qubesctl state.sls qvm.default-dispvm
-	qubes-prefs default_dispvm fedora-28-dvm
+	@./scripts/update-fedora-templates
 
 template: ## Builds securedrop-workstation Qube template RPM
 	./builder/build-workstation-template
--- a/Makefile
+++ b/Makefile
@@ -125,11 +125,11 @@ flake8: ## Lints all Python files with flake8
 		| perl -F':\s+' -nE '$$F[1] =~ m/text\/x-python/ and say $$F[0]' \
 		| xargs flake8
 
-update-fedora-templates: assert-dom0 ## Upgrade Fedora 25 to Fedora 26 templates
-	sudo qubes-dom0-update qubes-template-fedora-26
-	sudo qubes-prefs default_template fedora-26
+update-fedora-templates: assert-dom0 ## Upgrade to Fedora 28 templates
+	sudo qubes-dom0-update qubes-template-fedora-28
+	sudo qubes-prefs default_template fedora-28
 	sudo qubesctl state.sls qvm.default-dispvm
-	qubes-prefs default_dispvm fedora-26-dvm
+	qubes-prefs default_dispvm fedora-28-dvm
 
 template: ## Builds securedrop-workstation Qube template RPM
 	./builder/build-workstation-template
--- a/scripts/update-fedora-templates
+++ b/scripts/update-fedora-templates
@@ -1,0 +1,28 @@
+#!/bin/bash
+# Utility to update the default Fedora templates used by Qubes
+# when creating other VMs.
+set -e
+set -u
+set -o pipefail
+
+
+# Install the Qubes-maintained Fedora template if absent.
+# We must use conditional logic, since the update command
+# will fail if the package is already installed.
+if ! rpm -qa | grep -q ^qubes-template-fedora-28 ; then
+    sudo qubes-dom0-update qubes-template-fedora-28
+fi
+
+# Check whether a suitable DispVM exists.
+if ! qvm-ls --quiet fedora-28-dvm 2>&1 > /dev/null ; then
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -125,11 +125,11 @@ flake8: ## Lints all Python files with flake8
 		| perl -F':\s+' -nE '$$F[1] =~ m/text\/x-python/ and say $$F[0]' \
 		| xargs flake8
 
-update-fedora-templates: assert-dom0 ## Upgrade Fedora 25 to Fedora 26 templates
-	sudo qubes-dom0-update qubes-template-fedora-26
-	sudo qubes-prefs default_template fedora-26
+update-fedora-templates: assert-dom0 ## Upgrade to Fedora 28 templates
+	sudo qubes-dom0-update qubes-template-fedora-28
```
emkll: If the `qubes-template-fedora-28` is installed, this command will fail and return `Error: unable to find a match`. I think we need to check for the presence of the fedora 28 template beforehand (e.g., `rpm -qa | grep qubes-template-fedora-28`).

Makefile
```diff
@@ -125,11 +125,11 @@ flake8: ## Lints all Python files with flake8
...
+	sudo qubes-prefs default_template fedora-28
 	sudo qubesctl state.sls qvm.default-dispvm
-	qubes-prefs default_dispvm fedora-26-dvm
+	qubes-prefs default_dispvm fedora-28-dvm
```
emkll: I get a qubes-prefs error on this line: `No such domain \'fedora-28-dvm\'`
It seems like we need to create a fedora-28 dvm manually from the template (as I believe fedora-26-dvm was included by default in the Qubes R4 install). The following two commands should do the trick [0] :
```
qvm-create --template fedora-28 --label red fedora-28-dvm
qvm-features fedora-28-dvm appmenus-dispvm 1
```

[0] : https://www.qubes-os.org/doc/dispvm-customization/

conorsch:
Great feedback, @emkll! Implemented the logic you suggested; commands appear idempotent now, so you should be able to re-run without error.

scripts/update-fedora-templates
```diff
@@ -0,0 +1,28 @@
...
+    qvm-features fedora-28-dvm appmenus-dispvm 1
+fi
+
+# Set latest Fedora as the default DispVM throughout.
+qubes-prefs default_template fedora-28
+qubes-prefs default_dispvm fedora-28-dvm
+
+# The Qubes Salt config must override the fedora-26-dvm setting.
+# We aren't managing that explicitly yet.
+# sudo qubesctl state.sls qvm.default-dispvm
```
conorsch: N.B. Running the `qvm.default-dispvm` Salt command still identifies `fedora-26-dvm` as the correct DispVM; which it isn't. The declaration defaults in `/srv/formulas/base/virtual-machines-formula/qvm/default-dispvm.sls`. The README in that directory states that the settings can be overridden via `/srv/pillar/base/qvm/init.sls`, which we're _not_ currently doing.

scripts/update-fedora-templates
```diff
@@ -0,0 +1,28 @@
...
+# We must use conditional logic, since the update command
+# will fail if the package is already installed.
+if ! rpm -qa | grep -q ^qubes-template-fedora-28 ; then
+    sudo qubes-dom0-update qubes-template-fedora-28
+fi
+
+# Check whether a suitable DispVM exists.
+if ! qvm-ls --quiet fedora-28-dvm 2>&1 > /dev/null ; then
+    qvm-create --template fedora-28 --label red fedora-28-dvm
+    qvm-features fedora-28-dvm appmenus-dispvm 1
```
emkll: I have encountered another error while running  `make all` : an exception is raised 
```
exception template for DispVM (fedora-28-dvm) needs to be an AppVM with template_for_dispvms=True
```

the following command seems to resolve this [0]:
`qvm-prefs fedora-28-dvm template_for_dispvms True`

[0] https://www.qubes-os.org/doc/dispvm-customization/

conorsch: Thanks for reporting, @emkll. Encountered that stipulation in the docs, and was surprised when I didn't need to implement it in the script, during transcription of your requested changes. Let me check my setup and see if I made the mistake of running the action via CLI (which would explain the lack of error)...

conorsch: Was indeed able to reproduce viewing the error; but the `make all` command was still exiting zero. See #51 for tracking those silent failures.

conorsch:
Revised, @emkll! Ready for re-re-review.

Issue #122, opened 2018-08-13
Update whonix to v14


emkll:
Whonix 14 has been released [0]. Whonix 14 uses Debian Stretch as a base OS. Whonix 13 uses Debian Jessie, which is no longer supported by the Debian security team [1] but still receives community support. We should update the whonix gateway (and workstation?).

The template can be installed in `dom0` as follows, and changing the netvm of the appropirate VMs should be more reliable than updating whonix 13 vms in place:
```
sudo qubes-dom0-update --enablerepo=qubes-template-community qubes-template-whonix-gw-14
```

[0] : https://forums.whonix.org/t/whonix-14-has-been-released/5404/5
[1] : https://wiki.debian.org/LTS

edenemmanuel:
On Friday, 24 August, the Qubes OS team [announced that Whonix 13 is approaching EOL](https://www.qubes-os.org/news/2018/08/24/whonix-13-approaching-eol/) and have provided helpful instructions.

Also, worth noting:

>We also provide a fresh Whonix 14 TemplateVM package through the Qubes repositories, which you can install in dom0 by following the [Whonix installation guide](https://www.whonix.org/wiki/Qubes/Install). If you encounter any difficulties when attempting to upgrade or install Whonix templates, please consult the [Whonix Support](https://www.whonix.org/wiki/Support) page.

redshiftzero:
Oops, we updated to Whonix 14 in commit b17faf0928c02f38f1290b6a6e369514a11cc9cd in #124, closing!

kushaldas:
>sudo qubes-dom0-update --enablerepo=qubes-template-community qubes-template-whonix-gw-14

Reopening the issue as the exact same command, which worked before, can not find the template today. I will close it when more people can verify that it is working again. I found more people in the `#qubes` IRC channel with the same issue.

kushaldas:
False alarm, the `qubes-dom0-update` tool behaves strangely if a package is already installed.

PR #124, opened 2018-08-15
Enable apparmor for whonix gateway and workstation vms

### LINKED ISSUE #108: Enable AppArmor in the sd-journalist VM
For Whonix templates the use of AppArmor profiles is disabled by default and is opt-in [0]. We should consider enabling AppArmor in both Whonix templates :
* whonix-ws: which will be used by journalist to browse the Internet on the securedrop-workstation
* whonix-gw: which will be used by the securedrop client to communicate with the SecureDrop server's journalist interface.

Since we should be moving to Whonix 14, this should be done as part of or after #122
[0] : https://www.whonix.org/wiki/Qubes/AppArmor

emkll:
Fixes #108 and #122 
Based on the instructions in https://www.whonix.org/wiki/Qubes/AppArmor, AppArmor is enabled on the whonix template prior to creating the sd-whonix and sd-journalist AppVMs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,12 +129,15 @@ flake8: ## Lints all Python files with flake8
 update-fedora-templates: assert-dom0 ## Upgrade to Fedora 28 templates
 	@./scripts/update-fedora-templates
 
+update-whonix-templates: assert-dom0 ## Upgrade to Whonix 14 templates
+	@./scripts/update-whonix-templates
+
 template: ## Builds securedrop-workstation Qube template RPM
 	./builder/build-workstation-template
 
-prep-whonix: ## enables apparmor on whonix-ws and whonix-gw
-	qvm-prefs -s whonix-gw kernelopts "nopat apparmor=1 security=apparmor"
-	qvm-prefs -s whonix-ws kernelopts "nopat apparmor=1 security=apparmor"
+prep-whonix: ## enables apparmor on whonix-ws-14 and whonix-gw-14
+	qvm-prefs -s whonix-gw-14 kernelopts "nopat apparmor=1 security=apparmor"
+	qvm-prefs -s whonix-ws-14 kernelopts "nopat apparmor=1 security=apparmor"
 
 # Explanation of the below shell command should it ever break.
 # 1. Set the field separator to ": ##" to parse lines for make targets.
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,9 @@ ifneq ($(HOST),dom0)
 	exit 1
 endif
 
-all: assert-dom0 validate clean sd-whonix sd-svs sd-gpg sd-journalist sd-decrypt \
-     sd-svs-disp ## Builds and provisions all VMs required for testing workstation
+all: assert-dom0 validate clean prep-whonix sd-whonix sd-svs sd-gpg \
+	sd-journalist sd-decrypt sd-svs-disp
+	## Builds and provisions all VMs required for testing workstation
 
 proj-tar: assert-dom0 ## Create tarball from "work" VM for export to dom0
 	qvm-run --pass-io $(DEVVM) 'tar -c -C $(dir $(DEVDIR)) $(notdir $(DEVDIR))' > ./sd-proj.tar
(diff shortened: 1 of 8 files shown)


conorsch:
Ran into some trouble with the test plan; after running `make all`:

```
ID: sd-whonix
Function: qvm-vm
Result: False
    require: 
        pkg: template-whonix-ws
```

The same error message appears on both `master` and this feature branch, so it appears we have some breakage that warrants a separate issue. Right now, it appears I'm blocked on reviewing the changes here, since I can't get the provisioning logic to complete.

Will poke more and see if any fixes in-line can be appended here to resolve; the need for such fixes is tracked in #122.

emkll:
Moving to whonix-14 fixes the issue @conorsch observed, but introduces a regression where `/etc/tor/torrc` or `/usr/local/etc/torrc.d/50_user.conf` do not get populated by `sd-whonix-hidserv-key.sls`

conorsch:
After a reboot, I'm seeing notification spam in dom0 related to the apparmor changes to whonix:

whonix-apparmor-spam

In light of #122, I'll note that I applied the apparmor changes described here to the _old_ whonix vm, running Jessie. Adjusting the template to use `whonix-gw-14` instead did not resolve the problem. Worse, it appears I've broken whonix entirely on my Qubes system with the template change.

@emkll can you recommend remediation steps to get whonix working on my system again? Might be best to put that discussion in #122. These tickets are going to be closely related either way.

emkll:
Thanks for the report @conorsch , sorry for breaking your computer. Prior to submitting, I had tested the pr for both Whonix versions 13 and 14, and I have not seen the errors you are experiencing. I still am struggling with idempotency and machine state, as I am currently using Qubes for daily usage as well. If Qubes permits template installs (and use of community repos) within salt, we should consider doing so in the medium term.

I think the best way forward would be to have a single whonix-gw template (14, as 13 is EOL), and two whonix-ws based AppVMs based on this template (`sys-whonix` [*] and `sd-whonix`). Given the level of breakage that occurred, I think it should be addressed as part of this PR. At a high level, a new sys-whonix vm AppVM must be created, based on the whonix-gw-14 template. Renaming the sys-whonix  in-place has been unreliable for me, here are the steps I took (which can certainly be optimized):
1. Create `sys-whonix-14` AppVM (provides internet) via template `qubes-template-whonix-gw-14`
2. For any domain that uses `sys-whonix` , set netvm to `sys-whonix-14`
3. Delete `sys-whonix`
4. Delete version 13 of the `qubes-template-whonix-ws` in dom0 as it is dom0-provisioned and cannot be deleted via the Qubes Manager (`rpm -qa | grep qubes-template-whonix` and removing `qubes-template-whonix-{ws,gw}` that are not version 14)
5. Rename `sys-whonix-14` to `sys-whonix` : In my testing, the rename basically acted as a clone, and I had to change all `sys-whonix-14` references back to `sys-whonix` and delete `sys-whonix-14`

[*] The reason we should maintain/preserve `sys-whonix` is that it's the default way Whonix retrieves updates (Template VMs go over Tor by default), using `sys-whonix` by default (as noted by the policy found `dom0`: `/etc/qubes-rpc/policy/qubes.UpdatesProxy`)

conorsch:
Pleased to report that @emkll's remediation steps have indeed resolved the whonix-related problems I encountered while reviewing this PR. I'll note that I _also_ had to run:

```
sudo qubes-dom0-update --enablerepo=qubes-templates-community --action=reinstall qubes-template-whonix-ws-14
```

Thereafter, `make all` in dom0 on this branch passes without error for me. `make test`, however, fails dude to the salt config drift that @emkll reported above:

> a regression where `/etc/tor/torrc` or `/usr/local/etc/torrc.d/50_user.conf` do not get populated by `sd-whonix-hidserv-key.sls`

We should fix that issue here, then we're good for merge.

conorsch:
Latest changes resolve the whonix config issues. Now able to run `make all` and `make test` in dom0 without issue. @emkll, if you can confirm successful access to Journalist Interface via `sd-journalist` using these latest changes, we should be good for merge.

kushaldas:
`make all` passed on this branch. I followed the steps given to have a version 14 of the whonix vm.

`make tests` fail for the `test_dispvm` as in Python2 there is not `qubes.tests` module, it is only available for Python3 in my `dom0`.

Just open python interpreter, and then try to `import qubes.tests` to verify.

kushaldas:
It seems we only have `python3` version of the tests module coming from the `qubes-core-dom0` package.

joshuathayer:
`make all` and `make test` work for me, `aa-status` on `sd-whonix` shows apparmor is installed. my existing whonix-based VMs seem to work fine without jumping through further hoops (ie, without upgrading sys-whonix)

so, looks good from my perspective!

redshiftzero:
Merging this, issues with python 2 vs python 3 can be tackled in a followup

PR #129, opened 2018-08-23
Update Fedora and Whonix as part of make all

### LINKED ISSUE #126: Run template update actions as part of "make all"
We currently have two provisioning targets in the Makefile used to update EOL templates to maintained versions:

* `make update-fedora-templates`
* `make update-whonix-templates` (#124)

These are not run automatically as part of `make all`, but probably should be. We'll need to make sure that the scripts invoked are idempotent; at a glance, the conditional logic should be sufficient.

joshuathayer:
Small change to include `update-fedora-templates` and `update-whonix-templates` targets as part of the `all` make target.

Closes #126

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-DEVVM=work
+DEVVM=securedrop-dev
 DEVDIR=/home/user/projects/securedrop-workstation # important: no trailing slash
 HOST=$(shell hostname)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,4 +1,4 @@
-DEVVM=work
+DEVVM=securedrop-dev
```
conorsch: I've been using `sd-dev`, which means I need to edit the Makefile every time I run the `make clone` action from dom0. Fine with that workflow for now, just noting the target VM change in case it affects others.

redshiftzero: hmm, probably should keep it as `work` (e.g. we have comments in a few places in the README referencing the example `work` AppVM)

joshuathayer: Oops, that should not have snuck in there- I'll move it back to `work`.

I've taken to editing `DEVVM` on my development VM so `make clone` works in dom0 consistently. I wonder if a different workflow would make more sense- setting an environment var maybe?

conorsch: The env var approach (for both `DEVVM` and `DEVDIR`) would satisfy my needs; we could port to a small script in `scripts/` and preserve the defaults there via e.g. `devvm="${DEVVM:-work}"`. 

Zooming out a bit, the whole take-a-lap-through-dom0 wouldn't be necessary if we configured the admin API to let the dev VM, whatever it is, create other VMs—but that's best discussed in a separate issue.

PR #142, opened 2018-09-06
Slightly friendlier dev env

### LINKED ISSUE #134: Clean up dev scripting output
The current Makefile-based provisioning workflow for development has a lot of output; the majority of info printed to stdout during a normal dev sessions is not actionable. A few suggested targets for cleanup:

* [x] silence `make clone` tar command
* [x] silence `make prep-salt` target
* [x] quieter VM destroy (currently lots of stderr output on a normal `make clean`, even if VMs don't exist)
* [x] The `make clone` command should also support environment variables, to support development machines that are not named `work`, as discussed in #124. 

Most of the changes in the brief wish list above I already store in patches locally, and I understand at least @joshuathayer does the same for the devvm name override. Let's get those patches in the repo to make iterating on the workstation more pleasant.

conorsch:
Closes #134. Implements a few comfort foods in dev env, making output more actionable, for the most part. 

@joshuathayer Please take a look; in particular, the env var for determining dev VM name should be useful to everyone.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,32 +62,25 @@ prep-salt: assert-dom0 ## Configures Salt layout for SD workstation VMs
 #sudo cp -r sd-svs-disp /srv/salt/sd  # nothing there yet...
 
 remove-sd-whonix: assert-dom0 ## Destroys SD Whonix VM
-	-qvm-kill sd-whonix
-	-qvm-remove -f sd-whonix
+	@./scripts/destroy-vm sd-whonix
 
 remove-sd-svs-disp: assert-dom0 ## Destroys SD Submission reading VM
-	-qvm-kill sd-svs-disp
-	-qvm-remove -f sd-svs-disp
+	@./scripts/destroy-vm sd-svs-disp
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -140,6 +126,12 @@ prep-whonix: ## enables apparmor on whonix-ws-14 and whonix-gw-14
 	qvm-prefs -s whonix-gw-14 kernelopts "nopat apparmor=1 security=apparmor"
 	qvm-prefs -s whonix-ws-14 kernelopts "nopat apparmor=1 security=apparmor"
 
+list-vms: ## Prints all Qubes VMs managed by Workstation salt config
+	@./scripts/list-vms
+
+destroy-all: ## Destroys all VMs managed by Workstation salt config
+	@./scripts/list-vms | xargs ./scripts/destroy-vm
```
conorsch: These two targets are far from necessary, but they strike me as super convenient. So problems with removing if they strike you as cruft, @joshuathayer.

conorsch:
Thanks, @joshuathayer !

PR #144, opened 2018-09-11
Migrate all scripts to Python 3

### LINKED ISSUE #128: Update all scripts to use Python 3
We're currently using Python 2 in this repository - given the rapidly approaching EOL of Python 2, we should upgrade to Python 3 and continue further development on Python 3

joshuathayer:
This contains the minor changes necessary to our existing scripts for them to run using Python 3 rather than Python 2. Mostly that involves using the new `print` syntax and changing the hashbang lines.

Closes #128

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,17 +2,11 @@
 version: 2
 jobs:
   lint:
-    docker:
-      - image: debian:stretch
+    machine: true
     steps:
       - checkout
-      - setup_remote_docker
-      # Check version output for debugging
-      - run: apt-get update
-      - run: apt-get install -y python-pip make
-      - run: pip install flake8
-      - run: python --version
-      - run: flake8 --version
+      - run: sudo apt-get update
+      - run: sudo apt-get install -qq make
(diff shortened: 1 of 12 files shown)


joshuathayer:
circleci is failing because it's running flake8 in a python2 environment, so we have a chicken/egg problem: if we update our circleci configuration to run under python3, other PRs will fail CI. But until we do that update, this PR will fail CI. :thinking:

conorsch:
Should have seen that one coming, @joshuathayer. Let's update the CI in this branch to switch the flake8 logic to python3, then all other active PRs remain passing. We have work elsewhere that we can port here, will share a patch...

conorsch:
@joshuathayer Fixed the CircleCI config here. It now runs python3 flake8, and it's failing. The setup should work for you locally, as well, so you can `make flake8` and get feedback on what still needs to be fixed. Apologies that the new target is slow; I cut corners and installed the requisite packages via the entrypoint script, so there's no caching to take advantage of on repeated runs. Still, hopefully you're unblocked now!

PR #149, opened 2018-09-13
Fixes GUI feedback in sd-journalist

conorsch:
Installs python-qt4 in whonix-ws-14

The `python-qt4` package is required for the GUI feedback integration,
so status messages can be displayed to the user. The package isn't
present by default, and we need it in sd-journalist. Since we don't
have SD TemplateVMs yet, we're adding the package to the base template,
installing in `whonix-ws-14`, rather than `sd-journalist`, to get
persistence for the package.

Marking as WIP, because we still have another follow-up to fix the submission process pipeline. See https://github.com/freedomofpress/securedrop-workstation/issues/127#issuecomment-421124029 for details; I propose we append a resolving commit here, then use the following test plan.

### Testing
1. Run `make all` in `dom0`.
2. Run `qvm-shutdown --wait whonix-ws-14`, to ensure the modified TemplateVM will be used on subsequent creation.
3. Re-run `make all` in `dom0`.
4. Run `make test` and confirm no errors.
5. Browse to configured Journalist Interface in the `sd-journalist` VM.
6. Download submission.
7. Confirm submission is processed successfully, and you can view the decrypted content of the submission.

PR #150, opened 2018-09-13
Prefer plain yaml in salt configs

joshuathayer:
This builds on @charles-boyd 's #141 , catching it up to changes in `master` and fixing a small issue.

From #141:

> These changes address #130 by removing most of the Jinja2 template syntax from the Salt configs by using qvm.vm instead.

Closes #130

Issue #138, opened 2018-09-05
Convert sd-journalist to sd-proxy


conorsch:
Porting the discussion from https://github.com/freedomofpress/securedrop-workstation/issues/107#issuecomment-409779664:

> We confirmed that a new `sd-proxy` VM (which takes the place of the `sd-journalist` VM) should expose a qrexec service called by `sd-svs`.

It's not just a nomenclature change, it's a full tooling swap. We'll need shake out the [sd-proxy](https://github.com/freedomofpress/securedrop-proxy) config (and packaging story) separately, which will unblock this work.

eloquence:
Edited top-level task since #127 was resolved a while ago. To recap the key blocking tasks are #156, #131 (will be closed by #160) and #161.

PR #151, opened 2018-09-14
Remove sd-decrypt vms

### LINKED ISSUE #121: Remove sd-decrypt AppVM
In the current incarnation of the workstation, the decrypt AppVM doesn't seem to be necessary. Instead, we can allow `qubes.Gpg` from `sd-svs`, which is where the UI will run and it is not network-connected. Note that the submission key will still be stored in vault VM `sd-gpg`. The advantages here are: simplify architecture and reduce memory consumption. Is there a reason why the `sd-decrypt` AppVM should be kept? (i.e. which threats are mitigated by `sd-decrypt`)

emkll:
Fixes #121

Minimal changes introduced to move sd-decrypt logic in to the sd-svs vm.
Removed config-dvm as it appeared deprecated and no longer in use.
Removed accept-sd-xfer-extracted mime handlers as logic is now on same vm that provides decryption.

Run make clean *before* checking out this branch, as sd-decrypt machines will not be removed from the workstation.

Test plan: 
- [x] #149 is merged into master
- [x] Workstation was provisioned as described in #149 to ensure python-qt in installed in journalist VM
- [x] make test passes
- [ ] Submissions are properly decrypted and stored in SVS
- [x] Qubes Manager does not have an sd-decrypt vm listed
- [ ] Updated data flow diagram is accurate
- [x] temp files are properly deleted from `/tmp`


sd-svs/decrypt-sd-submission
```diff
@@ -137,11 +137,5 @@ fh.close()
 
 shutil.rmtree(tmpdir)
```
redshiftzero: do we need the lines above starting in `fh = tempfile.NamedTemporaryFile(suffix=".sd-xfer-extracted", delete=False)`? looks like no since we have a directory with the decrypted files in a folder in `/tmp` which we can just directly move into place in `~/Sources` (previously we needed this logic as we had to ship off a directory of files/messages to another VM)

sd-svs/decrypt-sd-submission
```diff
@@ -137,11 +137,5 @@ fh.close()
...
-# finally! ship this off to sd-svs
-subprocess.call(["qvm-open-in-vm", "sd-svs", fh.name])
-os.unlink(fh.name)
-
-# work around a qubes open-in-vm bug
-# (https://groups.google.com/forum/#!topic/qubes-users/dxFa4ZXmPYc)
-# shouldn't be an issue for disp vm's
-os.unlink(input)
+# finally! ship this off to the next stage to store in SVS
+subprocess.call(["python3", "/usr/local/bin/accept-sd-xfer-extracted", fh.name])  # noqa:E501
```
redshiftzero: if my comment above is accurate, then the logic in `/usr/local/bin/accept-sd-xfer-extracted` should not be necessary (once we move the folder of docs/messages in place in `~/Sources`, then we can `send_progress` at the end of this script)

sd-svs/decrypt-sd-submission
emkll: great points @redshiftzero . I have added a commit to address the issue raised and updated the diagrams,

conorsch:
Requires rebase, given merge of #150. Should be straightforward: #150 changed files that this PR removes.

conorsch:
Ran through the test plan with the latest round of changes, including the commits addressing @redshiftzero's comments earlier:

- [x] #149 is merged into master
- [x] Workstation was provisioned as described in #149 to ensure python-qt in installed in journalist VM
- [x] make test passes
- [x] Submissions are properly decrypted and stored in SVS
- [x] Qubes Manager does not have an sd-decrypt vm listed
- [x] Updated data flow diagram is accurate
- :red_circle:  temp files are properly deleted from `/tmp`

Running `ls -l /tmp` on `sd-journalist` shows that I've still got temp files lingering there. The file as reported on my test machine is: `/tmp/mozilla_user0/all--2018-09-21--20-49-03.zip`.

Have not investigated why that file isn't being cleaned up. Requesting changes to address that last bit, then we're good.

Also want to flag that during `make all`, I saw several provisioning failures related to the `apt-get install -qq python-futures` command. I removed that line from `dom0/sd-journalist.sls` while reviewing. Not requesting changes on that front, because this PR didn't introduce those lines (#149 did).

conorsch:
Upon closer inspection, the test plan comment: 

>  temp files are properly deleted from `/tmp`

referred to `sd-svs`, not `sd-journalist`. Can indeed confirm that no temp files linger on `sd-svs`, as intended. While there _are_ some temp files on sd-journalist, it's the encrypted bundle downloaded from the Journalist Interface. Given that sd-journalist is slated for removal anyway (#138), not going to block merge on that improvement. The changes here did not introduce a regression regarding temp files.

Issue #158, opened 2018-10-04
Update mimetype handling for Debian Stretch


conorsch:
The submission handling logic relies on custom mimetype associations in order to ensure that submissions are never opened in the `sd-svs` AppVM (where the SecureDrop Client runs), but rather opens in a network-less DispVM. Once inside the DispVM, mimetypes are also important: the proper application must be selected to ensure a positive user experience. For example, we want to ensure that plaintext submission are opened in `gedit`, not `nano` or `vim`. 

Use of `xdg-open` _should_ be sufficient to ensure proper handling of mimetypes, but some additional dependencies may be required:

* `gvfs-bin` and `libgnomevfs2-bin` (to ensure `xdg-mime` operates correctly)
* `nautilus` should be present in the `sd-svs` VM (but isn't included by default in the `debian-9` template); we can revisit after #138 is implemented

We also want to ensure that system paths such as `/usr/share/applications/mimeapps.list` are not vulnerable to clobbering by upstream changes. It's also an option to declare the mimetype associations in `~/.config/mimeapps.list`, but using a non-system path will complicate both packaging logic and the use of TemplateVMs (since AppVMs receive their own $HOME by default, not copied from the base TemplateVM).

See the discussion thread in https://github.com/freedomofpress/securedrop-workstation/pull/155#issuecomment-426804059 for reference.

conorsch:
While testing #160, noticed that JPG images open in a DispVM in Firefox. That's not what we want (although it does display the image). Let's force an appropriate handler for that type.

kushaldas:
>`/usr/share/applications/mimeapps.list`

Just now verified that on Debian Stretch that file does not exists.

## How to test?

```
sudo apt-get install apt-file
sudo apt-file update
sudo apt-file find /usr/share/applications/mimeapps.list
```

I even checked this:

```
~$ sudo apt-file find /usr/share/applications/ | grep list
cinnamon-desktop-data: /usr/share/applications/x-cinnamon-mimeapps.list
gnome-session-common: /usr/share/applications/gnome-mimeapps.list
kde-telepathy-contact-list: /usr/share/applications/org.kde.ktpcontactlist.desktop
knowthelist: /usr/share/applications/knowthelist.desktop
```

@conorsch We are good to go :)

Issue #156, opened 2018-10-04
Create securedrop-workstation TemplateVM


conorsch:
All AppVMs used by the Workstation should be based on Debian 9, along with a grsecurity-patched kernel built specifically for use in Qubes. We can call this template `securedrop-workstation-template`, and use it as the base for all VMs. A brief checklist of desired attributes:

* [x] Derived from `debian-9` template
* [x] Runs grsecurity-patched kernel
* [x] Virt mode set to HVM (required to run custom kernel)
* [x] Installs updates via FPF-controlled repository (apt-test-qubes)

The MVP implementation can use the existing config scripts to add the FPF apt-test repo pubkey and pull in the grsecurity-patched kernel. Ultimately, we should consider hosting an RPM to be installed in dom0, so that development environments for the Workstation are not contaminated by customizations to the `debian-9` template, which was discovered to complicate partial review of #155.

joshuathayer:
To satisfy this issue, do we intend to use Salt to build and configure the template (before building RPMs)?

emkll:
@joshuathayer : Some of this work has already be done, using the qubes-builder in https://github.com/freedomofpress/qubes-template-securedrop-workstation, which is very heavily inspired by the way Whonix builds their templates: https://github.com/adrelanos/qubes-template-whonix . At a high level, it works by building the Debian-9 template, and running script prior to packaging that installs packages and configures the kernel. You can try building your own template locally by following the Building the Templates section in the readme of this repo.

In those templates, 2 debian packages should be installed, both from an fpf-controlled apt-repo, so that they can be updated:
`securedrop-workstation-grsec`: Which is a metapackage that will pull in the latest workstation kernel (this is already done - https://github.com/freedomofpress/securedrop-workstation/tree/master/builder/packages/securedrop-workstation-grsec/debian)
`securedrop-workstation-config`: Pulls in the most basic configurations or dependencies for the base template. (wip here: https://github.com/freedomofpress/securedrop-debian-packaging/tree/securedrop-workstation-config-metapackage)

Once templates are baked with those two packages, they can be reliably updated via apt and the packages we control (kernel and config).

Since most of that I propose we use the above as a base template for all workstation VMs, and use salt to build and configure other templates during the install phase of the workstation. Does this approach make sense?

ninavizz:
Question: is it the goal of this Qube to ONLY run the SecureDrop client (as a GUI element accessible to users)?

eloquence:
[...]

Issue #135, opened 2018-08-30
Tag all Salt-managed VMs with "securedrop-workstation"


conorsch:
There are a few "target-all-VMs" operations, e.g. destroy-all, create-all. Right now we're explicitly listing each VM referenced in the Salt config for use in these contexts, which creates a fair amount of duplication.

Let's tag all VMs with at least "securedrop-workstation" (in addition to any other more specific tags). Doing so would allow us to refer to the set of all VMs for programmatic actions such as base config, or destroy actions.

redshiftzero:
done in 7b7b1a1

PR #160, opened 2018-10-13
Implements Debian9-based custom templates for all AppVMs

### LINKED ISSUE #131: Achieve persistent configs via custom TemplateVMs

conorsch:
Converts all the AppVMs to reference a custom TemplateVM caused `sd-workstation-template`, which is cloned from the system `debian-9` TemplateVM. Summary of changes:

* All AppVMs are now based on `<app_vm_name>-template`, with a single exception: `sd-gpg` has no corresponding `sd-gpg-template`, because its customizations are simply a keyring that lives in $HOME, and therefore would not be inherited from the TemplateVM.
* Creates `sd-workstation-template` as a base TemplateVM for use across the multiple AppVM configurations. Does _not_ include much of the hardening documented in #156, we can follow up after these changes are merged.
* Script paths have changed throughout to convert `/usr/local/` to `/usr/` equivalents. Qubes handles `/usr/local/` directories specially, and allows them to be unique per VM.
* Similarly, use of `/home/user/` paths is now discouraged, since AppVMs don't inherit those from TemplateVMs. Wherever possible, appropriate system paths are now used. 
* VM config tests expanded somewhat to confirm new template settings

### Testing
The test plan remains the same as usual:

3. Run **Qubes Manager** and visually confirm all the `sd-` VMs are running Debian 9 (or Whonix 14, which is itself based on Debian 9). 
4. Confirm you can download and view a test submission. A plaintext submission should be displayed in `gedit` _in a DisposableVM_. 

As usual, 4 is the real clincher here. Without any out-of-band modifications, you should be able to view a plaintext submission. Have not tried other file formats yet; we should write up a test plan to account for those.

Closes #131. Related to #156 and #158.


conorsch:
Good catch, @emkll! While working on this I discovered that `decrypt-sd-submission` was not ported to Python3 along with the others in #144. We should be able to solve that here.

emkll:
`make clean` target does not clean all workstation related templates likely due to `scripts/list-vms` not matching all the templates. After running `make clean`, the following templates are still available on my machine:
* sd-whonix-template
* sd-svs-template
* sd-svs-disp-template

conorsch:
Both concerns are addressed, @emkll, please have another look. There was a single change required for the Python 3 port (file-writing bytes-vs-strings, standard fare for 2 to 3 porting), and I updated the destroy logic to be hardcoded. In #135 we can make that logic dynamic via tags, let's keep it simple for now.

Confirmed opening plaintext files (in gedit) and images (which opened in Firefox; flagged in #158 for follow-up).

conorsch:
> A note that with these changes, make all took almost 30 minutes in my local testing, which is significantly more than with previous iterations of the provisioning code, and that this doesn't include updates to the templates

Yes, it's taking a long time. Shortly I believe we should omit the "clean" step from `make all`, and instead treat `make all` as a provision action. Testing instructions can still use `make clean && make all`, but treating these machines as a bit more permanent within the dev env will help us make progress as we integrate the packages.

PR #163, opened 2018-10-17
Get builder dependencies after retrieving sources

emkll:
Sources and builder plugins used in making the template require different dependencies, thus the
`make install-deps` should be moved after `make get-sources`.

Test:
* Create new fedora-28 based appvm
* Increase private storage size (20GB is good)
* git clone this repo and checkout this branch
* cd into `builder/` and run `build-workstation-template` script


joshuathayer:
This looks great. I build a VM based on this template and confirmed it at least starts and is able to run a terminal, but didn't poke around much.

:ship: !

Issue #157, opened 2018-10-04
Host securedrop-workstation RPM for installation in dom0


conorsch:
During partial review of #155, we discovered that various developers have divergence in their `debian-9` templates, which caused different behaviors of crucial features such as mimetype handling. Our manual testing should not be subject to variation caused by local customizatoins. 

The work by @emkll in #115 provides us with an RPM of the workstation template. We should host this RPM in a yum repo, so that it's available for installation in `dom0`. Doing so will ensure a stable baseline for development efforts, and force a clean configuration for use in prod installs.

While the build process for the RPM is a bit onerous, we are _not_ required to update the template RPM in order to ship changes: we can pull in package updates via an FPF-controlled apt repository. We should still periodically update the RPM to avoid package drift and minimize setup and testing time, but most release-related changes will likely occur via deb packages.

This work should come after the work described in #156.

conorsch:
Discussed the next steps here with @msheiny on a call; summarized:

1. RPMs will be hosted in the existing S3 "dev" bucket (see here for example https://github.com/freedomofpress/securedrop-debian-packaging/blob/893cae8f109c3ec913af02f4bb827ed9225c4582/wheelsurls.txt)
2. We'll use a subdir inside that dev bucket, called e.g. `rpms`. We'll likely want a nested structure, e.g. `rpms/dom0/archive` and `rpms/dom0/repo`.
3. Signed RPMs (#192 for details on signing) will be uploaded to the "archive" subdir. 
4. Scripts will be used to generate the repo metadata in the "repo" subdir; the RPM will be copied over. (We may have to reupload the RPM to get it into the repo; that's not ideal, since the template RPM is several GB, but acceptable for alpha.)

Step 4 requires a bit of scripting and tooling. @msheiny has thoughts on that, so he'll take lead on implementing. Then @emkll / myself can review that work to upload RPMs can thereby close this issue.

Issue #161, opened 2018-10-15
Install securedrop-proxy package in template for sd-journalist AppVM


redshiftzero:
The initial version of `securedrop-proxy` is here: https://apt-test-qubes.freedom.press/pool/main/s/securedrop-proxy/ 

We need to: 

1. Install this package in  the template for `sd-journalist` (Related: renaming `sd-journalist` AppVM to `sd-proxy` (#138))
2. Create `config.yaml` on disk in the sd-journalist AppVM, which contains e.g. the onion URL of the journalist interface 
3. Make Qubes RPC policy changes in `dom0` needed for the proxy's operation as described [here](https://github.com/freedomofpress/securedrop-proxy#qubes-integration)

conorsch:
As discussed during sprint planning today, these changes here should be _additions_ to the `sd-journalist` config, not replacements. The goal is to preserve a working `master` branch on this repo, and until the client code is packaged and installed, we'll still need to use Tor Browser to download submissions for testing.

conorsch:
@redshiftzero Can you provide an example YAML configuration? I'm assuming the format is this: https://github.com/freedomofpress/securedrop-proxy/blob/9d040ce240fb837f34756c65e9f97c938a7c9e7f/config-example.yaml

What should `target_vm` be, for example?

conorsch:
Great, thanks, that's clear. Is there a specific filepath that the proxy app expects to find this file at? Reading the code in https://github.com/freedomofpress/securedrop-proxy/, it's not immediately obvious to me. Happy to shove it in `/etc/securedrop/sd-proxy-config.yml`, for instance, but want to make sure the app picks it up.

redshiftzero:
in the logic right now we have it at `/etc/sd-proxy.yaml` ([ref](https://github.com/freedomofpress/securedrop-proxy/blob/master/qubes/securedrop.Proxy#L1))

conorsch:
Perfect, just what I needed! Get back to ya soon with a candidate PR.

PR #169, opened 2018-10-18
Uses grsecurity-patched kernels for all templates

### LINKED ISSUE #156: Create securedrop-workstation TemplateVM

conorsch:
Closes #156. 

Installs FPF-maintained custom kernels in the workstation TemplateVMs. Includes updates to config tests to confirm grsec-patched kernels are running.

### Testing

3. Confirm you can download and view a test submission. A plaintext submission should be displayed in `gedit` _in a DisposableVM_. 

I also confirmed that a JPG submission was displayed successfully, but it first tried to open Firefox, failed, then tried and succeeded with `/usr/bin/display`. The error message regarding Firefox is ugly, but we should resolve that behavior as part of #158, not here. 

Note also that we intend to remove these custom apt logic for the VM config via #157 in the near future. This PR still allows us to start using custom kernels and shake out the issues as they arise.


conorsch:
Just flagging that this PR should be merged before the `securedrop-proxy` package inclusion (#161) changes land, reason being the apt repo logic is already presented here, and required for that implementation, as well. In the meantime, I'll branch from this work to keep going.

PR #185, opened 2018-11-01
updated README to reflect current install process

zenmonkeykstop:
The installation instructions are a little out of date. Looks like the project layout and Makefile changed a bit. Updating README to match current state.

(Docs-only, hence no test plan)


README.md
```diff
@@ -48,22 +48,28 @@ Next we need to do some SecureDrop-specific configuration:
...
 Once the configuration is done and this directory is copied to `dom0`, you must update existing Qubes templates and use `make` to handle all provisioning and configuration by your unprivileged user:
 
-    $ make update-fedora-templates
-    $ make update-whonix-templates
-    $ cd securedrop-workstation
-    $ make all
+```
+make update-fedora-templates
+make update-whonix-templates
+make all
```
conorsch: The `update-` targets are now a dependency of `make all`, so they can be omitted from the instructions here. That reminds me, we need to move that logic to Salt... will open issue.

zenmonkeykstop: updated!

Issue #147, opened 2018-09-13
Add tests for dom0 configuration


redshiftzero:
We have tests in the `tests/` directory that verify that the VMs have been provisioned as expected. What we don't have are tests verifying that `dom0` has been configured as expected, i.e. the contents of `/etc/qubes-rpc`. A nice addition to decrease risk of changes like #146 would be to add tests verifying the state in `dom0`.

redshiftzero:
Hey @charles-boyd if you're looking for tickets to pick up, this might be a good one!

redshiftzero:
tests added in #187, closing

PR #174, opened 2018-10-26
Packaging the configs into RPM

kushaldas:
Fixes #171.

We can now create a source tar ball using `python3 setup.py sdist` command.
And then use the given spec file to build a RPM.

## How to test?

* [ ] In `sd-dev` (or other dev AppVM): run `make dom0-rpm` to create the RPM package.
* [ ] Switch to dom0 for all remaining commands.
* [ ] Run `make clone` to fetch latest dir, including freshly built RPM
* [ ] Run `make clean` to remove any prior dev env state.
* [ ] Install the RPM: `sudo dnf install ./rpm-build/RPMS/noarch/securedrop-workstation-0.0.1-1.fc25.noarch.rpm`
* [ ] Run `make all`.
* [ ] Ensure no errors: `echo $?` should show `0`.
* [ ] Run `make test` and confirm no errors.

PR #187, opened 2018-11-02
Restrict and test and qubes-rpc policies for dom0

### LINKED ISSUE #175: Default deny qubes-rpc policies for sd-gpg and sd-svs
Qubes rpc policies have an 'ask' default policy:
`$anyvm $anyvm ask`

This means that an action with no explicit authorization will present a user with the following message:
image

This popup screen requesting authorization might confuse a user and might result in a file being copied, decrypted, or executed, due to user error.

For the VMs used in the context of the SecureDrop Workstation, we should ensure that the default policies for qubes-rpc commands are set to deny other vms. Specifically, qubes.Filecopy, qubes.Gpg, qubes.Openinvm, as well as any other policies that are not required for the functionality.

emkll:
## Status

Ready for review. 

Fixes #175, #147 
- Add tag for workstation provisioned templates and AppVM (`sd-workstation`)
- Create tests to ensure workstation-provisioned templates and AppVMs have label properly applied
- Consolidated qubes-rpc policies in `dom0/sd-dom0-qvm-rpc.sls`, to make it easier to manage and slightly more idempotent.
- Create tests to ensure workstation provisioned templates and AppVMs have deny-by-default and least privilege qubes-rpc policies, by way of `sd-workstation` label.

## Test plan
- [ ] `clean && make all` should succeed with no error
- [ ] `make test` should pass
- [ ] Editing the top of the rpc policies should cause tests to fail
- [ ] Ensure qubes-rpc policies in dom0-sdkfjdsf are sane
And, as always:
- [ ] Download, decrypt and view a plaintext submission in disposable VM
- [ ] Download, decrypt and view a submission (e.g., image) in disposable VM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,10 +10,14 @@ endif
 all: assert-dom0 validate clean update-fedora-templates			\
 	update-whonix-templates prep-whonix prep-dom0 sd-workstation-template \
 	sd-whonix sd-svs sd-gpg	\
-	sd-journalist sd-svs-disp
+	sd-journalist sd-svs-disp qubes-rpc
 
 clone: assert-dom0 ## Pulls the latest repo from work VM to dom0
 	@./scripts/clone-to-dom0
+
+qubes-rpc: prep-salt ## Places default deny qubes-rpc policies for sd-svs and sd-gpg
+	sudo qubesctl top.enable sd-dom0-qvm-rpc
+	sudo qubesctl --show-output --targets sd-dom0-qvm-rpc state.highstate
 
 sd-workstation-template: prep-salt ## Provisions base template for SDW AppVMs
 	sudo qubesctl top.enable sd-workstation-template
--- a/dom0/sd-dom0-qvm-rpc.sls
+++ b/dom0/sd-dom0-qvm-rpc.sls
@@ -1,0 +1,114 @@
+# -*- coding: utf-8 -*-
+# vim: set syntax=yaml ts=2 sw=2 sts=2 et :
+##
+# Explicitly deny as a catch-all for SecureDrop workstation provisioned VMs.
+# All SecureDrop-workstation provisioned VMS should have the sd-workstation tag.
+# To be both be mindful of developers using the workstation and ensure
+# RPC policies are not too permissive, this should be the first action
+# performed by the install. All other provisioning steps will prepend to this
+# list grants.
+# using blockreplace will ensure that we will be able to more reliably update
+# these policies during updates.
+##
+dom0-rpc-qubes.ClipboardPaste:
+  file.blockreplace:
+    - name: /etc/qubes-rpc/policy/qubes.ClipboardPaste
+    - prepend_if_not_found: True
+    - marker_start: "### BEGIN securedrop-workstation ###"
+    - marker_end: "### END securedrop-workstation ###"
+    - content: |
+        $anyvm sd-journalist ask
+        $anyvm $tag:sd-workstation deny
+dom0-rpc-qubes.FeaturesRequest:
+  file.blockreplace:
+    - name: /etc/qubes-rpc/policy/qubes.FeaturesRequest
+    - prepend_if_not_found: True
+    - marker_start: "### BEGIN securedrop-workstation ###"
+    - marker_end: "### END securedrop-workstation ###"
+    - content: |
+        $anyvm $tag:sd-workstation deny
+dom0-rpc-qubes.Filecopy:
+  file.blockreplace:
+    - name: /etc/qubes-rpc/policy/qubes.Filecopy
+    - prepend_if_not_found: True
+    - marker_start: "### BEGIN securedrop-workstation ###"
+    - marker_end: "### END securedrop-workstation ###"
+    - content: |
+        sd-journalist sd-svs allow
+        $anyvm $tag:sd-workstation deny
+dom0-rpc-qubes.OpenInVM:
+  file.blockreplace:
+    - name: /etc/qubes-rpc/policy/qubes.OpenInVM
+    - prepend_if_not_found: True
+    - marker_start: "### BEGIN securedrop-workstation ###"
+    - marker_end: "### END securedrop-workstation ###"
+    - content: |
+        sd-journalist sd-svs allow
+        $tag:sd-svs-disp-vm sd-svs allow
+        sd-svs $dispvm:sd-svs-disp allow
+        $anyvm $tag:sd-workstation deny
+dom0-rpc-qubes.OpenURL:
+  file.blockreplace:
+    - name: /etc/qubes-rpc/policy/qubes.OpenURL
+    - prepend_if_not_found: True
+    - marker_start: "### BEGIN securedrop-workstation ###"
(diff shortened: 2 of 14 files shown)


conorsch:
Tested and uncovered one error. Output from test failure:

```
python -m unittest discover tests
............Policy for OpenInVM is:
sd-journalist sd-decrypt allow
sd-journalist sd-svs allow
$tag:sd-svs-disp-vm sd-svs allow
sd-svs $dispvm:sd-svs-disp allow
$anyvm $tag:sd-workstation deny
sd-svs $dispvm:sd-svs-disp allow
## Note that policy parsing stops at the first match,
## so adding anything below "$anyvm $anyvm action" line will have no effect

## Please use a single # to start your custom comments

$anyvm	$dispvm	allow
$anyvm	$anyvm	ask

Policy for OpenInVM should be:
sd-journalist sd-svs allow
$tag:sd-svs-disp-vm sd-svs allow
sd-svs $dispvm:sd-svs-disp allow
$anyvm $tag:sd-workstation deny
F
.
.
.
.
....
.
.
.
.
..
======================================================================
FAIL: test_Policies (test_qubes_rpc.SD_Qubes_Rpc_Tests)
  File "/home/user/securedrop-workstation/tests/test_qubes_rpc.py", line 25, in test_Policies
    self.assertFalse(fail)
AssertionError: True is not false

Ran 27 tests in 209.334s

FAILED (failures=1)
```

The Qubes workstation I'm using was installed fresh about a week ago, so looks like the RPC policy files have had additional comments added. I've also been reviewing #174, so will re-run the test plan here to ensure I'm testing in a rather clean environment.

conorsch:
Was able to reproduce the error above, even after cleaning up my dom0 env of leftovers from reviewing #174. Ported the `OpenInVM` logic to use [Salt's blockreplace](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html#salt.states.file.blockreplace) module. With that change in place, the RPC policy was fixed on disk in dom0, and the tests began passing. I was able to add junk lines to the policy and re-provision, causing the tests to fail; reverting the junk lines caused the salt logic to "fix" the files again, with tests passing.

We should update all the policy adds to use similar logic, so we can confidently manage changes over time. As long as we keep the "markers" stable, we can insert or remove customizations at will. Some items of note for making the requested changes here:

* Use `|`, not `|-`, for YAML scalar syntax; we _want_ the trailing newline, so lines aren't concatenated
* Make sure the markers are identical throughout all config tasks
* Consider updating the RPC config tests (added in this PR) to hardcode the BEGIN/END markers; that should catch any typos in the future

Makefile
```diff
@@ -7,14 +7,18 @@ ifneq ($(HOST),dom0)
...
 	update-whonix-templates prep-whonix prep-dom0 sd-workstation-template \
 	sd-whonix sd-svs sd-gpg	\
 	sd-journalist sd-svs-disp
 
 clone: assert-dom0 ## Pulls the latest repo from work VM to dom0
 	@./scripts/clone-to-dom0
 
+qubes-rpc: prep-salt ## Places default deny qubes-rpc policies for sd-svs and sd-gpg
+	sudo qubesctl top.enable sd-dom0-qvm-rpc
+	sudo qubesctl --targets sd-dom0-qvm-rpc state.highstate
```
conorsch: Change

`sudo qubesctl --targets sd-dom0-qvm-rpc state.highstate`

to

`sudo qubesctl --show-output --targets sd-dom0-qvm-rpc state.highstate`

to match changes in #190.

dom0/sd-dom0-qvm-rpc.sls
```diff
@@ -0,0 +1,84 @@
...
+# Explicitly deny as a catch-all for SecureDrop workstation provisioned VMs.
+# All SecureDrop-workstation provisioned VMS should have the sd-workstation tag.
+# To be both be mindful of developers using the workstation and ensure
+# RPC policies are not too permissive, this should be the first action
+# performed by the install. All other provisioning steps will prepend to this
+# list grants.
+##
+
+dom0-rpc-qubes.ClipboardPaste:
+  file.prepend:
```
conorsch: Change `file.prepend` to `file.blockreplace` throughout; see the `OpenInVM` config task for an example.

dom0/sd-gpg.sls
```diff
@@ -17,3 +17,8 @@ sd-gpg:
       - label: purple
     - prefs:
       - netvm: ""
+
+# Add tag for default qubes-rpc deny catch-all rule
+# (see dom0/sd-dom0-qvm-rpc.sls)
+qvm-tags sd-gpg add sd-workstation:
+  cmd.run
```
conorsch: Convert the `cmd.run` calls to `qvm-tags` to a `- tags:  ` block on the VM spec; see the `- prefs:` block as an example. Docs here: https://github.com/QubesOS/qubes-mgmt-salt-dom0-qvm/blob/master/README.rst#qvm-vm Note that the syntax is slightly different when using the `qvm` wrapper lib; e.g.

```
  - tags:
      - add:
        - sd-workstation
```

tests/test_vms_exist.py
```diff
@@ -41,6 +41,7 @@ def test_sd_whonix_config(self):
         self.assertTrue(vm.template == "sd-whonix-template")
         self.assertTrue(vm.provides_network)
         self.assertFalse(vm.template_for_dispvms)
+        self.assertTrue('sd-workstation' in vm.tags)
```
conorsch: :ok_hand: Great addition; keeping these tests will help in confirming that the `cmd.run` -> `qvm.tags` conversion suggested above works well.

tests/test_qubes_rpc.py
```diff
@@ -0,0 +1,50 @@
...
+    def _startsWith(self, filename, expectedPolicy):
+        filePath = os.path.join(QUBES_POLICY_PREFIX + filename)
+        with io.open(filePath, 'r') as f:
+            actualPolicy = f.read()
+            if actualPolicy.startswith(expectedPolicy):
+                return True
+            else:
+                print("Policy for {} is:\n{}".format(filename,
+                                                     actualPolicy))
+                print("Policy for {} should be:\n{}".format(filename,
```
conorsch: Nit: more aggressive output would be helpful here for when the policies fail. Consider adding more newlines and header blocks like:

```
print("\n\n#### BEGIN RPC policy error report ####\n\n")
print("Policy for {} is:\n{}".format(filename,
                                                     actualPolicy))
print("Policy for {} should be:\n{}".format(filename,
                                                     expectedPolicy))
print("\n\n#### END RPC policy error report ####\n\n")
```

emkll:
@conorsch thanks for all the great feedback. I've incorporated it and cleaned up the git history, this should now be ready for re-review!

conorsch:
Roger that, re-reviewing!

conorsch:
Works like a charm. After full provision with passing tests, I edited one of the Salt files to write a bad config, and the tests started failing. Reverted the fix, reapplied, and tests are passing—so the new logic is correctly updating the files in-place, as we want.

PR #192, opened 2018-11-06
Add instructions for verifying integrity of sources and binaries

### LINKED ISSUE #37: Code signing / verification
As a Qubes Workstation user, I do not want to execute malicious code in `dom0`.

Provide a way for a user to validate the integrity of the `securedrop-workstation` code. This could be achieved through:

- Sign tags/releases with using GitHub
- Building (signed) packages

emkll:
Fixes #37 

Docs-only update which provides guidelines for verification of source code and built binaries (debs and rpms) for use in the SecureDrop workstation release process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,33 +356,41 @@ git checkout <tag>
 
 ### Signing binaries/packages
 
-#### Debs
+#### Debian packages
 Apt repository Release file will be signed, containing checksum of the debs.
 
-#### Rpms
-The entire RPM must be signed. This process also requires a Fedora machine/VM on which the GPG signing key (either in GPG keyring or in qubes-split-gpg) is setup. You will need to add the public key to RPM for verification (see below).
+#### RPM packages
+The entire RPM must be signed. This process also requires a Fedora machine/VM on which
+the GPG signing key (either in GPG keyring or in qubes-split-gpg) is setup.
+You will need to add the public key to RPM for verification (see below).
 
-`rpm -Kv` indicates if digests and sigs are OK. Before signature it should not return signature, and `rpm -qi <file>.rpm` will indicate an empty Signature field.
+`rpm -Kv` indicates if digests and sigs are OK. Before signature it should not return signature,
+and `rpm -qi <file>.rpm` will indicate an empty Signature field. Set up your environment:
 
-Set up environment:
-`sudo dnf install rpm-build rpm-sign`
-Set your vault vm in /rw/config/gpg-split-domain
-Edit ~/.rpmmacros with the following contents:
 ```
+sudo dnf install rpm-build rpm-sign  # install required packages
+echo "QUBES_GPG_DOMAIN=vault" | sudo tee /rw/config/gpg-split-domain  # edit 'vault' as required
+cat << EOF > ~/.rpmmacros
 %_signature gpg
 %_gpg_name <gpg_key_id>
 %__gpg /usr/bin/qubes-gpg-client-wrapper
 %__gpg_sign_cmd %{__gpg} --no-verbose --detach-sign %{__plaintext_filename} --output %{__signature_filename}
+EOF
+```
+Now we'll sign the RPM:
+
+```
+rpm --resign <rpm>.rpm  # --addsign would allow us to apply multiple signatures to the RPM
+rpm -qi<file.rpm>  # should now show that the file is signed
+rpm -Kv  # will complain that signature is not OK: "Digests SIGNATURES NOT OK"
+# This is because the the (public) key of the RPM signing key is not present,
+# and must be added to the RPM client config to verify the signature:
+sudo rpm --import <publicKey>.asc
+rpm -Kv  # will now say signatures are OK: "Digests signatures OK"
 ```
 
-Sign the rpm:
-`rpm --resign <rpm>.rpm` (--addsign would allow us to apply multiple signatures to the RPM)
-Verify that the rpm is signed:
-`rpm -qi<file.rpm>` should now show that the file is signed
-`rpm -Kv` will complain that signature is not ok (`Digests SIGNATURES NOT OK`)
-This is because the the (public) key of the RPM signing key is not present and must be added to the RPM client config to verify the signature:
-`sudo rpm --import <publicKey>.asc`
-`rpm -Kv` will now say signatures are ok (`Digests signatures OK`)
+You can then proceed with distributing the package, via the "test" or "prod" repo,
+as appropriate.
 
 ## Threat model
 
--- a/README.md
+++ b/README.md
@@ -374,7 +374,7 @@ cat << EOF > ~/.rpmmacros
 %_signature gpg
 %_gpg_name <gpg_key_id>
 %__gpg /usr/bin/qubes-gpg-client-wrapper
-%__gpg_sign_cmd %{__gpg} --no-verbose --detach-sign %{__plaintext_filename} --output %{__signature_filename}
+%__gpg_sign_cmd %{__gpg} --no-verbose -u %{_gpg_name} --detach-sign %{__plaintext_filename} --output %{__signature_filename}
 EOF
 ```
 Now we'll sign the RPM:
--- a/README.md
+++ b/README.md
@@ -335,16 +335,16 @@ make test
 
 SecureDrop Workstation code spans across the following repositories:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -200,6 +200,58 @@ pipenv shell
...
+## Signing sources
+
+SecureDrop Workstation code spans across the following repositories:
+
+https://github.com/freedomofpress/securedrop-client
+https://github.com/freedomofpress/securedrop-debian-packaging
+https://github.com/freedomofpress/securedrop-proxy
+https://github.com/freedomofpress/securedrop-sdk
+https://github.com/freedomofpress/securedrop-workstation
+https://github.com/freedomofpress/qubes-template-securedrop-workstation
```
conorsch: These URLs must be formatted as a bulleted list, otherwise Markdown will concatenate them into a paragraph.

README.md
```diff
@@ -200,6 +200,58 @@ pipenv shell
...
+
+
+### Release
+1. For each release, a tag will be signed in all the above repos.
```
conorsch: Ambiguously worded (perhaps intentionally so). What constitutes a "release" of the SecureDrop Workstation? We should be prepared to bump individual package versions, and release those versions, without bumping versions for _all_ packages. Assuming you agree, I suggest the following rewording:

> For each release of a package in the above repos, a signed tag will be pushed documenting the commit hash used to perform the build. 

We don't yet have a "parent" version of the SecureDrop Workstation, just various versions of the different components.

README.md
```diff
@@ -200,6 +200,58 @@ pipenv shell
...
+
+3. Metadata (e.g. commit hash for release) should be tracked inside the .deb (e.g.: `/usr/share/packagename/release-info.txt`)
+
+### Signing binaries/packages
+
+#### Debs
+Apt repository Release file will be signed, containing checksum of the debs.
+
+#### Rpms
+The entire RPM must be signed. This process also requires a Fedora machine/VM on which the GPG signing key (either in GPG keyring or in qubes-split-gpg) is setup.
```
conorsch: Drat. You're completely right, of course. In #174 we'll have a container to _build_ the RPM (even on a debian-9 host), but we'll still need to use a dedicated signing VM. That's fine, thanks for documenting it.

README.md
```diff
@@ -200,6 +200,58 @@ pipenv shell
...
+
+`rpm -Kv` indicates if digests and sigs are OK. Before signature it should not return signature, and `rpm -qi <file>.rpm` will indicate an empty Signature field.
+
+Set up environment:
+`sudo dnf install rpm-build /usr/bin/rpmsign`
```
conorsch: Double-check that command for me? Looks like `/usr/bin/rpmsign` is a path, rather than a package name, so it should either be removed, or changed to `rpmsign`.

README.md
```diff
@@ -200,6 +200,58 @@ pipenv shell
...
+Set your vault vm in /rw/config/gpg-split-domain
+Edit ~/.rpmmacros with the following contents:
+```
+%_signature gpg
+%_gpg_name <gpg_key_id>
+%__gpg /usr/bin/qubes-gpg-client-wrapper
+%__gpg_sign_cmd %{__gpg} qubes-gpg-client-wrapper --no-verbose --detach-sign %{__plaintext_filename} --output %{__signature_filename}
```
conorsch: Have not tested, but on L244, looks like `qubes-gpg-client-wrapper` is duplicated, given the `%{__gpg}` macro interpolation preceding it.

README.md
```diff
@@ -200,6 +200,58 @@ pipenv shell
...
+```
+
+Sign the rpm:
+`rpm --resign <rpm>.rpm` (--addsign would allow us to apply multiple signatures to the RPM)
+Verify that the rpm is signed:
+`rpm -qi<file.rpm>` should now show that the file is signed
+`rpm -Kv` will complain that signature is not ok (`Digests SIGNATURES NOT OK`)
+This is because the rpm key needs to be added locally:
+`sudo rpm --import <publicKey>.asc`
+`rpm -Kv` will now say signatures are ok (`Digests signatures OK`)
```
conorsch: Have not run these commands yet, but the phrase "the rpm key needs to be added locally" is confusing. Given that the `~/.rpmmacros` file contains an override to the gpg binary (according to definitions at http://rpm5.org/docs/api/config_macros.html), the keys should be available for verification, no? A bit more language here explaining the unintuitive behavior is warranted—particularly what "rpm key" refers to—or, better, we can hopefully resolve it.

The entire verify/sign workflow would perhaps best be provided as a script, though I recognize ferrying the RPM to be signed, as well as the signing logic, between AppVMs is going to be painful.

README.md
emkll: It was confusing to me at first, but not a typo, at least for me locally in fedora-28:
```
[user@my-new-fedora ~]$ sudo dnf install rpmsign
Fedora 28 - x86_64 - Updates                    2.7 MB/s |  26 MB     00:09    
Qubes OS Repository for VM (updates)            165 kB/s | 180 kB     00:01    
Last metadata expiration check: 0:00:00 ago on Tue 06 Nov 2018 01:54:16 PM EST.
No match for argument: rpmsign
Error: Unable to find a match
[user@my-new-fedora ~]$ sudo dnf install /usr/bin/rpmsign
Last metadata expiration check: 0:00:16 ago on Tue 06 Nov 2018 01:54:16 PM EST.
Dependencies resolved.
================================================================================
 Package                       Arch      Version               Repository  Size
================================================================================
Installing:
 rpm-sign                      x86_64    4.14.2.1-2.fc28       updates     72 k
Upgrading:
 python2-rpm                   x86_64    4.14.2.1-2.fc28       updates    120 k
 python3-rpm                   x86_64    4.14.2.1-2.fc28       updates    120 k
 rpm                           x86_64    4.14.2.1-2.fc28       updates    533 k
 rpm-build-libs                x86_64    4.14.2.1-2.fc28       updates    138 k
 rpm-libs                      x86_64    4.14.2.1-2.fc28       updates    331 k
 rpm-plugin-selinux            x86_64    4.14.2.1-2.fc28       updates     68 k
 rpm-plugin-systemd-inhibit    x86_64    4.14.2.1-2.fc28       updates     69 k
 rpm-sign-libs                 x86_64    4.14.2.1-2.fc28       updates     74 k

Transaction Summary
================================================================================
Install  1 Package
Upgrade  8 Packages

Total download size: 1.5 M
Is this ok [y/N]: 
```

But reading this again made me discover the package is called `rpm-sign`, thanks ! will change

README.md
emkll: Yes, ambiguous given that we haven't gone through the process, and I assume this is subject to change. The main idea here is that we want release to have signed tags for all it's components, and we need to (easily) know which specific commit hash has made its way into the release. 

We have two choices: 
* Create a tag for each version of the specific packages and store metadata in securedrop-packaging or elsewhere to know which package tag is included in which workstation release
*  Tag each securedrop release in the above repos. If `securedrop-proxy` is not updated, the tag for 0.1 and 0.2 point to the same commit hash. 

Tracking the tags in a build-release makefile target in securedrop-debian-packaging seems like the straightforward way.

README.md
```diff
@@ -200,6 +200,58 @@ pipenv shell
...
+### Release
+1. For each release, a tag for each release will be signed and pushed to each of the above repos.
+
+2. Create a Makefile target in securedrop-debian-packaging repo that contains release tags / commit hashes for each repository used for the release. To verify the tag signature and check out the packaging logic:
+```
+git tag -v <tag>
+git checkout <tag>
+```
+
+3. Metadata (e.g. commit hash for release) should be tracked inside the .deb (e.g.: `/usr/share/packagename/release-info.txt`)
```
kushaldas: After giving more thoughts on this during holiday, I think the better place can be the changelog of the debian package itself. As that will make sure the final things getting installed on the system are only from the tarball, and nothing dynamic there. We can add a check to verify this during final packaging CI (when we will have it).

conorsch:
Used these instructions recently during signing of test RPMs, and they worked splendidly. After merge of #248 (which also edits the README), will rebase and approve, so we can get these in.

conorsch:
Rebased on top of latest master (4633259624cbeb494299d00df6177d2a8e6f4b79), and addressed a few formatting tweaks. I ran through these instructions while preparing a candidate RPM as part of #157, and they worked swimmingly!

PR #204, opened 2018-11-10
Moves update-templates logic to Salt

### LINKED ISSUE #186: Port template update commands from makefile to Salt
There are two Makefile targets dedicated to ensuring dom0 packages are up to date:

```
$ make help | grep -i update
update-fedora-templates    Upgrade to Fedora 28 templates
update-whonix-templates    Upgrade to Whonix 14 templates
```

The changes specified in those scripts should be ported to Salt, so they'll be included in the dom0 configuration (#171). We should also add tests to ensure those options are set on the relevant VMs. 

Worth considering updating the default dispvm to `sd-svs-disp`, but that discussion can happen separately.

conorsch:
Let's not rely on the Makefile to pull in critical dependencies such as
Qubes template RPMs in dom0. We'll enforce those dependencies via Salt,
providing an update mechanism over time.

The AppArmor enabling via kernelopts is required on the sd-whonix and
sd-journalist VMs, which are cloned from the whonix-gw and whonix-ws
templates respectively, but we'll set it on the whonix workstation
templates as well, since those may be used by sd-research in the near
future.

Also culls the "prep-whonix" Makefile target, since that's been migrated
to Salt, as well.

Closes #186.


zenmonkeykstop:
Tested this by nuking the fedora-28 template on an existing Qubes install and running `make all` - fedora-28 template was reinstalled so :+1: for that.

Issue #154, opened 2018-09-26
Consider suppressing gpg permission dialog or increasing timeout


eloquence:
The workstation currently triggers a gpg key access permission dialog with a short timeout when decrypting data. Frequent dialogs like this are likely to cause alert fatigue -- users just click "Yes" without scrutinizing the context. In the event of malicious code execution, it's doubtful that the user would know _not_ to grant permission that one time.

![gpg dialog](https://raw.githubusercontent.com/freedomofpress/securedrop-workstation/master/docs/images/step3-decrypt-cropped.png)

We should consider increasing the timeout so that the dialog appears only once per boot, or suppressing it altogether, unless there are security benefits that are not clear. Note that this dialog provides no protection against workstation compromise, as it requires no further authorization.

zenmonkeykstop:
This definitely needs doing - if the recommended workflow is to log out and/or shut down the workstation between sessions, I would agree with making it as long as possible. But it should definitely be as longer than 5 mins. An hour strikes me as a good minimum value.

ninavizz:
@zenmonkeykstop Could we increase it from 1 hour to 8 or 12 hours? Only asking, because biggy files can take +1 hours to download, and updates may take a similar amount of time to process. Curious how this may impact the threat-model or if it would violate it. Thoughts/compromises @emkll or @redshiftzero?

conorsch:
Big fan of removing the prompt entirely, but not sure we have the ability to do so easily, given the splitgpg tooling. The [splitgpg docs](https://www.qubes-os.org/doc/split-gpg/#setting-up-the-gpg-backend-domain) provide recommendations on lengthening the window:

```
[user@work-gpg ~]$ echo "export QUBES_GPG_AUTOACCEPT=86400" >> ~/.bash_profile
```

Let's add the equivalent file management targeting `sd-gpg` to the Salt config; I believe @zenmonkeykstop is already working on this.

redshiftzero:
i definitely think we can make it 8 hours (same time as the API token timeout)

conorsch:
Resolved via #207.

PR #207, opened 2018-11-11
updates sd-gpg autoaccept timeout to 8 hours

zenmonkeykstop:
This addresses #154 by increasing the split GPG autaccept interval to 1 hour for the sd-gpg vault VM

In UX terms, this will mean that the user will be prompted with the gpg accept dialog once at the beginning of their session and then again every hour. General consensus seems to be that the dialog offers no security benefits, so the interval could even be bumped up higher. It doesn't look like it can be suppressed without changes in split-gpg.

### DIFF
--- a/dom0/sd-gpg-files.sls
+++ b/dom0/sd-gpg-files.sls
@@ -9,6 +9,10 @@
 #
 ##
 
+/home/user/.profile:
+  file.append:
+    - text: "export QUBES_GPG_AUTOACCEPT=28800"
+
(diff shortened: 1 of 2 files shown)


ninavizz:
Noted in Issue #154 that it'd be preferred to increase the time to 8hrs; Conor & Mikael & Jen thumbs-up'd the req! 💃

zenmonkeykstop:
Not a problem, PR will get some minor refactoring anyway. If nobody else
picks it up I'll make the change ASAP

PR #208, opened 2018-11-12
Uses .example suffix on testing submission key

conorsch:
Added to the "validate" step, same as with config.json, so developers
must copy a file to `sd-journalist.sec` in order for the provisioning to
succeed. Using the test key provided in the repo will work, or
developers can provide another submission key, used with a hardware
instance for interactive testing.

In the process, I updated the config.json to use ".example"
consistently, since it was ".sample" in some places.


conorsch:
Heads up @eloquence, there are minor README changes in here; pinging so that the incoming docs updates are based on these changes.

PR #210, opened 2018-11-13
Declare salt dependencies explicitly in provisioning

conorsch:
Uses proper "include" and "require" syntax in the Salt configs. We had been applying these directives inconsistently, and `--show-output` reveals complaining about inability to import, causing race conditions when the salt states are ordered (which happens automatically prior to run). 

Changes:
* Drops use of Makefile dependencies for "make all"; uses qubesctl directly, as with the securedrop-update logic
* Drops use of "clean" as automatic step in "make all"; machines are now configured as expected
* Removes `sd-whonix-template`; there were zero configurations of the template, so it's unnecessary. We've always added the HidServAuth line directly to `sd-whonix`, where it persists due to `/usr/local/` persistences in AppVMs.

The rebuild story is now vastly faster: on clean installs, 35m -> 15m. On rebuilds (where VMs already exist), 35m -> 6m.

### Testing
Use the normal test plan:

* `make clean`
* `make all`
* `make test`
* Decrypt and view a document in SecureDrop Client, ensuring it opens in a DispVM.

Try running `make all` again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,16 +129,16 @@ qubes-prefs default_dispvm fedora-28-dvm
 You can then delete the end-of-life fedora-26 template in dom0 by running:
 
 ```
-sudo yum remove qubes-template-fedora-26
+sudo dnf remove qubes-template-fedora-26
 ```
 
 #### Upgrading `sys-whonix` and `whonix-ws` AppVMs to Whonix 14
 
-Qubes 4.0 also ships with end-of-life Whonix templates (`whonix-gw` and `whonix-ws`).`sys-whonix` is used by `sd-whonix` to fetch updates, and should be upgraded. You should destroy `whonix-gw` from the Qube Manager and re-provision a new `sys-whonix` AppVM with the potion **provides network** based on `whonix-gw-14`. You will need to delete the `whonix-ws-dvm` and `anon-whonix` vmsYou can then remove the end-of-life templates:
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -85,7 +96,39 @@ make all
...
 The build process takes quite a while. You will be presented with a dialog asking how to connect to Tor: you should be able to select the default option and continue. You may wish to increase the scrollback in the dom0 terminal from 1000 (the default) to 100000, to ensure you can review any errors in the verbose output.
 
-**NOTE:** Due to [issue #202](https://github.com/freedomofpress/securedrop-workstation/issues/202), the installation may fail with disk quota errors. If this happens, reboot the entire workstation and run `make all` again.
+**NOTE:** Due to [issue #202](https://github.com/freedomofpress/securedrop-workstation/issues/202), the installation may fail with disk quota errors. If this happens, reboot the entire workstation and run `make all` again. The error will contain the following informating in your dom0 terminal:
+
+```
+qfile-agent : Fatal error: File copy: Disk quota exceeded; Last file: <...> (error type: Disk quota exceeded) '/usr/lib/qubes/qrexec-client-vm dom0 qubes.Receiveupdates /usr/lib/qubes/qfile-agent /var/lib/qubes/dom0-updates/packages/*.rpm' failed with exit code 1!
+```
+
+##### Upgrading `sys-net`, `sys-usb` and `sys-firewall` to fedora-28
```
redshiftzero: we also need to add instructions for updating whonix

PR #218, opened 2018-11-15
Fix fedora update check test

### LINKED ISSUE #217: test_all_fedora_vms_uptodate fails
Workstation test for checking if Fedora-28 template is update does not reliably report the presence (or not) of updates available. `dnf update-check doesn't` always return 0 when no updates are available. It returns 1 with the following error: 
```
error failed to synchronize cache for repo 'updates'
```

emkll:
Fixes #217 

Test plan:

- [x] Enure all your VMs are up to date (you can run `sudo securedrop-update`)
- [x] `make test` does not error out on `test_all_fedora_vms_uptodate`

### DIFF
--- a/tests/test_vms_platform.py
+++ b/tests/test_vms_platform.py
@@ -52,7 +52,7 @@ class SD_VM_Platform_Tests(unittest.TestCase):
             # so expect only that string.
             self.assertEqual(results, "Listing...")
         else:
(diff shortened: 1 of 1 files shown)


conorsch:
Updater's running for me right now, so I'll be in a fine position to test this shortly.

PR #228, opened 2018-11-16
Remove qubes.OpenInVM for sd-svs-disp

emkll:
I recommend we re-tag 0.1.0 based on this fix.

Test plan: 
- make all
- make test
- Ensure that submissions are decrypted and opened properly in dispvms
- No qubes-rpc denies in  journalctl


conorsch:
Agreed on 0.1.1 @redshiftzero ; and +1 to getting this in. Testing locally now.

redshiftzero:
this is a very good catch, re-provisioning in qubes...

Issue #241, opened 2019-02-08
Shipped securedrop-client incompatible with latest journalist API


deeplow:
## Description

When installing the securedrop-workstation, following the README.md and testing it with the latest securedrop journalist server version, the securedrop-client is not able to view the documents ("waiting to refresh" indefinitely)

## Steps to Reproduce

1. configure sercuredrop server (as detailed in the docs)
2. follow README.md
3. start securedrop-client in sd-svs
4. login with credentials

## Expected Behavior
A successful login and to be able to see the files shared with journalist

(the problem is not is not with the server, as suggested on gitter. Image below shows a successful login on the web interface)
![login-successful](https://user-images.githubusercontent.com/47065258/52505014-1798bd80-2be2-11e9-8beb-6be0c87cd964.png)

## Actual Behavior

It looks like the login is successful because it shows the button log out, but a message shows "waiting to refresh" indefinitely.

![waiting-to-refresh](https://user-images.githubusercontent.com/47065258/52505351-1d42d300-2be3-11e9-88b3-796f583dbb41.png)

## Comments
I was able to pinpoint the specific issue to a "Malformed authorization header." sent by the client. (Not that this matters, because it must have been fixed in the meantime)

The problem seems to stop when we copy from the source code to sd-svs and run that version (but that process is not documented and it took me a while to figure out - rpc-policies, enable networking, removing --no-proxy flag, etc.).

I ended up taking a few days to complete the whole installation process and to figure out a solution. But to avoid that people interested in the project also need to figure it out too I would suggest a warning in the section about the securedrop client saying that it might not work and it should instead be tested in the sd-dev machine (that's the easy solution that shouldn't take away the devs' time)

Another other solution would involve either making sure the master of this repo passes on the test with the journalist api server version of the upstream version (which I don't think it is much worth it in terms of time at this stage)

redshiftzero:
wow - big apologies for the confusion @deeplow. So this was issue https://github.com/freedomofpress/securedrop-client/issues/232 / https://github.com/freedomofpress/securedrop/issues/4064, an issue with the auth header as you correctly point out. 

We should probably cut another release of `securedrop-client` so that this doesn't continue to bite people, but until then, want to make a PR to note in the README that developers from qubes should set up the development environment instead of using the latest release?

conorsch:
[...]

PR #242, opened 2019-02-08
add note on securedrop client testing (issue #241)

deeplow:
Following the suggestion of @redshiftzero in #241, I've added a temporary note, telling people interested in testing the workstation to test the client in the development environment instead, while the a newer version of the client with the issue fixed is not shipped by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,7 @@ sudo securedrop-update
 
 #### Using the *SecureDrop Client*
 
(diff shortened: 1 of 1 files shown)


conorsch:
Rebased on top of latest master (57679a95c0b4bc798b5f753968160da38f3d86e7) to include CI fixes from #247.

PR #248, opened 2019-04-06
Updates docs and config for Qubes 4.0.1

### LINKED ISSUE #189: Fedora-26 still used for sys- VMs
On a clean install, the current config logic will leave the core `sys-*` VMs based on `fedora-26`, which has been EOL'd. We should explicitly reassign a dependency on the latest Fedora VM (currently fedora-28) for those machines.

There may be existing Qubes-maintained Salt logic we can reuse in `/srv/formulas/base/virtual-machines-formula/qvm`.

Related to #111, #181, #186.

### LINKED ISSUE #191: Whonix 13 templates still used for sys-whonix
On a clean install, the current provisioning logic still leaves `sys-whonix` configured to use `whonix-gw` as its template, which is EOL'd. We should update the logic to force the association on `whonix-gw-14`. 

Note that there is logic in place in `/srv/formulas/base/virtual-machines-formula/qvm/*whonix*` to update the associations already, so we should begin by analyzing that logic and determining why it's insufficient. It may be as simple as the "template" pref not being updated for running VMs (at least that's true when using `qvm-prefs -s <vm> template <new_template>` via CLI). 

Related: #186, #189.

### LINKED ISSUE #245: Updates for Qubes 4.0.1
Qubes 4.0.1 includes Whonix 14 and Fedora 29, so we should:

- [x] Run through workstation provisioning on Qubes 4.0.1 and update the README to indicate we should install on Qubes 4.0.1
- [x] Handle removal of jessie-backports repo - https://github.com/QubesOS/qubes-issues/issues/4917
- [x] Update `test_all_fedora_vms_uptodate ` [here](https://github.com/freedomofpress/securedrop-workstation/blob/5d7a5642b6a8675a7b247a755aa0401a0687be93/tests/test_vms_platform.py#L71) to point to fedora 29
- [x] Remove section in docs instructing users to remove EOL whonix templates: https://github.com/freedomofpress/securedrop-workstation#upgrading-sys-whonix-and-whonix-ws-appvms-to-whonix-14
- [x] Remove section in docs instructing users to upgrade to Fedora 28: https://github.com/freedomofpress/securedrop-workstation#upgrading-sys-net-sys-usb-and-sys-firewall-to-fedora-28

conorsch:
Ensures the docs are accurate as of Qubes 4.0.1. Ran through a clean install and fixed errors encountered during the provisioning process. Fortunately, much of the manual recommendations we'd made around EOL Fedora templates can now be removed from the README—although I expect we'll soon be forced to do the same for 29 -> 30 (Fedora-29 goes EOL ~ 2019-09). 

Closes #245.
Closes #189.
Closes #191.

### Testing
Ideally, perform a clean installation of Qubes 4.0.1 following these new docs! That's a large ask, though, so more practically, the following should work:

1. Check out this branch in `sd-dev` (or other AppVM you use for development)
2. Switch to dom0 for following commands
3. Run `make clean` to purge any old config.
4. Run `make all` to provision all VMs, and confirm no errors.
5. Run `make test` and confirm all passing.

Assuming those steps go smoothly, take a moment to run `qvm-ls | grep fedora` and make sure the _only_ references you see are to `fedora-29`. If you see other versions, you may want to clean up, e.g.

```
qvm-remove fedora-28-dvm
sudo dnf remove qubes-template-fedora-28
```

And similar for any other versions of Fedora VMs you may have.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,9 @@ While we are strongly committed to piloting the use of Qubes OS for SecureDrop, 
 
 Installing this project is involved. It requires an up-to-date Qubes 4.0 installation running on a machine with at least 12GB of RAM. You'll need access to a SecureDrop staging server as well.
 
-#### Qubes 4.0
+#### Qubes 4.0.1
 
-Before trying to use this project, install [Qubes 4.0](https://www.qubes-os.org/downloads/) on your development machine. Accept the default VM configuration during the install process.
+Before trying to use this project, install [Qubes 4.0.1](https://www.qubes-os.org/downloads/) on your development machine. Accept the default VM configuration during the install process.
 
 After installing Qubes, you must update both dom0 and debian-9 template VM to include the latest version of the `qubes-kernel-vm-support` package.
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -36,9 +36,9 @@ While we are strongly committed to piloting the use of Qubes OS for SecureDrop,
 
 Installing this project is involved. It requires an up-to-date Qubes 4.0 installation running on a machine with at least 12GB of RAM. You'll need access to a SecureDrop staging server as well.
 
-#### Qubes 4.0
+#### Qubes 4.0.1
 
-Before trying to use this project, install [Qubes 4.0](https://www.qubes-os.org/downloads/) on your development machine. Accept the default VM configuration during the install process.
+Before trying to use this project, install [Qubes 4.0.1](https://www.qubes-os.org/downloads/) on your development machine. Accept the default VM configuration during the install process.
```
emkll: In order to mitigate against CVE-2019-3462, Qubes recommends installing all debian-based templates from scratch. Shipping out own Debian-based template should help/mitigate for all non-whonix based templates

https://www.qubes-os.org/news/2019/01/23/qsb-46/

conorsch: My understanding is that thanks to [this patch](https://github.com/QubesOS/qubes-manager/pull/149/files), upgrading the debian-9 template in-place on first boot of the workstation will handle the case appropriately. You're right that you should add logic to verify it's done, to guard against mistakes. Hopefully Qubes 4.0.2 will back in the latest apt version in the Debian templates, but no reason to expect that'll be in place by the beta.

creviera:
Just making a note that @emkll and I are able to see the same RPC policy error for `GetDate` at this point. Will review and step through the doc as is in this PR and followup with comments.

conorsch:
Pushed up a small tweak to the GetDate tests, after conferring with @emkll. @creviera can you try again and see if you can reproduce the failing test? All tests are now passing for me locally.

creviera:
All tests pass :tada:

README.md
conorsch: We opened issue #249 to track.

conorsch:
Thank you for the careful review, @creviera !

PR #255, opened 2019-05-09
Autostart sd-proxy and sd-gpg at login

### LINKED ISSUE #200: Set required AppVMs to autostart
During testing of the client integration branch (#194), there's a bit of a workflow snag in that a user can fill out the securedrop-client login form, and only after clicking on **Sign in** are the corresponding helper VMs `sd-journalist` and `sd-whonix` started. Those should be marked autostart, along with `sd-gpg`.

emkll:
Fixes #200 

Test plan:

- `make all`  on this branch
- reboot
- [x] observe sd-gpg and sd-proxy have started without user intervention at boot

### DIFF
--- a/dom0/sd-gpg.sls
+++ b/dom0/sd-gpg.sls
@@ -20,6 +20,7 @@ sd-gpg:
       - label: purple
     - prefs:
       - netvm: ""
+      - autostart: true
     - tags:
       - add:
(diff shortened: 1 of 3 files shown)


redshiftzero:
Also closes https://github.com/freedomofpress/securedrop-client/issues/181

PR #251, opened 2019-04-15
Uploads dom0 RPM package for securedrop-workstation template

conorsch:
Provides scripts for uploading the RPM generated by `make template` to an S3-backed RPM repo. Uses a local container to generate the repo metadata locally, then uploads to S3 (assumes valid AWS credentials). 

Note that this implementation clobbers: whatever local repo is generated, that'll be pushed to the remote, with no regard for state maintenance, meaning prior versions of packages will no longer be available. That's fine for our near-term needs with testing, just be aware that changes you push to the remote are not currently version controlled, and are destructive. Stateful handling can be added as part of #157, once we resolve #250.

### Does the RPM repo work?
1. Clone this branch to dom0, then run `make all`. 
2. Confirm that you can see the package: `sudo qubes-dom0-update --action=search qubes-template-securedrop-workstation`
3. Install the package: `sudo qubes-dom0-update qubes-template-securedrop-workstation`
4. Create a new AppVM based on the `securedrop-workstation` [sic; the Salt-provisioned template is still called `sd-workstation`] template. Confirm you can log into the VM and update it.
5. Clean up the AppVM: `qvm-remove <vm_name>`
6. Uninstall the package: `sudo dnf remove qubes-template-securedrop-workstation`

### Can you upload packages to the repo? 
0. Confirm you have valid AWS credentials. (@msheiny is a great resource for this.)
1. Follow the README docs to create an `sd-template-builder` AppVM, based on fedora-29, and run `make template` inside that VM to generate an RPM.
2. Follow the README docs to sign that RPM with the test key.
3. Copy that signed RPM into `rpm-repo/` on in your dev machine (e.g. `sd-dev`). 
4. Run `make publish-rpm`. 

The script will builder a local container, verify the signatures on the RPM, generate a RPM repo metadata, then upload both the signed RPM and the repo metadata to a remote S3 bucket.

PR #260, opened 2019-05-29
Installs SD workstation TemplateVM from RPM

### LINKED ISSUE #157: Host securedrop-workstation RPM for installation in dom0

conorsch:
Installs the newly hosted (#251) RPM package `qubes-template-securedrop-workstation` in dom0. The TemplateVM is used for all the SDW components. Here we update the clone logic from `debian-9` to `securedrop-workstation`, which is the VM as named by the RPM package.

The changes are straightforward in terms of the diff:

  1. Install the RPM in dom0 (was previously commented out in #251).
  2. Update the qvm.clone calls to omit debian-9.
  3. Updates all apt packages inside the TemplateVM.
  4. Adds a few config tests (that should have been included in #251).

Step 3 above is important because as soon as we start preferring the RPM to the debian-9 for clone operations, we'll see apt package drift that would be replicated across all cloned VMs. Therefore we run apt upgrade against _just_ the `securedrop-workstation` TemplateVM, prior to cloning, to ensure the packages are only fetched once.

Closes #157.

### Testing
Before starting, make sure updates are applied to your current Qubes TemplateVMs. Otherwise, the apt upgrade checks may show errors that are not related to the changes presented in this PR.

1. `make all` in dom0
2. `make test` in dom0 shows no errors
3. Run `securedrop-client` in `sd-svs` to confirm packages are installed

### DIFF
--- a/dom0/sd-gpg.sls
+++ b/dom0/sd-gpg.sls
@@ -16,7 +16,7 @@ sd-gpg:
   qvm.vm:
     - name: sd-gpg
     - present:
-      - template: sd-workstation-template
+      - template: securedrop-workstation
       - label: purple
     - prefs:
       - netvm: ""
--- a/README.md
+++ b/README.md
@@ -163,11 +163,11 @@ Replies and Source Deletion will be added in the next major release of the *Secu
 
 Exporting documents directly from within the *SecureDrop Client* is not currently supported, but you can export documents manually via USB by following these steps:
 
-1. Create an export VM based on `sd-workstation-template`.
+1. Create an export VM based on the `securedrop-workstation` template.
    1. Click the Qubes menu in the upper left of the screen.
    2. Click **Create Qubes VM**
    3. Name the VM `sd-export`
-   4. Set the template as `sd-workstation-template`
+   4. Set the template as `securedrop-workstation`
    5. Set networking to (none).
    6. Click **OK** to create the VM.
 2. Start the VM. Again from the Qubes menu:
--- a/dom0/fpf-apt-test-repo.sls
+++ b/dom0/fpf-apt-test-repo.sls
@@ -31,3 +31,11 @@ configure apt-test apt repo:
     - key_url: "salt://sd/sd-workstation/apt-test-pubkey.asc"
     - require:
       - pkg: install-python-apt-for-repo-config
+
+# Ensure all apt updates are applied, since the VMs
+# will be cloned, duplicating package version drift.
+update-all-apt-packages:
+  pkg.uptodate:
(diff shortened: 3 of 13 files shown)


dom0/sd-gpg.sls
```diff
@@ -16,7 +16,7 @@ sd-gpg:
   qvm.vm:
     - name: sd-gpg
     - present:
-      - template: sd-workstation-template
```
emkll: do you think it's worth circling back in https://github.com/freedomofpress/qubes-template-securedrop-workstation/ to update the name of the template? I  _think_ we would only need to change this line: https://github.com/freedomofpress/qubes-template-securedrop-workstation/blob/master/securedrop-workstation.conf#L26 (and rebuild, of course)

conorsch: No, I don't think it's worthwhile to alter the template name. You're right that it's inconsistent: `securedrop-workstation`, rather than `sd-workstation-template`. On the other hand, however, I see a useful distinction in that the naming convention here matches what's standard in the Qubes repos. To that end, we have a reason to leverage the slight inconsistency:

* `sd-<x>-template` for all cloned, Salt-managed VMs
* `securedrop-workstation` for the RPM-based image (currently only one)

Open to revising if folks find the naming consolidation useful, but would prefer to avoid the lengthy rebuilds at present, especially since we'd have to update the various config files in the builder repo, as well—having `securedrop-workstation.conf` emit `sd-workstation-template`, rather than `securedrop-workstation`, also has the potential to confuse.

conorsch:
> make clean does not remove securedrop-workstation template, dnf remove must be called to remove it. Would it be useful to include it there, or in another target?

The only reason I didn't purge the RPM in dom0 is the assumption that doing so would trigger another ~700MB download on the subsequent `make all`, which would drastically slow down rebuilds in the dev env. Let's make sure to flag in test plans when the underlying template should be removed.

PR #291, opened 2019-07-17
Provision submission key fingerprint in config.json to sd-svs

emkll:
Towards https://github.com/freedomofpress/securedrop-client/issues/475

The SecureDrop client expects a file named config.json in /home/user/.securedrop_client/ containing the journalist key fingerprint. This key is used to encrypt replies to the submission (journalist key), so that replies can be decrypted by journalists in the client. This is because replies are encrypted in the client.

/home/user/.securedrop_client folder is populated on first run, and since it's the home directory, must be applied to `sd-svs` and not `sd-svs-template`

dom0 config.json file should now contain the GPG fingerprint of the submission key, which will populate config.json file in sd-svs.

### Test plan
- `make clean`
- checkout this branch and `make clone`
- add a `submission_key_fpr` containing the GPG fingerprint of the SecureDrop instance used 
- `make all`
- [ ] all tests pass with `make test`
- [ ] I can send a reply to a source from the client

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,7 @@ sd-gpg: prep-salt ## Provisions SD GPG keystore VM
 sd-svs: prep-salt ## Provisions SD SVS VM
 	sudo qubesctl top.enable sd-svs
 	sudo qubesctl top.enable sd-svs-files
+	sudo qubesctl top.enable sd-svs-config
 	sudo qubesctl --show-output --targets sd-svs-template state.highstate
 	sudo qubesctl --show-output --targets sd-svs state.highstate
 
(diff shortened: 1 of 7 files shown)


creviera:
i ran `make test` and see a couple issues (the first one ive seen before):
1. policy issue
```
Policy for VMShell should be:
### BEGIN securedrop-workstation ###
$anyvm $tag:sd-workstation deny
### END securedrop-workstation ###


FAIL
```
2. submission_key_fpr error
```
ERROR: test_sd_client_config (test_svs.SD_SVS_Tests)
...
KeyError: 'submssion_key_fpr'
...
FAIL: test_Policies (test_qubes_rpc.SD_Qubes_Rpc_Tests)
```

creviera:
oh i see the new `submission_key_fpr` was added to the config.json.example. will copy that over and retest!

PR #298, opened 2019-07-24
Remove salt provisioning for sd-export files

### LINKED ISSUE #271: Packaging securedrop export code
We currently have logic in the `send-to-usb` script for exporting files to USB drives. We'll be adding more export logic as time goes on e.g. as we add support for printing and (eventually) other export options. We should consider how we'll be updating this logic: we can put files in place via salt in dom0, but a more robust approach would be taking the approach we use in the sd-proxy and sd-svs VMs, which is packaging up the code (in the `securedrop-proxy` and `securedrop-client` debian packages) and then just using salt to ensure that package is installed.

emkll:
Closes #271 
These tasks should now be done via the securedrop-export package: https://github.com/freedomofpress/securedrop-export

- [x] securedrop-export is uploaded to apt-test-qubes.freedom.press: https://apt-test-qubes.freedom.press/pool/main/s/

Once the above package is uploaded test plan should be as follows:
- Run `make clean`
- [ ] `make all`completes without error
- [ ] all tests pass with `make test`
- [ ] USB drive export works
- [ ] USB printing works


zenmonkeykstop:
Built and all tests pass with the exception of the policy one, due cruft in `/etc/qubes-rpc/policy/qubes.VMShell`.

File and printer work without error.

Branch needs rebasing against master, however, and should probably be retested. Also the logo PNG is now part of the securedrop-export package and should not need to be added via Salt.

Will rebase and retest.

conorsch:
>  due to cruft in /etc/qubes-rpc/policy/qubes.VMShell.

In my experience, such cruft is usually the result of interrupted upgrades, _not_ caused by drift in the branches for this repo. Clean up manually for now—we should aim to provide a more complete solution to the automated logic, however.

> File and printer work without error.

:tada:! Great news.

> Will rebase and retest.

Thanks, agree that's warranted now that we have conflicts, likely due to merge of https://github.com/freedomofpress/securedrop-workstation/pull/305

PR #304, opened 2019-08-01
Bump whonix-gw and whonix-ws to 15

### LINKED ISSUE #300: Whonix VMs approaching EOL
The Whonix 15 VMs [have been released](https://forums.whonix.org/t/whonix-15-has-been-released/7616). According to the [support schedule](https://www.whonix.org/wiki/About#Support_Schedule), Whonix 14 will no longer be supported after 2019-08-01. 

We should update the salt configs to use Whonix 15 in appropriate locations. Note that Qubes 4.0.2 will ship Whonix 15 by default (#279). Still, we must provide an upgrade path. See #262 for challenges in handling upgrades of the `sys-*` VMs in an unattended fashion.

emkll:
Fixes #300

### DIFF
--- a/dom0/sd-dom0-files.sls
+++ b/dom0/sd-dom0-files.sls
@@ -111,15 +111,15 @@ dom0-install-fedora-29-template:
     - pkgs:
       - qubes-template-fedora-29
 
-dom0-enabled-apparmor-on-whonix-gw-14-template:
+dom0-enabled-apparmor-on-whonix-gw-15-template:
   qvm.vm:
-    - name: whonix-gw-14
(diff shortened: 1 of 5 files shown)


emkll:
Debian packages currently do not support Debian Buster, the distribution on which Whonix-15 is based. As a result, the proxy package does not work in this branch.

This PR is therefore blocked on the migration to Buster which is described in https://github.com/freedomofpress/securedrop-workstation/issues/306

emkll:
closing in favor of https://github.com/freedomofpress/securedrop-workstation/pull/358

PR #310, opened 2019-08-19
Update qrexec policy keywords

### LINKED ISSUE #309: Update qrexec policy keywords
Qubes RPC policy definition keywords should use the `@` character instead of `$`.

For context see this security bulletin: https://github.com/QubesOS/qubes-secpack/blob/master/QSBs/qsb-038-2018.txt

pierwill:
Qubes advises that all qrexec policy keywords should now begin with `@` (instead of `$`).

For context, see the security bulletin here: https://github.com/QubesOS/qubes-secpack/blob/master/QSBs/qsb-038-2018.txt

Fixes #309

### DIFF
--- a/dom0/sd-dom0-qvm-rpc.sls
+++ b/dom0/sd-dom0-qvm-rpc.sls
@@ -17,7 +17,7 @@ dom0-rpc-qubes.ClipboardPaste:
     - marker_start: "### BEGIN securedrop-workstation ###"
     - marker_end: "### END securedrop-workstation ###"
     - content: |
-        $anyvm $tag:sd-workstation deny
+        @anyvm @tag:sd-workstation deny
 dom0-rpc-qubes.FeaturesRequest:
   file.blockreplace:
     - name: /etc/qubes-rpc/policy/qubes.FeaturesRequest
@@ -25,7 +25,7 @@ dom0-rpc-qubes.FeaturesRequest:
     - marker_start: "### BEGIN securedrop-workstation ###"
     - marker_end: "### END securedrop-workstation ###"
     - content: |
-        $anyvm $tag:sd-workstation deny
+        @anyvm @tag:sd-workstation deny
 dom0-rpc-qubes.Filecopy:
   file.blockreplace:
     - name: /etc/qubes-rpc/policy/qubes.Filecopy
@@ -33,8 +33,8 @@ dom0-rpc-qubes.Filecopy:
     - marker_start: "### BEGIN securedrop-workstation ###"
     - marker_end: "### END securedrop-workstation ###"
     - content: |
-        sd-proxy $tag:sd-client allow
-        $anyvm $tag:sd-workstation deny
+        sd-proxy @tag:sd-client allow
+        @anyvm @tag:sd-workstation deny
 dom0-rpc-qubes.OpenInVM:
   file.blockreplace:
     - name: /etc/qubes-rpc/policy/qubes.OpenInVM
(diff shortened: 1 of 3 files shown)


creviera:
To test this and https://github.com/freedomofpress/securedrop-client/pull/537:

1. checkout pr branch on sd-dev vm
2. make clone on dom0 vm
3. set env vars needed for make scripts
4. make clean && make all
- [ ] builds without error
5. make test to ensure the config tests for rpc policies are passing
- [ ] tests pass 
- [ ] sanity-check: visually inspect the rpc policies in dom0
6. give sd-svs network access and clone client repo
7. apply changes from the securedrop-client PR (https://github.com/freedomofpress/securedrop-client/pull/537/files) to sd-svs: copy .desktop file to `/usr/share/applications`, and `logic.py` file alongside the others configured by the venv in securedrop-client
8. run the client via `python -m securedrop_client` and trigger inter-vm communication that uses vm policies
- [ ] encryption and decryption of messages is successful
- [ ] able to open up files

creviera:
> 5. make test to ensure the config tests for rpc policies are passing 
> - [ ] tests pass
> - [ ] sanity-check: visually inspect the rpc policies in dom0

`test_Policies` fails in `test_qubes_rpc.py` for the following policies:
* GetDate
* GetRandomizedTime
* GetImageRGBA
* InputKeyboard
* InputMouse
* NotifyTools
* NotifyUpdates
* ReceiveUpdates
* SyncAppMenus
* UpdatesProxy

It looks like some of the rpc policies in `dom0` are showing up with `$` instead of `@`, e.g. qubes.GetDate:
```
$tag:anan-vm          $anyvm    deny
$anyvm   $anyvm    allow,target-dom0
```
And the test expects this to look like the policy specified in the file that you updated (tests/var/qubes-rpc.yml):
```
@tag:anan-vm          @anyvm    deny
@anyvm   @anyvm    allow,target-dom0
```

There doesn't seem to be sls files in this repo for the policies that are failing for me, so first step is to figure out where those policies are being defined.

pierwill:
:wave:

Those policies are created by Qubes in dom0 at /etc/qubes-rpc/policy. The $ keywords are still compatible and have not all been changed yet.

pierwill:
How hard would it be to overwrite these default policies using the @ character as part of the sls provisioning?

pierwill:
For instance, could there be a step in /dom0/sd-dom0-qvm-rpc.sls to replace all $ characters with @ in /etc/qubes-rpc/policy/* ?

I'm not familiar with salt, so I might be off base here.

conorsch:
Thanks for the thorough review, @creviera.

@pierwill You raise good points here. We initial designed the RPC policy update logic (see #187) to respect non-SDW-specific lines, by using a blockreplace strategy.

Upstream updates to dom0 don't clobber modified files within `/etc/`, though, instead placing the newer upstream files adjacent to the existing files, with a `.rpmnew` suffix. That means we're not actually getting the upstream updates, so perhaps we should indeed update the Salt logic to clobber files by templating the entirety of the file as we expect it to be structured. 

The major downside to the full-template strategy is that _any_ customizations, even on developer laptops, will be destroyed. That's going to be difficult to support. To your question about substituting `@` for `$`, yes, certainly that's something we can do—it may not resolve all of our problems, though. We'll discuss in an upcoming SD Engineering meeting (see [schedule](https://github.com/freedomofpress/securedrop/wiki/Standup-Notes)—you're welcome to join!) and report back here with next steps.

Many thanks for your continued contributions, @pierwill! Great to have you involved!

eloquence:
Hi @pierwill, since this isn't impacting us during development, we've decided to defer further review on this and the other qrexec changes for now, but we'll definitely aim to take another look at this before the first beta release. Thanks again for your work on this.

emkll:
Thanks @pierwill for your contribution, and @creviera for careful testing. I have rebased on latest `master` and appended a commit to your branch, to ensure test pass on a clean Qubes 4.0.1 install and provisioning run of the workstation. The current provisioning logic scopes qrexec policy changes to specific blocks. 

The rationale here is that we want to make sure that the policies are consistent across installs, and without careful logic, it might pose availability issues or hard to maintain consistent across several installs. I think we should revisit this later when Qubes 4.0.2 is released, and decide to replace `$` by `@` accross the board, if upsteam does not replace the `$` by `@` by default by then.

Since I've appended a commit (but tested these changes in a qubes install), could someone else take a look at this?

### Proposed test plan (based on @creviera 's above)
- [ ] make test (baseline to ensure your qrexec policies have not been modified / the tests are passing
- check out this branch and make clone
- [ ] make all / make test should pass
- [ ] make clean / make all / make test should pass

PR #326, opened 2019-11-09
doc: add note about provisioning logs

redshiftzero:
adds a note for people running `make all` re: a place they can inspect for logs on issues during provisioning


creviera:
oh question before merging, do you think we should also add that if you want to monitor those logs as they're being written you can run `journalctl -ef`?

redshiftzero:
we can add that if we note that there is a _lot_ of verbose output - want to add a commit saying that?

PR #327, opened 2019-11-09
Ensure whonix repos are removed before any other apt command on sd-proxy-template

### LINKED ISSUE #325: sd-proxy not getting installed in sd-proxy AppVM
Several people on recent `make all` runs have reported not having `sd-proxy` installed in the `sd-proxy` AppVM. 

This ticket is both for awareness for people that are trying to provision for the first time but can't log in - ensure that you are installing the `securedrop-proxy` package in the `sd-proxy` AppVM (package name is `securedrop-proxy`, the command it exposes is `sd-proxy`) and to remind us that we should investigate what's going on here.

emkll:
Fixes #325, related to #313: Ensures the securedrop-proxy package is correctly installed on sd-proxy-template.

### Test plan
- `make clean`
- `make all`
- [x] all tests pass when running `make test`
- [x] the client can successfully log into a server


emkll:
Thanks for the review @redshiftzero. Unfortunately, the current approach for the fix was less reliable as previously thought. I have made changes to the approach here, to ensure the whonix repos are more reliably disabled (and ensuring upstream Debian changes still make it into the templates).

Some comments related to this changes:
1. `sd-whonix` is based on `whonix-gw-14`, so we need to apply changes to that template.
2. Per whonix's instructions (both in `/etc/apt/sources.list.d/whonix.list` and [1]), Whonix recommends `apt_repository --disable` be used, as `/etc/apt/sources.list.d/whonix.list` may be overwritten.
4. In first-time-installs, python-futures installation may fail, which means we must disable the whonix repos reliably before that step.

This should now be ready for re-review.

[1] https://www.whonix.org/wiki/Project-APT-Repository

PR #328, opened 2019-11-09
Update screenshots per latest nightly builds

eloquence:
Archived old ones in alpha-workflow directory for historical
reference.

## Status

Ready for review. (Tip: Use "rich diff" icon next to `<>` icon for individual files, for image reviews.)


conorsch:
> As we approach The Great Docs Push, let's have a team discussion about the "right" way to store these large screenshots.

Mentioned during standup today. Agreed that we'll discuss again closer to the docs push. For now, fine with the increase in filesize for the repo.

PR #331, opened 2019-11-13
Upgrades all sys-* VMs to Fedora 30

### LINKED ISSUE #262: Update sys-* VM templates in unattended upgrades
Fedora 28 is now end-of-life [0], AppVMs that use Fedora 26 template should be updated to a more recent release. We should use Fedora 29 now or wait for a stable Fedora 30 template.

[0] : https://fedoraproject.org/wiki/End_of_life

@emkll @conorsch

### LINKED ISSUE #312: Set sd-svs-disp as default DispVM
Encountered this failure on my workstation:

```
FAIL: test_dispvm_default_platform (test_vms_platform.SD_VM_Platform_Tests)
  File "/home/user/securedrop-workstation/tests/test_vms_platform.py", line 143, in test_dispvm_default_platform
    self.assertEqual(result, "fedora-29-dvm")
AssertionError: 'sd-svs-disp' != 'fedora-29-dvm'
- sd-svs-disp
+ fedora-29-dvm
```

Indeed, I had set my workstation to use the `sd-svs-disp` template, partially to dogfood it more, but also because it's quite a good DispVM: handles common formats and doesn't have networking enabled. The SDW workstation config still expects the Qubes default of `fedora-29-dvm`, but `sd-svs-disp` seems like a saner default for the Workstation.

### LINKED ISSUE #329: Use fedora-30 for sys-* VMs
It's that time again! Fedora-29, currently, used for the `sys-*` VMs (viz. `sys-firewall`, `sys-net`, `sys-usb`) is approaching EOL. As of this writing, Fedora 29 is listed on neither the EOL Fedora info page, nor the supported versions page:

* https://fedoraproject.org/wiki/End_of_life
* https://fedoraproject.org/wiki/Releases

Still, F29's days are numbered, so let's make the jump. In the past, we've recommended that developers make the change manually. We still have #262 looming, however, so I'll take a closer look at what are options are to make these tricky changes in the configuration logic. 

Related issues:

* #114
* #189
* #245

### LINKED ISSUE #330: Make SDW dom0 dnf repo config persistent
The initial support for dom0 RPMs such as the `qubes-template-securedrop-workstation` config introduced in #260 has a major deficiency: the configuration does not persist across workstation reboots. That's due to the fact that the pubkey for the RPM repo is placed directly inside the `sys-firewall` AppVM. In order for the config to persist, we have two options:

1. Add the configuration to the TemplateVM for `sys-firewall` (e.g. `fedora-30`)
2. Add the configuration to `/rw/config/` in `sys-firewall`, so the config is restored on each boot of the AppVM.

Between the two, 2) is preferable: we don't want _all_ Fedora-based machines to include the dom0 pubkey repo info, and we don't want to have to update the Template settings every time a Fedora version goes EOL (see #329).

conorsch:
Implements unattended updates to the built-in Qubes `sys-*` VMs, specifically:

* `sys-net`
* `sys-firewall`
* `sys-usb`

If the VMs are _not_ running the latest version, they will briefly be halted to change the template setting, then restarted on the new template.

Updates the default DispVM template to be `sd-svs-disp`. Enabling this as the default means that `qvm-open-in-dvm <file>` from _any_ VM will result in `sd-svs-disp` being used. 

Improves the dom0 RPM repo logic to be persistent across reboots. Doing so was required in order to get the tests passing after the template update specifically for `sys-firewall`.

Addresses changes specified in a few different issues; thus:

* Closes #262
* Closes #312
* Closes #329
* Closes #330

### Testing
In order to test, first confirm that your environment is *not* what's enforced by this PR. Specifically, your system should be configured as follows:

```
[user@dom0 ~]$ qvm-prefs sys-net template
fedora-29
[user@dom0 ~]$ qvm-prefs sys-firewall template
fedora-29
[user@dom0 ~]$ qvm-prefs sys-usb template
fedora-29
[user@dom0 ~]$ qubes-prefs default_template
fedora-29
[user@dom0 ~]$ qubes-prefs default_dispvm
fedora-29-dvm
```

The above shows the pre-merge "correct" settings, namely F29 rather than F30 throughout.
If you already have a prior version of the SDW config provisioned, do **not** run `make clean`.
Instead, simply run `make all`, which will confirm that the VMs are updated in-place despite
a pre-existing config.

Afterwards, the output of the above commands should be:

```
[user@dom0 ~]$ qvm-prefs sys-net template
fedora-30
[user@dom0 ~]$ qvm-prefs sys-firewall template
fedora-30
[user@dom0 ~]$ qvm-prefs sys-usb template
fedora-30
[user@dom0 ~]$ qubes-prefs default_template
fedora-30
[user@dom0 ~]$ qubes-prefs default_dispvm
sd-svs-disp
```

As always, ensure that `make test` passes without error. New tests have been added to verify
the above settings; the example commands are provided for clarity.

Succinct checklist for testing:

- [ ] `make all` shows no errors
- [ ] `make test` show no errors
- [ ] all `sys-*` VMs running `fedora-30`
- [ ] start/restart an AppVM, confirm `qvm-open-in-dvm <file>` opens in an SDW netless VM

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,10 +218,10 @@ In the future, we plan on shipping a *SecureDrop Workstation* installer package 
 
 ### Building the Templates
 
-1. Create a `fedora-29` AppVM for building the templates. It's going
+1. Create a `fedora-30` AppVM for building the templates. It's going
    to need Docker and several other packages every time you use it, so
    it might be worth creating another template derived from
-   `fedora-29`, into which you can install those extras, and basing
+   `fedora-30`, into which you can install those extras, and basing
    the builder VM on that, or just using a StandaloneVM to save time
    and repetition.
 2. Increase the disk size to at least 15GB (as the build uses over
@@ -454,8 +454,8 @@ make sd-export
 
 ### Transferring files via OnionShare
 
-1. Create an `sd-onionshare-template` VM based on `fedora-29`:
-   1. Click on the Qubes menu in the upper left, select "Template: Fedora 29", click on "fedora-29: Qube Settings", and click on **Clone Qube**
+1. Create an `sd-onionshare-template` VM based on `fedora-30`:
+   1. Click on the Qubes menu in the upper left, select "Template: Fedora 30", click on "fedora-30: Qube Settings", and click on **Clone Qube**
    2. Name the cloned qube `sd-onionshare-template`
    3. In the Qubes menu on the top-left, select "Template: sd-onionshare-template" and click on "sd-onionshare-template: Terminal"
    4. Install OnionShare: `sudo dnf install onionshare`
--- a/dom0/sd-dom0-files.sls
+++ b/dom0/sd-dom0-files.sls
@@ -24,23 +24,6 @@ dom0-rpm-test-key-import:
     - require:
       - file: dom0-rpm-test-key
 
-dom0-rpm-test-key-sys-firewall:
-  cmd.run:
-    # Pass in the pubkey directly to sys-firewall, so it's available on the
-    # UpdateVM for dom0 while we're configuring dom0 repos.
-    - name: >
-        qvm-run -p sys-firewall '
-        sudo tee /etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation-test'
-        < /srv/salt/sd/sd-workstation/apt-test-pubkey.asc
-
-dom0-rpm-test-key-sys-firewall-import:
-  cmd.run:
-    - name: >
-        qvm-run --no-gui sys-firewall '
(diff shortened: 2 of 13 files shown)


conorsch:
> One improvement we could make here is to also uninstall the template (sudo dnf remove qubes-template-fedora-29), as it will no longer be required, and will only increase upgrade time and likelihood of misconfigurations.

Good point. During standup today, @kushaldas mentioned he'd _like_ to keep the older F templates around. We may be forced to do so for a while, or at least update the logic here to vigorously replace _all_ templates. For instance, I had `sd-template-builder` and `sd-onionshare` that were still on F29 (and manually updated them to F30).

I'll open a follow up ticket to discuss cleanup, which will also be relevant for the W14 -> W15 upgrades (#300), deferring changes for now.

Issue #345, opened 2019-11-20
Specify Buster-based securedrop-workstation template as new default


eloquence:
Part of #306, dependent on #344 and the final task of the Buster migration: Once all dependencies are met, we can declare the Buster-based template to be the new default, after which a `make all` run should provision a SecureDrop Workstation in which all VMs are fully based on Debian Buster.

emkll:
For in-place updates, there are two approaches we can consider. There's added complexity compared to #329 , because the templates are cloned from a base template. We can either

1. Create a whole new set of templates when changing distributions (e.g. sd-svs-buster-template), and then perform the swap and remove the old templates at the end of the run as part of the cleanup task. In this case, it appears that we *cannot* rename the templates, template names are read-only when using qubes-prefs (It can only be done by the UI)
2. At the start of the run, if the template is debian-9 based, swap it with a known exiting template (e.g. securedrop-workstation-buster), delete the existing (debian-9) template, and then continue with the install, creating all buster-based templates with a naming scheme consistent with the current approach (`sd-svs-template`).

Option 1: 
(+)Name for templates changes
(-) Name for templates is distibution specific
(-) Larger diff overall, requires replacing names for all the templates

Option 2:
(+) smaller diff
(-) requires more careful logic for conditional execution of old template cleanup (perhaps using tags)

I have tested both options and believe they are both feasible. I would be inclined to go with option 1 since option 2 requires a fair amount of conditional logic, thus more error-prone and harder to maintain.

redshiftzero:
agreed on 1, a more straightforward implementation even if the diff is larger is wise. I don't see any issue with having the template names distribution specific (indeed, it might even be advantageous so admins/devs can easily see at a glance in the name which distro they're on)

emkll:
An additional challenge with the in-place upgrade, that was also present in #329: When changing a template for a given VM, this VM cannot be powered on. This means that we would need to either:
a) automatically shut down VMs which templates need updating (and interrupt a user's 
b) have the update user-intiated

Issue #342, opened 2019-11-20
Consider removing build/infra tooling and requirements to another repository


emkll:
In the current distribution method, we use git to deliver code to workstations. Pipfile(.lock) or requirements.txt (introduced in https://github.com/freedomofpress/securedrop-workstation/pull/318) are not needed to run the SecureDrop Workstation in a production context. 

We should consider removing the logic and requirements to another repo, to avoid accidental install of dependencies in a production environment, as these are intended to be used in a dev-only context, and not reviewed as thoroughly as other dependencies we include.

Once we address #174 , this will no longer be an issue.

conorsch:
Since we're moving away from S3 buckets for RPMs, switching to git-lfs, we'll shortly be able to drop the `requirements*` files altogether, given that they were used only for pulling in the AWS/S3 tooling. 

Still, your point stands: we have more thinking to do about the "install" workflow for a fresh instance. I don't see a better method than git -> tar -> dom0 for the initial configuration of the dom0 yum/dnf repos. After that point, everything will be package-based, but we want to reduce the chances of misconfiguration even before that step. 

We currently have two RPMs that this repo can create:

* `securedrop-workstation-dom0-config` 
* `qubes-template-securedrop-workstation{,-buster}`

The latter is a special case, but the small amount of build logic we currently have (https://github.com/freedomofpress/securedrop-workstation/blob/e63773d059c97f9d9d712ac38593591c31cc4bc0/builder/build-workstation-template) could be moved to https://github.com/freedomofpress/qubes-template-securedrop-workstation. 

For the salt package, I don't mind the logic living here, but perhaps we should create an RPM-equivalent of https://github.com/freedomofpress/securedrop-debian-packaging. That, or we could rename that repo to "securedrop-packaging" and handle all of it there.

redshiftzero:
> I don't see a better method than git -> tar -> dom0 for the initial configuration of the dom0 yum/dnf repos

i wonder if we could eventually get our packages included in qubes rpm repository (not sure what the process is for this)... otherwise until then yeah I agree

> For the salt package, I don't mind the logic living here, but perhaps we should create an RPM-equivalent of https://github.com/freedomofpress/securedrop-debian-packaging. That, or we could rename that repo to "securedrop-packaging" and handle all of it there.

[...]

PR #349, opened 2019-11-22
Updates dom0 repo URL

### LINKED ISSUE #334: `make clone` should default to `sd-dev` VM
`make clone` currently defaults to `work`; per the updated recommendation to create an `sd-dev` VM and removal of all references to the `work` VM, it should default to `sd-dev` if no VM is specified via `SECUREDROP_DEV_VM`.

### LINKED ISSUE #347: RPM key script appends same line to end of rc.local
I have observed by rc.local file in sys-firewall containing a large number of repeated entries: 
```
[...]
/rw/config/sd-copy-rpm-repo-pubkey.sh
/rw/config/sd-copy-rpm-repo-pubkey.sh
/rw/config/sd-copy-rpm-repo-pubkey.sh
/rw/config/sd-copy-rpm-repo-pubkey.sh
/rw/config/sd-copy-rpm-repo-pubkey.sh
/rw/config/sd-copy-rpm-repo-pubkey.sh
/rw/config/sd-copy-rpm-repo-pubkey.sh[user@sys-firewall ~]$ 
```
This is because the `file.line` mode is set to `insert` in [0] . Per the salt docs [1], `ensure` might be a better fit here to prevent multiple entries. 

[0] : https://github.com/freedomofpress/securedrop-workstation/blob/28e3f42cebb5853b2e845436129dab406b95f660/dom0/sd-sys-firewall-files.sls#L17:L23

[1] : https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html#salt.states.file.line

conorsch:
Moves the package backend for dom0-specific RPM hosting from S3 buckets to git-lfs-backed storage. Now, PRs into https://github.com/freedomofpress/securedrop-workstation-dev-rpm-packages-lfs will result in packages automatically being served up to dom0.

Since the S3 buckets (for the purposes of RPM hosting) are obsolete after merge of this PR, we also remove the scripting in this repo required to push packages manually from developer workstations (see #342).

Also includes some minor fixes, therefore:

* Closes #347 
* Closes #334
* Refs #345

### Review
Before performing functional review, consider the following:

* Is the URL appropriate? Right now it's `https://yum.securedrop.org/workstation/dom0/f25/`; perhaps we shold be using the "yum-test" subdomain, to distinguish between prod/test, same as we do with e.g. `apt{,-test}.freedom.press`. 

### Testing

In order to test that the repo is correctly configured, you must purge RPM packages from dom0. It'll take a while to pull in and install the new ones, so consider obtaining caffeine through your favorite method while the below tasks are running.

```
make clone
make clean
sudo dnf remove -y qubes-template-securedrop-workstation
qvm-check securedrop-workstation # should show an error, because it's now missing!
make all
qvm-check securedrop-workstation # should report success, because it's been reinstalled
```


dom0/sd-dom0-files.sls
```diff
@@ -38,7 +38,7 @@ dom0-workstation-rpm-repo:
         gpgcheck=1
         gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation-test
         enabled=1
-        baseurl=https://dev-bin.ops.securedrop.org/dom0-rpm-repo/
+        baseurl=https://yum.securedrop.org/workstation/dom0/f25
```
redshiftzero: ah yeah you're right we should make this `yum-test` so it's clear this is not for production use

conorsch: Roger, agreed: I'll update the URL here and re-ping for review.

conorsch: Updated, ready for full review.

kushaldas: I wish for it to be `rpm.securedrop.org` as yum is already obsolete.

conorsch: > I wish for it to be `rpm.securedrop.org` as yum is already obsolete.

The only reason I chose "yum{,-test}" is that's what the Qubes repo use: https://yum.qubes-os.org/ Since we're still in dev, we can change the URL again in the near future, but it'll take some coordination with the backend.

PR #352, opened 2019-11-27
Update base templates to buster

### LINKED ISSUE #345: Specify Buster-based securedrop-workstation template as new default

emkll:
Closes #345 

For new installs: Buster templates by default
For existing installs: In-place update to buster

Note that the upgrade logic will *shut down* sd-svs and sd-gpg (normally only a single time)

#### Clean scenario
* make all and all tests should pass
#### Upgrade scenario
* make all and all tests should pass


scripts/prep-upgrade
```diff
@@ -0,0 +1,73 @@
...
+# Finally for sd-gpg, we simply shutdown the machine
+qvm-check sd-gpg > /dev/null 2>&1
+if [[ $? -eq 0 ]]; then
+	BASE_TEMPLATE=$(qvm-prefs sd-gpg template)
+
+	if [[ ! $BASE_TEMPLATE =~ "buster" ]]; then
+		qvm-shutdown sd-gpg
+		sleep 5
+	fi
+fi
```
conorsch: At the end of this script seems to be the appropriate time to clean up the old VMs. That, or in a subsequent Salt task. Hardcoding the VMs to remove seems to be what we want. This Python one-liner describes the logic:

```
python3 -c 'from qubesadmin import Qubes; q = Qubes(); sdw_vms = [x.name for x in q.domains \
            if "sd-workstation" in x.tags and x.klass == "TemplateVM" \
            and "buster" not in x.name \
            and not x.installed_by_rpm ]; print("\n".join(sdw_vms))'
```

I _don't_ recommend we use that convoluted one-liner, although it does illustrate how useful the Admin API can be for use cases like this. For now, we can simply hardcode the list of "old" VMs in a Salt list and ensure they're gone.

scripts/provision-all
```diff
@@ -7,7 +7,7 @@ set -o pipefail
 
 # Format list of all VMs comma-separated, for use as qubesctl target
 all_sdw_vms_target="$(./scripts/list-vms | perl -npE 's/\n/,/g' | perl -npE 's/,$//' )"
-
+templates_to_remove="$(./scripts/list-vms old-templates | perl -npE 's/\n/ /g')"
```
conorsch: N.B. that #351 _removes_ the `list-vms` script, in favor of the new `qvm-ls --tags` support. As above, if we perform the removal in Salt, we can execute in parallel against all of them, as long as we wait until the bash script has finished.

scripts/provision-all
```diff
@@ -17,7 +17,12 @@ echo "Create base Template to be used by others"
 sudo qubesctl --show-output --targets dom0 state.sls sd-workstation-template
 
 echo "Configure packages inside base Template"
-sudo qubesctl --show-output --skip-dom0 --targets securedrop-workstation state.sls sd-workstation-template-files
+sudo qubesctl --show-output --skip-dom0 --targets securedrop-workstation-buster state.sls sd-workstation-template-files
+
+# This is required only when swapping base templates, see scripts/prep-upgrade
+# for more details.
+echo "Prep for upgrade, shut down AppVMs that require reboot for template swap"
+./scripts/prep-upgrade
```
conorsch: Recommend moving the script call _out_ of the `provision-all` script and into a Salt state, reason being that then we have the ability to trigger the change via unattended-upgrades, e.g. via dom0 state.highstate. Currently only developers run this bash script, so let's not depend on it overmuch to enforce updates we want to land on pilot/prod instances.

scripts/provision-all
```diff
@@ -27,3 +32,6 @@ echo "Provision all SecureDrop Workstation VMs with service-specific configs"
 # The max concurrency reduction (4->2) was required to avoid "did not return clean data"
 # errors from qubesctl. It may be possible to raise this again.
 sudo qubesctl --show-output --max-concurrency 2 --targets "$all_sdw_vms_target" state.highstate
+
+echo "Cleaning up old templates"
+./scripts/destroy-vm "$templates_to_remove"
```
conorsch: Same as above: let's port the removal to Salt, so it's enforced even during highstate runs when the bash script isn't called.

tests/test_dom0_rpm_repo.py
```diff
@@ -26,7 +26,7 @@ def test_rpm_repo_config(self):
             "gpgcheck=1",
             "gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation-test",  # noqa
             "enabled=1",
-            "baseurl=https://dev-bin.ops.securedrop.org/dom0-rpm-repo/",
+            "baseurl=https://yum-test.securedrop.org/workstation/dom0/f25",
```
conorsch: Thank you for fixing this! Heads up, fix also exists in #351.

scripts/prep-upgrade
emkll: Unfortunately, we cannot set them as part of this script, because the old (stretch) templates are still being used: we need to wait for the state of the templates to be applies, so that the stretch templates are no longer being used by an appvm

redshiftzero:
i'm seeing concurrent.futures as uninstalled for python 2 in the buster base template - my understanding is that salt is using this - can someone explain?

conorsch:
@redshiftzero Judging by https://github.com/QubesOS/qubes-issues/issues/4272#issuecomment-554623309, we should re-evaluate whether we must ensure this package is present.

conorsch:
After some code spelunking by @rmol, we're satisfied that the Qubes salt mgmt logic is using python3 on the target VMs, meaning that concurrent is no longer necessary.

emkll:
rebased on latest master, which now includes #351

scripts/handle-upgrade
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+set -uo pipefail
```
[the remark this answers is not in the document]
conorsch: As for the business logic of the script... nothing to change, it works remarkably well! Looking forward to flipping to a Python script, but as written, this is just what we need.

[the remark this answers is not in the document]
emkll: The reason `-e` is not set is because `qvm-check` returns non-zero if the machine does not exist. We can't assume that it's a run-once-and-forget script (if the first machine doesn't exist, it's possible the others do): If the script is interrupted, or if there's a situation like in #300 where we would consider appending other machines to the script, those commands would never be reached.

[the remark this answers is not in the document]
conorsch: > The reason -e is not set is because qvm-check returns non-zero if the machine does not exist.

Because all calls to `qvm-check` are in `if` conditionals:

```
$ grep -F qvm-check scripts/handle-upgrade 
	if qvm-check sd-svs --quiet; then
			if qvm-check --running sd-svs; then
	if qvm-check --quiet sd-svs-disp; then
	if qvm-check --quiet sd-export-usb; then
	if qvm-check --quiet sd-gpg; then
		if qvm-check "${template}" --quiet; then
			if qvm-check --running "${template}"; then
```

we needn't worry about the script exiting if one of the machines is missing: the `if` will gobble up any non-zero exit code, causing the conditional to evaluate as false, but `set -e` will _not_ trigger a full exit of the script. For example:

```
[user@dom0 ~]$ cat example.sh
#!/bin/bash
set -e
set -u
set -o pipefail


if qvm-check --quiet non-existent-vm; then
    echo "Yes, we found the VM"
else
    echo "No, that VM does not exist"
fi

echo "Finally, time to exit"
[user@dom0 ~]$ ./example.sh
usage: qvm-check [--verbose] [--quiet] [--help] [--all] [--exclude EXCLUDE]
                 [--running] [--paused] [--template] [--networked]
                 [VMNAME [VMNAME ...]]
qvm-check: error: no such domain: 'non-existent-vm'
No, that VM does not exist
Finally, time to exit
[user@dom0 ~]$ echo $?
0
```

The fact that "Finally, time to exit" is printed means that the script didn't exit early, even with `set -e` in place. Adding `set -e` to the script in this PR will ensure that problems running running e.g. `qvm-shutdown` cause the script to exit—which appears to me to be what we want. We don't want to blithely continue unless each and every one of the steps we list passes, otherwise the state is no longer predictable.

emkll:
Thanks @conorsch your changes look good to me, consider those approved. They have been tested in both install and upgrade scenarios: `make all` and `make test` pass in both cases (including additions described below)

I've resolved all but one review comment in b9f07e1, commented on that final comment (`set -e`).

I've also appended two further more commits, for your review, to ensure test coverage of the conditional apt-test logic, as well as improvements to the cleanup logic.

scripts/handle-upgrade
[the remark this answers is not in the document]
emkll: Thanks for the clarification @conorsch ! You are correct, just tested locally the script alone as well as the changes in a clean install (which would be the worst case for -e here), and everything works as expected

emkll:
Just pushed some final changes (see https://github.com/freedomofpress/securedrop-workstation/compare/469a3fe..7e56fca) addressing the comments in https://github.com/freedomofpress/securedrop-workstation/pull/352#discussion_r352920683

This is now ready for re-review

conorsch:
Roger that. Proceeding with upgrade test. In preparation, I provisioned from latest master (75de80263fe6e919fb0c8a1a70c7719be40f909e), and will apply these changes _without_ a `clean` to verify the upgrade-in-place logic.

conorsch:
All tests passing for the upgrade-in-place scenario. Overall, the process is remarkably smooth. Thanks for the hard work on this one, @emkll.

PR #354, opened 2019-12-02
Updates SecureDrop logo file (for v1.0 style)

conorsch:
Matches the changes shipped with SecureDrop v1.0.

To perform this update, I simply ran `cp ../securedrop/securedrop/static/i/logo.png ./sd-proxy/logo-small.png`, with the https://github.com/freedomofpress/securedrop repo cloned adjacent.

### Review
Make sure to view the rich diff to evaluate the difference here. The new file is not _just_ the logo, but also contains the "SecureDrop" text beneath it. It should be suitable for now, and likely better already than using the old style.


eloquence:
> Make sure to view the rich diff to evaluate the difference here. The new file is not just the logo, but also contains the "SecureDrop" text beneath it

Is this the logo that's shown in the pop-up notifications e.g. during automatic updates, or is it used anywhere else currently? If it's just the notifications, it might be preferable to use the version w/o text, given the small size (cc @ninavizz). If so, I can update accordingly.

emkll:
@eloquence For now, it is only used for notifications.
before: 
popup-icon-old
after:
popup icon

ninavizz:
OOF... @emkll @conorsch for both this icon and the desktop icon, do y'all need a PNG or SVG? I do need to draft brand guidelines with canonical artwork to reference that are is all correct, but for now I'd like to offer a Cube-only that is correct (with white lettering/keyhole). 

Also, is there a way to use a different version of the logo for the one that Qubes colorizes in the "spaces" menu-thingy at the top, and in the Q menu dropdowns? Sure, I'd rather they just not do that... but until that change potentially happens, I'd like to use monochromatic high-contrast art to improve icon recognition, there. If not, then I'd recommend using the 1-color icon, everywhere _except_ the desktop.

eloquence:
I pushed a commit to update it to a version without the wordmark, and with the white cube internals. Let me know what y'all think (thanks to @ninavizz for preparing an asset, which I cropped and resized for this purpose).

eloquence:
This is ready for re-review.

zenmonkeykstop:
`make all`ing now, but in case I forget, the logo will probably also need to be updated in the securedrop-export, as it's installed to sd-export repos via a Debian package, not via Salt.

PR #364, opened 2019-12-10
Add AppArmor test for securedrop-client

### LINKED ISSUE #234: Create AppArmor profile for Securedrop client
In the event of a client code vulnerability, we want to minimize the attack surface for a potential exploit. To do so, we want to configure an AppArmor profile that would restrict client access to the resources it needs to function.

See also #233, which would reduce the risk of an attacker achieving persistent changes to the VM state.

emkll:
Closes #234 

This PR provides tests to ensure the apparmor profile for the securedrop-client is properly running

### Merge order
The changes introduced here span several repositories, and must be merged in the order that follows: 

- [x] Kernel config:  https://github.com/freedomofpress/ansible-role-grsecurity-build/pull/54
- [x] Kernel packaging changes: https://github.com/freedomofpress/securedrop-debian-packaging/pull/116
- [x] Image (meta)packages: https://github.com/freedomofpress/securedrop-dev-packages-lfs/pull/30
- [x] securedrop-client changes (apparmor profile, version increment): https://github.com/freedomofpress/securedrop-client/pull/673
- [x] 0.0.11 tag created in securedrop-client
- [x] https://github.com/freedomofpress/securedrop-debian-packaging/pull/105 (ci should be re-run and pass after 0.0.11 client tag is pushed.
- [x] All the boxes above are checked, this PR (https://github.com/freedomofpress/securedrop-workstation/pull/364) can now be reviewed.
### Test plan
- run securedrop-update
- [ ] securedrop-client version running in sd-svs is 0.0.11
- make clone this branch (no need to make all, this branch only changes tests)
- make test
- [ ] all tests pass

### DIFF
--- a/tests/test_svs.py
+++ b/tests/test_svs.py
@@ -44,6 +44,11 @@ class SD_SVS_Tests(SD_VM_Local_Test):
         line = '{{"journalist_key_fingerprint": "{}"}}'.format(submission_fpr)
         self.assertFileHasLine("/home/user/.securedrop_client/config.json", line)
 
+    def test_sd_client_apparmor(self):
+        cmd = "sudo aa-status --json"
+        results = json.loads(self._run(cmd))
+        self.assertTrue(results['profiles']['/usr/bin/securedrop-client'] == "enforce")
+
 
 def load_tests(loader, tests, pattern):
     suite = unittest.TestLoader().loadTestsFromTestCase(SD_SVS_Tests)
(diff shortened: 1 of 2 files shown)


emkll:
Good news? If a profile is in /etc/apparmor.d/ and apparmor-utils is installed, based on my local testing, it will automatically be enforced. However, it would be prudent to block the start of the client if the profile cannot be loaded. I think we should address this in a follow up item, due to the complexities of handling this, as /usr/bin/securedrop-client is the process that is being confined. I am wondering if adding some additional scripting is required here (rc.local. Perhaps we can merge this as-is and observe, over time, if apparmor no longer enforced the profile for whatever reason (since we have the tests in place)

emkll:
rebased on latest master and updated the test plan. This is now ready for final review.

PR #367, opened 2019-12-11
Clean action: better handling of default dispvm

### LINKED ISSUE #360: Make clean will fail if fedora-30-dvm does not exist
There are a few ways we could address this issue: 
1. Setting the default DispVM to `''` in the clean Makefile target
2. (Check and) create a fedora-latest based by calling `qvm.default-dispvm.sls`
3. Search for existing DVM templates and choose one

conorsch:
We shouldn't assume that `fedora-30-dvm` exists. It will if a developer
has manually performed the following actions:

  * install fedora-30 template
  * set fedora-30 to the default_template
  * run `qvm.default-dispvm` Salt state

If the system has fedora-30-dvm already present, then it makes sense to
use it as the default DispVM, otherwise we'll just clear out the default
DispVM setting, leaving it blank.

Closes #360

### Testing
If you do _not_ have `fedora-30-dvm` locally:

```
qvm-check fedora-30-dvm # should show "no such domain"
make clone 
make clean
qubes-prefs default_dispvm # should show nothing (i.e. no default)
```

Then run:

```
make clone
make clean
```

And confirm no errors. 

If you _do_ have `fedora-30-dvm`, then:

```
qvm-check fedora-30-dvm # should show "exists"
make clone
make clean
qubes-prefs default_dispvm # should show 'fedora-30-dvm'
```

### DIFF
--- a/dom0/sd-clean-all.sls
+++ b/dom0/sd-clean-all.sls
@@ -3,7 +3,7 @@
 
 set-fedora-as-default-dispvm:
   cmd.run:
-    - name: qubes-prefs default_dispvm fedora-30-dvm
+    - name: qvm-check fedora-30-dvm && qubes-prefs default_dispvm fedora-30-dvm || qubes-prefs default_dispvm ''
 
 {% set gui_user = salt['cmd.shell']('groupmems -l -g qubes') %}
 
--- a/Makefile
+++ b/Makefile
@@ -84,9 +84,10 @@ remove-sd-export: assert-dom0 ## Destroys SD EXPORT VMs
 remove-sd-log: assert-dom0 ## Destroys SD logging VM
 	@./scripts/destroy-vm sd-log
 
-clean: assert-dom0 prep-salt destroy-all ## Destroys all SD VMs
+clean: assert-dom0 prep-salt ## Destroys all SD VMs
 	sudo qubesctl --show-output state.sls sd-clean-all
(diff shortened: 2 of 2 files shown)


conorsch:
Thanks, @emkll, you were certainly correct the logic wasn't operating correctly if `fedora-30-dvm` was missing. I believe this was caused by a _different_ `qubes-prefs` command that's now been cleaned up. Tested both with and without `fedora-30-dvm`, and `make clean` completes successfully in both cases. 

Ready for re-review.

dom0/sd-clean-all.sls
```diff
@@ -3,7 +3,7 @@
 
 set-fedora-as-default-dispvm:
   cmd.run:
-    - name: qubes-prefs default_dispvm fedora-30-dvm
+    - name: qvm-check fedora-30-dvm && qubes-prefs default_dispvm fedora-30-dvm || qubes-prefs default_dispvm ''
```
emkll: This line will likely need to be run before destroy-all in the Makefile

conorsch: Updated: moved the destroy action to _after_ the clean-up-via-salt logic.

conorsch:
> Since the diff is quite small, if you want to squash your commits, go ahead, otherwise feel free to merge.

Rebased on top of latest master, and squashed all commits. Thanks for the careful review on this one. Will merge after CI passes.

PR #372, opened 2019-12-12
Add sd-upgrade-template target to all sd vms

### LINKED ISSUE #371: Makefile targets for specific sd-workstation vms fail
`make sd-svs` and `make sd-gpg` fail because they require the `sd-update-templates` Salt state: 

```
[...]
----------
          ID: sd-gpg
    Function: qvm.vm
      Result: False
                                 require:
                                     sls: sd-upgrade-templates
     Changes:   

Summary for local
-------------
Succeeded: 28 (changed=2)
Failed:     1
-------------
Total states run:     29
Total run time:    3.368 s
DOM0 configuration failed, not continuing
Makefile:31: recipe for target 'sd-gpg' failed
make: *** [sd-gpg] Error 1

```

An easy fix would be to apply the `sd-upgrade-templates` as part of the makefile target, as they should effectively be no-ops.

emkll:
Fixes #371, ensures the makefile targets for each VMs can now provision VMs individually. Because sd-upgrade-templates was not a target, the provisioning would fail.

### Test plan
On `master`:
- `make sd-gpg` fails
- `make sd-whonix` fails
- `make sd-svs` fails

On this branch:
- `make clone` to bring in the changes of this branch into `dom0`
- `make sd-gpg` completes without error, make test passes
- `make sd-whonix` completes without error, make test passes
- `make sd-svs` completes without error, make test passes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,32 +20,32 @@ qubes-rpc: prep-salt ## Places default deny qubes-rpc policies for sd-svs and sd
 	sudo qubesctl --show-output --targets sd-dom0-qvm-rpc state.highstate
 
 sd-workstation-template: prep-salt ## Provisions base template for SDW AppVMs
-	sudo qubesctl --show-output state.sls sd-workstation-template
-	sudo qubesctl --show-output --skip-dom0 --targets sd-workstation-template state.highstate
+	sudo qubesctl --show-output state.sls sd-workstation-buster-template
+	sudo qubesctl --show-output --skip-dom0 --targets sd-workstation-buster-template state.highstate
 
 sd-proxy: prep-salt ## Provisions SD Proxy VM
 	sudo qubesctl --show-output state.sls sd-proxy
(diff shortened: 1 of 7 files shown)


emkll:
Hmm unfortunately that exact error is what this PR should be resolving. Could you confirm:
1. you did `make clone` before running the makefile target on this branch? (updated the test plan to that effect)
2. which Makefile target did you call?

emkll:
Good call, thanks @conorsch , this should now be fixed.

kushaldas:
Strange enough, rerunning the `make sd-whonix` again worked, and all tests passed.

emkll:
Good catch @kushaldas , there was an edge case in the upgrade logic that should be resolved now (sd-proxy would not be shut down when upgrading sd-whonix, if sd-proxy is updated to buster but sd-whonix is not). The upgrade script should now handle that scenario.

PR #379, opened 2019-12-16
More robust config validation logic

### LINKED ISSUE #240: Validate onion URLs before provisioning
The `sd-whonix` VM will fail to start the `tor` process if the onion address is malformed.

conorsch:
Retains the two previous checks:

  1. Does config.json exist?
  2. Does sd-journalist.sec exist?

Adds several checks verifying the contents of config.json.
There's a minimum of user-friendly messages explaining specifically what
the problem is. We can add those messages over time, aiming to balance
utility for Admins and developers.

Assumes that a v2 Onion URL is used. We can update that logic when we
add support for v3 Onion URLs.

Closes #240.

### Testing

- [ ] `make validate` passes in dom0 with v3 Onion URL info
- [ ] `make validate` passes in dom0 with v2 Onion URL info
- [ ] `make validate` fails when a known-bad value is in `config.json` as appending characters to the Onion URL

### DIFF
--- a/scripts/validate-config
+++ b/scripts/validate-config
@@ -8,9 +8,11 @@ import re
 import os
 
 
+TOR_V3_HOSTNAME_REGEX = r'^[a-z2-7]{56}\.onion$'
+TOR_V3_AUTH_REGEX = r'^[A-Z2-7]{52}$'
+
 TOR_V2_HOSTNAME_REGEX = r'^[a-z2-7]{16}\.onion$'
 TOR_V2_AUTH_COOKIE_REGEX = r'^[a-zA-z0-9+/]{22}$'
-
 
 # CONFIG_FILEPATH = "/srv/salt/sd/config.json"
 CONFIG_FILEPATH = "config.json"
@@ -25,8 +27,7 @@ class SDWConfigValidator(object):
         self.config_filepath = CONFIG_FILEPATH
         self.confirm_config_file_exists()
         self.config = self.read_config_file()
-        self.confirm_onion_v2_url()
-        self.confirm_onion_v2_auth()
+        self.confirm_onion_config_valid()
         self.confirm_usb_export_device()
         self.confirm_submission_privkey_file()
         self.confirm_submission_privkey_fingerprint()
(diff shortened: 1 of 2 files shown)


scripts/validate-config
```diff
@@ -0,0 +1,72 @@
...
+        assert re.match(TOR_V2_AUTH_COOKIE_REGEX, self.config["hidserv"]["key"])
+
+    def confirm_usb_export_device(self):
+        assert "usb" in self.config
+        assert "device" in self.config["usb"]
+        usb_config = self.config["usb"]["device"]
+        assert usb_config.startswith("sys-usb")
+
+    def confirm_submission_privkey_file(self):
+        assert os.path.exists("sd-journalist.sec")
```
rmol: How about running  `gpg sd-journalist.sec` to validate its content?

scripts/validate-config
```diff
@@ -0,0 +1,72 @@
...
+        assert "hidserv" in self.config
+        assert "hostname" in self.config["hidserv"]
+        assert re.match(TOR_V2_HOSTNAME_REGEX, self.config["hidserv"]["hostname"])
+
+    def confirm_onion_v2_auth(self):
+        assert "hidserv" in self.config
+        assert "key" in self.config["hidserv"]
+        assert re.match(TOR_V2_AUTH_COOKIE_REGEX, self.config["hidserv"]["key"])
+
+    def confirm_usb_export_device(self):
```
rmol: This is good but won't be necessary after #324 lands.

conorsch:
@rmol Great comments, thanks for taking a look. I've rebased and appended a few changes here, namely to add v3 support, and implement the `gpg` call you suggested. Also updated the test plan to ensure that both v2 & v3 config files work as expected.

> If this is something that will eventually be reporting to journalists, we might want to print nicer messages. If it's just for us, asserts are OK.

Very much dev-focused right now. Deferring a friendlier experience for when we tackle the overall admin provisioning story. At that time, we'll likely crib a fair amount of code from https://github.com/freedomofpress/securedrop/blob/43245829ff93e0a84999526f7ecd0435f89eb52a/admin/securedrop_admin/__init__.py#L60

Ready for re-review!

scripts/validate-config
```diff
@@ -0,0 +1,100 @@
...
+        assert "usb" in self.config
+        assert "device" in self.config["usb"]
+        usb_config = self.config["usb"]["device"]
+        assert usb_config.startswith("sys-usb")
```
rmol: I realized this time through that this doesn't check for a completely valid Qubes USB identifier; it should technically be another regex check for `r'^sys-usb:\d+-\d+$'`. I'm inclined to let it go at this point, given its impending obsolescence, but could also make that change if anyone feels differently.

rmol: Yeah, going to merge this as-is and just get the auto attachment in ASAP.

PR #381, opened 2019-12-17
Fixes #321 Adds v3 onion address support to workstation

### LINKED ISSUE #321: Add support for v3 Journalist Interface onion service
As of SecureDrop 1.0.0, we support v3 onion services, but the workstation/proxy currently only works with v2 addresses for connecting to the Journalist Interface. Because pilot organizations may have switched to v3-only by the time of the pilot, we need to ensure that the workstation can be configured to use a v3 Journalist Interface onion service.

kushaldas:
## Big change

The `config.json` file should only contain either v2 address + key, or v3 address + key.

### For existing installs

- [ ] Update the config.json file
- [ ]  Run `make all`
- [ ] Restart all the sd related vms, and then run `make test`
- [ ] All tests should pass and one v3_auth_private_file test should be skipped
- [ ] Test replying via `securedrop-client` in `sd-svs` 

### For v3 address

- [ ] Update the `config.json` with the v3 onion address and the `key` with the key from the `app-journalist.auth_private` file.

The `app-journalist.auth_private`  file has `onion_address_56_chars:descriptor:x25519:private_key` format. 

- [ ] `make all`.
- [ ] Reboot all related VMs, and then `make test`, all tests should pass.
- [ ] Test replying via `securedrop-client` in `sd-svs` 

This PR will still need one more commit to update the README file. That I will add after we can decide on the right wordings.

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -3,5 +3,5 @@
   "hidserv": {
     "hostname": "avgfxawdn6c3coe3.onion",
     "key": "Il8Xas7uf6rjtc0LxYwhrx"
-  }
+  },
 }
--- a/config.json.example
+++ b/config.json.example
@@ -3,5 +3,5 @@
   "hidserv": {
     "hostname": "avgfxawdn6c3coe3.onion",
     "key": "Il8Xas7uf6rjtc0LxYwhrx"
-  },
+  }
 }
--- a/dom0/sd-whonix-hidserv-key.sls
+++ b/dom0/sd-whonix-hidserv-key.sls
@@ -25,6 +25,6 @@ install-sd-whonix-tor-private-key:
         hostname: {{ hostname_without_onion }}
         key: {{ d.hidserv.key }}
     - mode: 0600
-    - user: root
-    - group: root
+    - user: debian-tor
+    - group: debian-tor
 {% endif %}
--- a/dom0/sd-whonix-hidserv-key.sls
+++ b/dom0/sd-whonix-hidserv-key.sls
@@ -4,7 +4,27 @@
 {% import_json "sd/config.json" as d %}
 
 # add hidden service auth key to torrc
+{% if d.hidserv.hostname|length == 22 %}
 sd-whonix-hidserv-key:
   file.append:
     - name: /usr/local/etc/torrc.d/50_user.conf
     - text: HidServAuth {{ d.hidserv.hostname }} {{ d.hidserv.key }}
+{% else %}
+sd-whonix-hidservv3-directory-path:
+  file.append:
+    - name: /usr/local/etc/torrc.d/50_user.conf
+    - text: ClientOnionAuthDir /var/lib/tor/keys
+
+{% set hostname_without_onion = d.hidserv.hostname.split('.')[0] %}
+install-sd-whonix-tor-private-key:
+  file.managed:
+    - name: /var/lib/tor/keys/app-journalist.auth_private
+    - source: salt://sd/sd-whonix/app-journalist.yaml
+    - template: jinja
+    - context:
+        hostname: {{ hostname_without_onion }}
+        key: {{ d.hidserv.key }}
+    - mode: 0600
+    - user: root
+    - group: root
+{% endif %}
--- a/sd-whonix/app-journalist.yaml
+++ b/sd-whonix/app-journalist.yaml
@@ -1,0 +1,1 @@
+{{ hostname }}:descriptor:x25519:{{ key }}
--- a/tests/test_sd_whonix.py
+++ b/tests/test_sd_whonix.py
@@ -6,7 +6,7 @@ from jinja2 import Template
 from base import SD_VM_Local_Test
 
 
-def skip_if_v3_is_missing():
+def v2_onion_services():
     """
     Returns True if v3 address is not setup
     """
@@ -38,7 +38,7 @@ class SD_Whonix_Tests(SD_VM_Local_Test):
 
             self.assertFileHasLine("/usr/local/etc/torrc.d/50_user.conf", line)
 
-    @unittest.skipIf(skip_if_v3_is_missing(), "Onion v3 address is not setup")
+    @unittest.skipIf(v2_onion_services(), "Onion v3 address is not setup")
     def test_v3_auth_private_file(self):
         with open("config.json") as c:
             config = json.load(c)
--- a/tests/test_sd_whonix.py
+++ b/tests/test_sd_whonix.py
@@ -2,7 +2,20 @@ import unittest
 import json
 from jinja2 import Template
 
+
 from base import SD_VM_Local_Test
+
+
+def skip_if_v3_is_missing():
+    """
+    Returns True if v3 address is not setup
+    """
+    with open("config.json") as c:
+        config = json.load(c)
+        if len(config["hidserv"]["hostname"]) == 22:
+            return True
+        else:
+            return False
 
 
 class SD_Whonix_Tests(SD_VM_Local_Test):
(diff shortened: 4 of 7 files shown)


conorsch:
We didn't discuss in #381, but I'll suggest it now: do we want to _require_ v3 URLs for the Workstation config? It'd simplify the logic presented here, and provide significant encouragement for the v2 -> v3 migration, particularly for the pilot users.

zenmonkeykstop:
That changes the parameters for folks that committed to the pilot, so I'd be loathe to do it for the pilot unless there's a compelling security argument. After the pilot, absolutely.

eloquence:
My preference is also to provisionally keep support for v2 in place for now. We are planning to make a very significant push, with increasing urgency, to help news organizations migrate to v3 through the first months of 2020, and we will certainly bring it up with pilot participants as well, but I don't think it should be a blocker for the pilot.

config.json.example
```diff
@@ -1,7 +1,11 @@
 {
   "submission_key_fpr": "65A1B5FF195B56353CC63DFFCC40EF1228271441",
   "hidserv": {
-    "hostname": "avgfxawdn6c3coe3.onion",
```
emkll: I think we should preserve the .onion here

config.json.example
```diff
@@ -1,7 +1,11 @@
...
+    "hostname": "avgfxawdn6c3coe3",
     "key": "Il8Xas7uf6rjtc0LxYwhrx"
+  },
+  "hidservv3": {
+    "hostname": "ADD_YOUR_VALUE",
+    "key": "ADD_YOUR_VALUE"
```
emkll: Does the Tor hidden service v3 support multiple key formats for v3 ATHS? The key format is hardcoded in https://github.com/freedomofpress/securedrop-workstation/pull/381/files#diff-95ca75b6cf53e335d99d46567e37e9d5R1 , I am wondering if we should expose it to the config

dom0/sd-whonix-hidserv-key.sls
```diff
@@ -4,7 +4,24 @@
...
+
+install-sd-whonix-tor-private-key:
+  file.managed:
+    - name: /var/lib/tor/keys/app-journalist.auth_private
+    - source: salt://sd/sd-whonix/app-journalist.yaml
+    - template: jinja
+    - context:
+        hostname: {{ d.hidservv3.hostname }}
+        key: {{ d.hidservv3.key }}
+    - mode: 0600
```
emkll: nit: since the permissions are 600 here it might be best to specify owner/group similar to https://github.com/freedomofpress/securedrop-workstation/blob/f4ae1e4d9de46092b77dcb16d4f4d940792758ac/dom0/sd-gpg-files.sls#L32-L33

tests/test_sd_whonix.py
```diff
@@ -5,6 +5,18 @@
 from base import SD_VM_Local_Test
 
 
+def skip_if_v3_is_missing():
+    """
+    Returns True if v3 address is not setup
+    """
+    with open("config.json") as c:
+        config = json.load(c)
+        if config["hidservv3"]["hostname"] == "ADD_YOUR_VALUE":
```
emkll: If I am not mistaken, a user copy pasting or using their existing `config.json` (without `hidservv3`) on this branch will cause the tests to fail due to keyerror on hidservv3? 

I don't think we should expect end-users to reconcile their `config.json` with `config.json.example` every time we introduce new configurations generally. What do you think?

dom0/sd-whonix-hidserv-key.sls
```diff
@@ -4,7 +4,24 @@
 {% import_json "sd/config.json" as d %}
 
 # add hidden service auth key to torrc
+{% if d.hidservv3.hostname == "ADD_YOUR_VALUE" %}
 sd-whonix-hidserv-key:
   file.append:
     - name: /usr/local/etc/torrc.d/50_user.conf
-    - text: HidServAuth {{ d.hidserv.hostname }} {{ d.hidserv.key }}
+    - text: HidServAuth {{ d.hidserv.hostname }}.onion {{ d.hidserv.key }}
```
emkll: I think it might be best to keep .onion in the hostname itself, as a random string of characters might be confusing without the .onion suffix, and will help maintainability and avoid confusion. Since the hostname without .onion is a special case for the v3 auth_private, you can use jinja conditionals before `install-sd-whonix-tor-private-key` to create and using a separate variable might work:
```
{% set hostname_without_dot_onion = d.hidservv3.hostname.split('.')[0] %}
```

sd-whonix/app-journalist.yaml
```diff
@@ -0,0 +1 @@
+{{ hostname }}:descriptor:x25519:{{ key }}
```
emkll: :heavy_check_mark: this will be removed everytime `clean-salt` is called via https://github.com/freedomofpress/securedrop-workstation/blob/master/Makefile#L56

tests/test_proxy_vm.py
```diff
@@ -20,10 +20,13 @@ def test_sd_proxy_package_installed(self):
     def test_sd_proxy_yaml_config(self):
         with open("config.json") as c:
             config = json.load(c)
-            hostname = config['hidserv']['hostname']
+            if config['hidservv3']['hostname'] == "ADD_YOUR_VALUE":
```
emkll: see comment above (re: pre-existing config.json)

config.json.example
kushaldas: This is the actual key value, and we are generating it in the vm as required.

kushaldas:
@emkll Please have a look, I updated based on feedback.

dom0/sd-whonix-hidserv-key.sls
```diff
@@ -4,7 +4,27 @@
...
+install-sd-whonix-tor-private-key:
+  file.managed:
+    - name: /var/lib/tor/keys/app-journalist.auth_private
+    - source: salt://sd/sd-whonix/app-journalist.yaml
+    - template: jinja
+    - context:
+        hostname: {{ hostname_without_onion }}
+        key: {{ d.hidserv.key }}
+    - mode: 0600
+    - user: root
```
emkll: It seems like whonix or tor will `chown` these files to `debian-tor:debian-tor`. Perhaps it might make sense to change this here to avoid confusion?

tests/test_sd_whonix.py
```diff
@@ -2,9 +2,22 @@
 import json
 from jinja2 import Template
 
+
 from base import SD_VM_Local_Test
 
 
+def skip_if_v3_is_missing():
```
emkll: nit: this method name in confusing since now we support v2 or v3 within the same config option

config.json.example
```diff
@@ -3,5 +3,5 @@
   "hidserv": {
     "hostname": "avgfxawdn6c3coe3.onion",
     "key": "Il8Xas7uf6rjtc0LxYwhrx"
-  }
```
emkll: Having a comma here results in an error for `sd-proxy-buster-template` (see [1] below). Removing the comma resolves.



[1]
```
sudo qubesctl --show-output --skip-dom0 --targets sd-proxy-buster-template,sd-proxy state.highstate
sd-proxy-buster-template:
      - Rendering SLS 'base:sd-proxy-template-files' failed: Jinja syntax error: Unable to load json from {
          "submission_key_fpr": "65A1B5FF195B56353CC63DFFCC40EF1228271441",
          "hidserv": {
            "hostname": "<something>.onion",
            "key": "hunter2"
          },
        }; line 51
        
        ---
        [...]
            - pkgs:
              - securedrop-proxy
            - require:
              - sls: fpf-apt-test-repo
        
        {% import_json "sd/config.json" as d %}    <======================
        
        install-securedrop-proxy-yaml-config:
          file.managed:
            - name: /etc/sd-proxy.yaml
            - source: salt://sd/sd-proxy/sd-proxy.yaml
        [...]
        ---
sd-proxy: SKIP (nothing to do)
```
failed: Jinja syntax errror: Unable to load json from { <config is here>

conorsch:
@emkll Great comments throughout. Given the changes @kushaldas has already made, we're in rather good shape. I see there are one or two outstanding requests by you, namely:

* ownership on tor auth files
* json syntax

If you want to push up changes to resolve in lieu of @kushaldas (who is AFK), I'm happy to perform final review of the functionality to approve changes by the both of you.

conorsch:
Reviewing...

Issue #368, opened 2019-12-11
`make all` fails on first run due to Whonix VM not being shut down yet


eloquence:
As of #358, on first `make all` before Whonix upgrade, seeing a "QubesVMNotHaltedError: Cannot change template while qube is running". `sys-whonix` step in Makefile fails. Mitigation in https://github.com/freedomofpress/securedrop-workstation/commit/e6798e225c27052d1c3ae1929f56ddb37ebf561e is in place but appears insufficient. Second `make all` run does not hit the same issue as VM is now shut down.

conorsch:
My inclination is to bump the `sleep 5` to `sleep 10` or so. If anyone else has suggestions, please chime in.

emkll:
It seems to me like 5 seconds would be sufficient to kill a vm. I wonder if in the case reported, the vm is being killed but then immediately restarted. If `sys-whonix` is set as upgradeVM for all whonix and whonix-based VMs, maybe it's being restarted immediately after being killed, because it's needed for updating the template. One way to attempt reproduction is have a whonix template vm open and running the whonix upgrade scenario.

This is a one-time upgrade, and the code path should not be reached again, but would be useful to debug this for future template upgrades.

redshiftzero:
closed by #383

PR #383, opened 2019-12-17
Fix make all on clean Qubes 4.0.1 install

### LINKED ISSUE #373: Missing concurrent.futures in sd-whonix
While running `make all` I am getting the following error:

```
sd-whonix:
      ----------
      _error:
          Failed to return clean data
      retcode:
          0
      stderr:
            File "/var/tmp/.root_62a99a_salt/salt-call", line 15, in <module>
              salt_call()
            File "/var/tmp/.root_62a99a_salt/py2/salt/scripts.py", line 395, in salt_call
              import salt.cli.call
            File "/var/tmp/.root_62a99a_salt/py2/salt/cli/call.py", line 5, in <module>
              import salt.utils.parsers
            File "/var/tmp/.root_62a99a_salt/py2/salt/utils/parsers.py", line 27, in <module>
              import salt.config as config
            File "/var/tmp/.root_62a99a_salt/py2/salt/config/__init__.py", line 98, in <module>
              _DFLT_IPC_WBUFFER = _gather_buffer_space() * .5
            File "/var/tmp/.root_62a99a_salt/py2/salt/config/__init__.py", line 88, in _gather_buffer_space
              import salt.grains.core
            File "/var/tmp/.root_62a99a_salt/py2/salt/grains/core.py", line 44, in <module>
              import salt.utils.dns
            File "/var/tmp/.root_62a99a_salt/py2/salt/utils/dns.py", line 32, in <module>
              import salt.modules.cmdmod
            File "/var/tmp/.root_62a99a_salt/py2/salt/modules/cmdmod.py", line 34, in <module>
              import salt.utils.templates
            File "/var/tmp/.root_62a99a_salt/py2/salt/utils/templates.py", line 32, in <module>
              import salt.utils.http
            File "/var/tmp/.root_62a99a_salt/py2/salt/utils/http.py", line 41, in <module>
              import salt.loader
            File "/var/tmp/.root_62a99a_salt/py2/salt/loader.py", line 28, in <module>
              import salt.utils.event
            File "/var/tmp/.root_62a99a_salt/py2/salt/utils/event.py", line 85, in <module>
              import salt.transport.ipc
            File "/var/tmp/.root_62a99a_salt/py2/salt/transport/ipc.py", line 21, in <module>
              from tornado.locks import Semaphore
            File "/var/tmp/.root_62a99a_salt/py2/tornado/locks.py", line 18, in <module>
              from concurrent.futures import CancelledError
          ImportError: No module named concurrent.futures
      stdout:
```

### LINKED ISSUE #382: set management_dispvm property on a per VM basis
Regardless of how #373 is resolved, I think we should set the `management_dispvm` property on each VM we're using for SecureDrop. This will ensure that we are using a management VM based on a template that we know will get frequent updates, i.e. the same one we are using for sys-net and sys-firewall, and thus will have predictable state. Otherwise, the template may remain in the state it was in at the time of install, leading to increasing drift as a function of time which may in turn lead to unpredictable behavior. Using `qubes-prefs` to set the management VM globally is also a fine approach.

emkll:
Closes #373, #368, #360 
Towards #382 (we do not change the default, but ensure the default is updated)
~Supersedes~ towards #367 (because with the changes introduced here, fedora-30-dvm should always be there) 

### Test plan
- [ ] On a fresh qubes 4.0.1 install, a single `make all` should complete in a single pass
- [ ] After running securedrop-update and rebooting, all tests in `make test` should pass
- [ ] `make clean` works as expected
- [ ] there is a fedora-30-dvm vm

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,8 +109,7 @@ After installing Qubes, you must update both dom0 and the base templates to incl
 ```
 sudo qubes-dom0-update
 ```
-
-Finally, update all existing TemplateVMs:
+After dom0 updates complete, reboot your computer to ensure the updates have been properly applied. Finally, update all existing TemplateVMs:
 
 ```
 qubes-update-gui
--- a/dom0/sd-sys-vms.sls
+++ b/dom0/sd-sys-vms.sls
@@ -27,7 +27,7 @@ set-fedora-default-template-version:
       - pkg: dom0-install-fedora-template
       - sls: qvm.default-dispvm
 
-{% for sys_vm in ['sys-usb', 'sys-net', 'sys-firewall'] %}
+{% for sys_vm in ['sys-usb', 'sys-net', 'sys-firewall', 'default-mgmt-dvm'] %}
 {% if salt['cmd.shell']('qvm-prefs '+sys_vm+' template') != sd_supported_fedora_version %}
 sd-{{ sys_vm }}-fedora-version-halt:
   qvm.kill:
--- a/dom0/sd-sys-vms.sls
+++ b/dom0/sd-sys-vms.sls
@@ -4,6 +4,11 @@
 # Ensures that sys-* VMs (viz. sys-net, sys-firewall, sys-usb) use
 # an up-to-date version of Fedora, in order to receive security updates.
 
+include:
+  # Import the upstream Qubes-maintained default-dispvm to ensure fedora-30-dvm
+  # is created.
+  - qvm.default-dispvm
+
 {% set sd_supported_fedora_version = 'fedora-30' %}
 
 # Install latest templates required for SDW VMs.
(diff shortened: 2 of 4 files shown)


ntoll:
Some "user" feedback on my experience of running `make all` from this branch. Not sure if this is helpful or if this is the right place to do it, but wanted to report it, just so it's written down somewhere.

It wasn't apparent to me that my laptop was woefully under-specced with RAM. The Qubes docs mention a minimum of 4Gb RAM with 8Gb preferable. It was only via discussion on Slack that I learned that 16Gb is probably the minimum amount of RAM needed for Qubes for testing/dev of SecureDrop. Caveat: **I may have missed this requirement in the mass of documentation I was reading**. 

In any case, I believe it would be helpful to check the amount of memory when the script starts and abandon with a friendly message if there's NOT enough memory, or warn the user if there's only a limited amount of memory.

In Python you could use the `psutil` module (found on PyPI):

```
import psutil
psutil.virtual_memory()
```

... or perhaps you could grep something from `cat /proc/meminfo` (first line of the result is the total number of KB of RAM)..?

If there's no such guard, the script chugs along into out of memory errors (see screenie below) after taking a long time to download a bunch of stuff. 

wat_qubes

Such a memory size check might save someone some time in the future.

emkll:
great points @ntoll . Unfortunately dom0 itself is a virtual machine, so it is likely that it will return the maximum of the default Qubes vm, which is 4GB, regardless of how much physical memory your laptop has. I think we could get this information through the `xentop` utility, which would also be a good tool to debug the status of the install/system memory pressure.

Regarding the progress, this is the verbose output that salt provides, the challenge here is that it is presented to the user after the state is applied, which I understand is a limitation of Salt, as the commands are running from within the vm itself, and output sent back to dom0 after completion. This is another point in favor of using another tool for provisioning (see https://github.com/freedomofpress/securedrop-workstation/issues/159)

PR #396, opened 2020-01-06
Enforce updates on login via dom0 launcher app

### LINKED ISSUE #238: Replace custom upgrade logic with qubes-gui-updater
The custom workstation update process described in #24 has now been largely superseded by the `qubes-gui-updater` tool in dom0. See screenshot:

qubes-gui-updater

One major problem is that the `sd-` templates never report any updates. Why is this? Since they were created via cloning the debian-9 TemplateVM, I'd expect the requisite qubes-specific tooling to be present on disk. More research required.

Even if we can leverage the `qubes-gui-updater` tool for use in upgrading the SDW VMs, we'll still likely want _unattended_ upgrades to ensure security fixes are applied in a timely manner.

### LINKED ISSUE #341: Enforce `sd-svs-disp` updates on login
We'd like to use Qubes' native facilities for templates updates to the greatest extent possible. This is #238. The native updater is an opt-in updater: it requires the user to actively launch the updater app.

Opt-in is not sufficient for the most security-sensitive template, `sd-svs-disp`, which is used for the creation of disposable VMs that contain viewer applications. In this case, we want to enforce updates.

This is a feature proposal to add an updater that launches on login in a manner that clearly indicates to the user that `sd-svs-disp` updates must be run before using the client. 

## Proposed Acceptance Criteria

**Given** that the Securedrop Workstation has just been booted
**When** I log into Qubes with my username/password
**Then** I should see a dialog that informs me that security updates must be applied before using SecureDrop
**And** it should give me the option to start the update
**And** no updates should start until I choose this option
**And** it should not be trivially possible to dismiss this dialog

**Given** that the "update required" dialog is open
**When** I start the update
**Then** security updates should be applied to `sd-svs-disp` template
**And** it should be obvious when this process is ongoing
**And** it should be obvious when this process is completed.

## User Stories

- As a journalist user, I want to be informed clearly and succinctly about any steps I need to take before using the SecureDrop Workstation, so that I can focus on my work to the greatest extent possible.

- As a journalist user, I want my system to be up-to-date with security updates, so my sources and I are secure.

emkll:
closes #341, closes #238

The goal is to enforce critical VM (incl dom0) updates on client login. For now, the interface is launched via the .desktop shortcut, but we also have this run on startup (similar to how updates are handled in #355)

### Test Plan
- make clone on this branch
- [ ] `make prep-dom0` succeeds without error
- [ ] clicking on Desktop icon spawns the launcher
- [ ] Logs for the updater appear ~/.securedrop-launcher/logs/launcher.log
- The following assumes your workstation is already up-to-date since 
#### Scenario 1: All VMs up to date:
- [ ] The launcher simply runs, checks for updates, no updates are required, the client can be launched.
- [ ] sd-svs flags are in place (`/home/user/.securedrop.client/{sdw-last-updated,sdw-update-flag}` are correct
- [ ] dom0 flags (`~/.securedrop-launcher/{sdw-last-updated,sdw-update-flag}` are correct

#### Scenario 2: Reboot required
- Open fedora-30, downgrade a package and shutdown the template (e.g: `sudo dnf downgrade zlib && sudo poweroff`)
- [ ] The launcher indicates update is required
- [ ] Updates are applied
- [ ] The launcher recommends a user reboot
- [ ] sd-svs flags are in place (`/home/user/.securedrop.client/{sdw-last-updated,sdw-update-flag}` are correct
- [ ] dom0 flags (`~/.securedrop-launcher/{sdw-last-updated,sdw-update-flag}` are correct
- [ ] after reboot, Scenario 1 completes without error

#### Scenario 3: Updates required but no reboot required
- Open sd-svs-buster-template and downgrade securedrop-client (`sudo apt install securedrop-client=<version>` and shutdown the template
- [ ] The launcher indicates update is required
- [ ] Updates are applied
- [ ] The launcher recommends you start the client
- [ ] sd-svs flags are in place (`/home/user/.securedrop.client/{sdw-last-updated,sdw-update-flag}` are correct
- [ ] dom0 flags (`~/.securedrop-launcher/{sdw-last-updated,sdw-update-flag}` are correct
- [ ] The client starts on the latest version.

#### Follow-ups
1. Improve the UI theme/styling, (e.g., add SecureDrop logo)
2. Parallelize updates (2x concurrency)
3. Start the Launcher on boot, similar to #355 
4.  Use dom0 date flag to skip checks if check was done less than 8h ago
5. remove daily `securedrop-update-cron` once 4. is completed (per https://github.com/freedomofpress/securedrop-workstation/pull/389#issuecomment-573910755)
6. Only reboot AppVMs that need rebooting (AppVMs whose TemplateVM was updated as part of this run, see https://github.com/freedomofpress/securedrop-workstation/pull/396#discussion_r367724007)


launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -0,0 +1,256 @@
...
+                results[vm] = True
+                progress = {"output": str(e), "run_completed": False}
+                self.signal.emit(progress)
+
+        result = {
+            "run_completed": True,
+            "action": "upgrade",
+            "dom0_upgrade_success": results.get("dom0", True),
+            "sd-svs-disp_upgrade_success": results.get("sd-svs-disp", True),
+            "fedora_upgrade_success": results.get("fedora", True),
```
redshiftzero: why `results.get("fedora", True)` here? do we want to use a default value of True? (if this is indeed what you want, let's add an explanatory comment)

launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -0,0 +1,256 @@
...
+                        )
+                        self.logOutputText += "Failed to write file to sd-svs, please contact your administrator"
+                        self.logOutput.setText(self.logOutputText)
+            elif result["action"] == "upgrade":
+                self.proposedActionLabel.setText(
+                    "Updates have been applied, system is now up-to-date."
+                    # TODO: Reboot if xen updates, don't remove flag for now
+                )
+                self.progressBar.setProperty("value", 100)
+                self.clientOpenButton.setEnabled(True)
```
redshiftzero: need `self.applyUpdatesButton.setEnabled(False)` here

launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -0,0 +1,256 @@
...
+
+        self.logOutputText += str(result)
```
redshiftzero: Nit: after each write to `self.logOutputText` we need a line break otherwise all the output is concatenated (if you want you could add a helper method on UpdaterApp takes a string, adds a line break, and calls `self.logOutput.setText`)

dom0/securedrop-updater-cli
```diff
@@ -0,0 +1,231 @@
...
+            sdlog.debug("Checking for sd-svs-disp updates")
+            sys.exit(check_updates_sd_svs_disp())
+        elif "fedora" in args.check:
+            sdlog.debug("Checking for fedora updates")
+            sys.exit(check_updates_fedora())
+        elif "airgapped" in args.check:
+            sdlog.debug("Checking for updates for airgapped VMs")
+            rc = check_updates_airgapped()
+            try:
+                subprocess.Popen("qubes-update-gui")
```
redshiftzero: am i right that `qubes-update-gui` will open even if updates aren't needed currently?

dom0/securedrop-updater-cli
```diff
@@ -0,0 +1,231 @@
...
+import logging
+import subprocess
+import sys
+
+sdlog = logging.getLogger(__name__)
+# these are the TemplateVMs that require full patch level at boot in order to start the client
+possible_vms = ["dom0", "sd-svs-disp", "fedora", "airgapped"]
+# these are the TemplateVMs that for which the AppVMs are airgapped.
+# The Qubes Update manager cannot check for updates, we must do so manually.
+airgapped_vms = ["sd-svs-buster-template", "sd-log-buster-template", "securedrop-workstation-buster"]
```
redshiftzero: also `sd-export-buster-template` or is this intentionally a subset?

launcher/sdw_updater_gui/sdw_updater.ui
```diff
@@ -0,0 +1,92 @@
...
+    <item row="1" column="0" colspan="3">
+     <widget class="QTextBrowser" name="logOutput">
+      <property name="acceptDrops">
+       <bool>false</bool>
+      </property>
+      <property name="acceptRichText">
+       <bool>false</bool>
+      </property>
+      <property name="openLinks">
+       <bool>false</bool>
```
redshiftzero: :+1:

redshiftzero:
note to other people trying this that right now the strings listed in https://github.com/freedomofpress/securedrop-workstation/issues/341#issuecomment-555754652 aren't used - this PR thus far is just a demo of the basic functionality, we can update the text shown to the user (and probably hide the window that has the full details of what is happening) once we agree on the basic functionality/technical approach

dom0/securedrop-updater-cli
```diff
@@ -0,0 +1,255 @@
...
+sdlog = logging.getLogger(__name__)
+# these are the TemplateVMs that require full patch level at boot in order to start the client
+possible_vms = [
+    "dom0",
+    "sd-svs-disp",
+    "fedora",
+    "sd-svs",
+    "sd-log",
+    "sd-export",
+    "securedrop-workstation",
```
redshiftzero: we could just access the keys of the current_templates dict to build this list

dom0/securedrop-updater-cli
```diff
@@ -0,0 +1,255 @@
...
+        elif "sd-svs" in args.check:
+            sdlog.debug("Checking for sd-svs updates")
+            sys.exit(check_updates_debian("sd-svs"))
+        elif "sd-log" in args.check:
+            sdlog.debug("Checking for sd-log updates")
+            sys.exit(check_updates_debian("sd-log"))
+        elif "sd-export" in args.check:
+            sdlog.debug("Checking for sd-export updates")
+            sys.exit(check_updates_debian("sd-export"))
+        elif "securedrop-workstation" in args.check:
```
redshiftzero: reduce code duplication here

dom0/securedrop-updater-cli
```diff
@@ -0,0 +1,255 @@
...
+            apply_updates_vm("sd-export")
+        elif "securedrop-workstation" in args.check:
+            sdlog.debug("Checking for securedrop-workstation")
+            apply_updates_vm("securedrop-workstation")
+        else:
+            sdlog.debug("Nothing to do")
+            return 2
+
+
+def check_updates_dom0():
```
redshiftzero: make `check_updates` wrapper function

launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -0,0 +1,301 @@
+from PyQt4 import QtGui
+from PyQt4.QtCore import QThread, pyqtSignal
+from sdw_updater_gui.UpdaterAppUi import Ui_UpdaterDialog
+from sdw_updater_gui import strings
+import logging
+import subprocess
+import sys
+
+VMS = (
```
redshiftzero: duplicated from above, the cli could return this

launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -0,0 +1,301 @@
...
+        self.rebootButton.setEnabled(False)
+        self.rebootButton.clicked.connect(self.reboot_workstation)
+        self.applyUpdatesButton.setEnabled(False)
+        self.applyUpdatesButton.clicked.connect(self.apply_all_updates)
+
+        self.show()
+
+        # Both these conditions must be true to have the client started:
+        # 1. critical updates to VMs (dom0, fedora and sd-svs-disp)
+        # 2. check for updates for other vms and present the qubes GUI updater
```
redshiftzero: update comment

launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -0,0 +1,301 @@
...
+        self.updates_required = False
+        self.updates_applied = False
+
+        self.proposedActionLabel.setText(strings.label_status_checking_updates)
+        self.proposedActionDescription.setText(
+            strings.description_status_checking_updates
+        )
+        # start sd-svs in the background
+        try:
+            logging.info("starting sd-svs" + "now")
```
redshiftzero: svs starts on boot so we can remove

launcher/sdw_updater_gui/UpdaterAppUi.py
```diff
@@ -0,0 +1,92 @@
...
+class Ui_UpdaterDialog(object):
+    def setupUi(self, UpdaterDialog):
+        UpdaterDialog.setObjectName(_fromUtf8("UpdaterDialog"))
+        UpdaterDialog.resize(520, 330)
+        self.layoutWidget = QtGui.QWidget(UpdaterDialog)
+        self.layoutWidget.setGeometry(QtCore.QRect(20, 20, 480, 300))
+        self.layoutWidget.setObjectName(_fromUtf8("layoutWidget"))
+        self.gridLayout = QtGui.QGridLayout(self.layoutWidget)
+        self.gridLayout.setMargin(0)
+        self.gridLayout.setObjectName(_fromUtf8("gridLayout"))
```
redshiftzero: nit: small thing but we should make it such that you cannot maximize the window (can be done in followup)

eloquence:
I may be catching this branch at a bad time but just noting that the progress bar is currently frozen at 1% for me during the "checking for updates" stage (it immediately switches to 100% once the checks are complete).

Screenshot_2020-01-09_17-13-36

eloquence:
(Regarding the icon, I noticed that it's possible to adjust the Qubes icon settings to a higher size, which also means more text for the label. It's set to 64 pixels in the above screenshot. If it's possible to change that preference during provisioning, that may be a nice follow-up tweak.)

eloquence:
I'm noticing that every time I run the launcher, the VMs get the green "restart required" indicator in the Qube manager widget (see top right in the screenshot below), even if no updates were available.

To verify this behavior, I did restart the `sd-svs` VM, which removed the indicator, then ran the launcher (via shortcut). Then I checked the Qube manager widget, and the green "restart required" indicator was back.

Screenshot_2020-01-09_17-33-55

launcher/tests/test_updater.py
```diff
@@ -0,0 +1,399 @@
...
+            "echo '{}' > {}".format(status.value, updater.FLAG_FILE_STATUS_SD_SVS),
+        ]
+    )
+    assert not mocked_error.called
+
+
+@mock.patch("subprocess.check_call")
+@mock.patch("Updater.sdlog.error")
+@mock.patch("Updater.sdlog.info")
+def test_write_last_updated_flags_to_disk(mocked_info, mocked_error, mocked_call):
```
redshiftzero: optional: use tmpdir here (and in other tests that will write files to user's homedir)

launcher/sdw_updater_gui/Updater.py
```diff
@@ -0,0 +1,350 @@
...
+import os
+import subprocess
+from datetime import datetime
+from enum import Enum
+
+DEFAULT_HOME = os.path.join(os.path.expanduser("~"), ".securedrop_launcher")
+
+FLAG_FILE_STATUS_SD_SVS = "/home/user/sdw-update-flag"
+FLAG_FILE_LAST_UPDATED_SD_SVS = "/home/user/sdw-last-updated"
+FLAG_FILE_LAST_UPDATED_DOM0 = os.path.join(DEFAULT_HOME, "sdw-last-updated")
```
redshiftzero: one other thought: can we also store the content of `FLAG_FILE_STATUS_SD_SVS` in dom0 somewhere? The reason is that we need from dom0 to read `/home/user/sdw-update-flag` in `sd-svs` (which might be shutdown) to distinguish when the user hasn't updated their workstation for 5+ days but just because they just started their workstation after a 5 day period ([case B here](https://docs.google.com/document/d/1cCpcpLRiKC0tm_vsD4k4Q1Fr5_R34htCb8FNTV8mRCI/edit)).

emkll:
This is now ready for review, I have updated the test plan. Thanks @redshiftzero for all your comments, I believe they have all been addressed.

launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+class UpdateThread(QThread):
+    """
+    This thread will check for TemplateVM updates
+    """
+
+    update_signal = pyqtSignal("PyQt_PyObject")
+    progress_signal = pyqtSignal("int")
+    progress_callback = None
```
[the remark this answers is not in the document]
emkll: We are no longer using it! removed.

launcher/sdw_updater_gui/Updater.py
```diff
@@ -0,0 +1,360 @@
...
+        )
+    except subprocess.CalledProcessError as e:
+        sdlog.error("Error writing update status flag to sd-svs")
+        sdlog.error(str(e))
+
+    try:
+        sdlog.info("Setting update flag to {} in dom0".format(status.value))
+        if not os.path.exists(os.path.dirname(flag_file_path_dom0)):
+            os.makedirs(os.path.dirname(flag_file_path_dom0))
+        f = open(flag_file_path_dom0, "w+")
```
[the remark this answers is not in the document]
emkll: thanks! fixed

launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -0,0 +1,249 @@
...
+    def run(self):
+        upgrade_generator = Updater.apply_updates(self.vms_to_upgrade)
+        results = {}
+        while True:
+            try:
+                vm, progress, result = next(upgrade_generator)
+                results[vm] = results
+                self.progress_signal.emit(progress)
+            except StopIteration:
+                break
```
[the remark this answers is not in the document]
emkll: I agree, definitely way more readable. I have applied this change, thanks

launcher/sdw_updater_gui/Updater.py
```diff
@@ -0,0 +1,356 @@
...
+from the parent directory.
+"""
+
+import logging
+import os
+import subprocess
+from datetime import datetime
+from enum import Enum
+
+FLAG_FILE_STATUS_SD_SVS = "/home/user/sdw-update-flag"
```
redshiftzero: nit/optional: worth dropping these sd-svs flags into `/home/user/.securedrop_client/*` to prevent a user from deleting

dom0/securedrop-launcher.desktop
```diff
@@ -0,0 +1,7 @@
+[Desktop Entry]
+Version=1.0
+Type=Application
+Terminal=false
+Icon=/usr/share/securedrop/icons/sd-logo.png
+Name=SecureDrop Workstation Launcher
```
eloquence: @ninavizz and I would recommend changing this icon label simply to "SecureDrop". The preflight updater will or will not run in some situations.

eloquence:
I've pushed a commit that updates the dialog text to language that Nina and I think is reasonable for the pilot; hope that language works for your as well. :) 

Two suggested issues to poke at prior to merge:
- One significant UX issue I've observed in one update is that during the download/apply updates stage, the progress bar hit 100% but the process continued for another minute or longer. Note that this particular update ultimately failed due the `fedora-30` kernel package issue, and I've not run another update since then (because there are no updates! ;-).
- Upon consideration, we would recommend _not_ showing the progress bar at 100% on the "Updates required" and "No updates today" screens. It is potentially confusing to show a bar at 100% without a context as to what it refers to. For now, the simplest solution is to simply hide it on those screens, and to show it again during the download/apply stage. If this results in whitespace in that section of the dialog, so be it.

We may want to consider some tweaks to the buttons as a follow-up, but those shouldn't hold up merge of this PR.

conorsch:
In order to unblock merge, I'm dismissing @kushaldas's prior review. Have confirmed that the changes requested were indeed addressed. Great attention to detail in that review—thanks, @kushaldas!

launcher/sdw_updater_gui/Updater.py
```diff
@@ -0,0 +1,356 @@
...
+        return UpdateStatus.UPDATES_FAILED
+    elif reboot_required:
+        return UpdateStatus.REBOOT_REQUIRED
+    elif updates_required:
+        return UpdateStatus.UPDATES_REQUIRED
+    else:
+        return UpdateStatus.UPDATES_OK
+
+
+def _shutdown_and_start_vms():
```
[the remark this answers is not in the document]
redshiftzero: now in https://github.com/freedomofpress/securedrop-workstation/issues/457

Issue #415, opened 2020-01-22
Start updater at boot, remove existing svs-disp backgound updater


emkll:
In https://github.com/freedomofpress/securedrop-workstation/pull/396, we have introduced a launcher to provide a way for users to control the update process. We would like the checks to happen as soon as the system boots to desktop.

We should re-use the autostart logic provided in [1] by replacing the background dispvm template update script (securedrop-login) by the QT application introduced in #396  

[1] https://github.com/freedomofpress/securedrop-workstation/pull/355/files#diff-19fc9e35327834b65d0d380e0da1a89bR19

eloquence:
This could be split into two iterations:
-first iteration: replace `securedrop-login` as described
-second iteration: ensure that only one updater can be launched

eloquence:
Regarding launching the preflight updater upon login, I admit to being somewhat ambivalent. Having two dueling paths to launch the app (login & desktop icon) could be confusing, and having the client login window pop up in every session feels a bit constraining for future use cases that may be independent of the client app. 

That said, since it's best to ensure that the workstation is up-to-date regardless of what the user does with it, I agree this is the way to go at this time. Adding the first two iterations to the sprint.

eloquence:
> third iteration: ensure that only one updater can be launched

This was actually done in #445.

PR #322, opened 2019-11-08
Removes auto-updater from workstation

kushaldas:
This PR removes the securedrop autoupdater. 

We can get it back in future, for now use the Qubes
updater UI.

## How to test?

In `dom0` after getting this branch.

```
make prep-dom0
```

Issue #378, opened 2019-12-16
qubesd gone missing during `make all`


kushaldas:
I was running `make all` from master, and during the process, the `securedrop-updater` also started, and I got the following error.

```
  Summary for sd-svs-disp-buster-template
  ------------
  Succeeded: 9
  Failed:    0
  ------------
  Total states run:     9
  Total run time:  22.206 s
sd-proxy-buster-template: ERROR (exception Failed to connect to qubesd service: [Errno 2] No such file or directory)
Traceback (most recent call last):
  File "/bin/qubesctl", line 100, in <module>
    sys.exit(main())
  File "/bin/qubesctl", line 87, in main
    return max(exit_code, runner.run())
  File "/usr/lib/python2.7/site-packages/qubessalt/__init__.py", line 265, in run
    if 'state.highstate' not in self.command or self.has_config(vm):
  File "/usr/lib/python2.7/site-packages/qubessalt/__init__.py", line 258, in has_config
    top = caller.function('state.show_top')
  File "/usr/lib/python2.7/site-packages/salt/client/__init__.py", line 2071, in function
    return func(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/salt/modules/state.py", line 1616, in show_top
    info=st_.opts['pillar']['_errors'])
salt.exceptions.CommandExecutionError: Pillar failed to render. Additional info follows:

- Failed to load ext_pillar qvm_prefs: Failed to connect to qubesd service: [Errno 2] No such file or directory
Makefile:11: recipe for target 'all' failed
make: *** [all] Error 1
```

eloquence:
Ah yeah, this just bit me as well. Indeed seems to be due to the updater cron job running in the middle.

kushaldas:
This time the `sd-app` could not come up due to qrexec service not responding. Had to remove the vm and redo the steps to get it working.

conorsch:
> seems to be due to the updater cron job running in the middle

Given that the legacy updater was removed in #430, and we haven't received any more reports of this particular problem, I'm inclined to close. Also worth noting that 

> Failed to load ext_pillar qvm_prefs

may be helped by #530, but that's a bit of a stretch.

eloquence:
Unfortunately that hypothesis cannot be fully sufficient as I encountered it again since the legacy updater removal;  see https://github.com/freedomofpress/securedrop-workstation/issues/514#issuecomment-614358812

conorsch:
[...]

PR #430, opened 2020-01-23
Remove legacy updater

### LINKED ISSUE #412: Remove update cron job
Now that we have #396 merged, once #389 or some improved version thereof is also merged (to ensure users reboot their workstations after some time), we can remove the daily cron job which runs `securedrop-update`.

Also blocked on #427.

eloquence:
## Status

Ready for review

## Description

Resolves #412

The preflight updater is now at feature parity with the cron job that fires every 24 hours to update VMs:
- We enforce updates before launching the client (#396)
- We remind the user if updates have not been applied recently (#445)
- We apply the dom0 state as part of the update process (#458)

Therefore, it is safe to remove the cron job at this time. This PR preserves the `securedrop-login` script, which enforces updates to the `sd-viewer` template on login, until #415 is resolved.

**NOTE:** Because it changes the contents of the RPM package, this PR bumps the RPM version.

### Preparatory steps

- Make sure you are using a workstation that has been previously provisioned from a different branch that includes the updater (e.g., `master`)
- Check out this branch in your dev VM, `make clone` it into `dom0`

### Test plan 
- Run `make clean`
- Confirm the following files do _not_ exist (`make clean` logic is still removing them)
  - [ ] `/srv/salt/securedrop-update`
  - [ ] `/usr/bin/securedrop-update`
  - [ ] `/etc/cron.daily/securedrop-update-cron` (symlink)
- Run `make all`
- Confirm the following files do _not_ exist (`make clean` logic is still removing them)
  - [ ] `/srv/salt/securedrop-update`
  - [ ] `/usr/bin/securedrop-update`
  - [ ] `/etc/cron.daily/securedrop-update-cron` (symlink)

## Checklist
- [x] `make flake8` passes
- [ ] `make test` passes in dom0
    - I am seeing mismatches in the RPC policy test, which appear to be unrelated to this PR, which does not touch RPC policies.

## Attribution

This PR is based on #322, originally by @kushaldas.


dom0/sd-dom0-files.sls
```diff
@@ -64,20 +64,16 @@ dom0-install-securedrop-workstation-template:
...
-    - user: root
-    - group: root
-    - mode: 755
 
-# Symlink update script into cron, for single point of update
+
+# Remove symlink to legacy updater
 dom0-update-securedrop-script-cron:
-  file.symlink:
+  file.absent:
```
eloquence: Note: may want to enforce removal of `/srv/salt/securedrop-update` as well

eloquence:
I've tested this PR and can confirm that it works as intended; until we land it, patching from this branch should be a viable workaround if you want to provision a workstation without the legacy updater getting in the way (#378) or conflicting with the preflight updater (#396).

eloquence:
To cherry-pick this change without creating a new commit, as part of provisioning, you can use this command on the `master` branch:

```
git cherry-pick -n <commit hash>
```

The change will be staged but not committed:

```
git diff --cached
```

Until we're ready to merge this, I'll keep it in sync with `master`.

dom0/sd-clean-all.sls
```diff
@@ -12,9 +12,7 @@ remove-dom0-sdw-config-files:
     - names:
       - /opt/securedrop
       - /etc/yum.repos.d/securedrop-workstation-dom0.repo
-      - /usr/bin/securedrop-update
       - /etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation-test
-      - /etc/cron.daily/securedrop-update-cron
```
conorsch: Because the clean action ensures state=absent, it actually makes sense to preserve the entries here. Note that `/srv/*` items may still need to be removed, either here or via `clean-salt`.

dom0/sd-dom0-files.sls
conorsch: Nit: since we're now enforcing state=absent for three different files, we can collapse into a single task:

```
dom0-remove-cron-updater-logic:
  file.absent:
    - names:
      - /usr/bin/securedrop-update
      - /etc/cron.daily/securedrop-update-cron
```

Regarding the `/srv/salt/securedrop-update` removal, that's best handled in the "clean" action above. (In prod, such removals will be handled by the RPM package spec.)

dom0/sd-clean-all.sls
eloquence: Done.

eloquence:
Since this changes the package, it should bump the release -- I've not done that for now, but will before promoting to "ready", once other dependencies for this PR are met.

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -43,16 +43,14 @@ install -m 755 -d %{buildroot}/srv/salt/sd/sd-workstation
 install -m 755 -d %{buildroot}/srv/salt/sd/sys-firewall
 install -m 755 -d %{buildroot}/usr/share/%{name}/scripts
 install -m 755 -d %{buildroot}/srv/salt/sd/usb-autoattach
+install -m 755 -d %{buildroot}/%{_bindir}
```
eloquence: This is necessary as the directory was previously created as part of the `setup.py` run.

eloquence: (As an aside,  `%{_bindir}` has a leading slash, so those sorts of invocations resolve to `//`. That's a noop, but can omit the unnecessary `/` - which is also used in the original code - if that's preferred.)

eloquence:
(Rebased to fix conflicts.)

Issue #406, opened 2020-01-13
Use rpm in dom0 for make all


redshiftzero:
We have an rpm package that contains the packaged configuration of dom0 but we're not using it. We need three environments:

* prod, rpm signed with release key (used for provisioning pilot machines and should be the default)
* staging, rpm signed with test key, used for QA / testing or development on subcomponents of the securedrop workstation where the configurations in dom0 do not need modification
* dev, does not use rpm (behavior of current `make all`)

proposal:
* `make all` -> will become prod (we can point this to staging until the first production release in the beta series occurs)
* `make staging` -> exactly like make all except uses the nightly rpms
* `make dev` -> preserves current behavior of `make all`

### Specific next steps
- [x] Create RPM backend: https://yum.securedrop.org/ (we already have `yum-test`)
- [x] Update scripts in LFS repos to perform signing
  - [x] In [dev LFS repo](https://github.com/freedomofpress/securedrop-workstation-dev-rpm-packages-lfs/), signing should be automatic with test key in CI; cf. https://github.com/freedomofpress/securedrop-debian-packaging/pull/129
  - [x] In [prod LFS repo](https://github.com/freedomofpress/securedrop-workstation-prod-rpm-packages-lfs/), signing should be manual action by human maintainers
- [x] Add conditional Salt logic for dev/prod yum repo URLs, _default to prod_
- [x] Add conditional Salt logic for dev/prod apt repo URLs, _default to dev_
- [x] Update make targets as described above (all, staging, dev)
- [x] Provide admin-facing wrapper for enforcing state, e.g. `/usr/local/bin/securedrop-install`

We'll also need to consolidate apt-test-qubes into the apt-test repo, now that we have dual-channel support, but that should be tracked in a separate ticket, and isn't critical-path for the RPM story.

conorsch:
> `make dev` -> preserves current behavior of `make all`, **i.e. copies files from the repo to target locations, and ensures the dom0 config RPM is absent**

Then this sounds pretty good. Note that pending a solution of the RPM signing story for nightly/staging vs prod, we'll have to update the config logic _at least_ to have different URL endpoints for prod/dev. If we decide to switch to signing repo metadata rather than the RPMs themselves, then we'll have to update the yum configs with `gpgcheck=0` and `repo_gpgcheck=1`.

conorsch:
> If we decide to switch to signing repo metadata rather than the RPMs themselves

[...]

PR #432, opened 2020-01-25
Begins dynamic dev/prod config logic

conorsch:
## Status
Ready for review

## Description of Changes

Refs #406 

* Adds new `config.json` value "environment", accepts "dev" or "prod"
* Updates Salt logic to dynamically use dev/prod vars, based on `config.json` "environment"
* Converts URLs `https://apt-test-qubes.freedom.press` -> `https://apt-test.freedom.press`

There are a few caveats with the current implementation:

* `config.json` is modified _on-the-fly_ as part of e.g. `make prod` / `make dev`. Is that what we want?
*  The prod URLs don't actually have stable, signed packages yet—so full testing isn't possible

### dev target
```
make clone
make clean
make all
make test
```

### staging target
**Only run on test hardware.** We don't want to install CI-built artifacts in dom0.

```
make clone
make clean
make staging # confirm prompt
make test
```

- [ ] Linting (`make flake8`) and tests (`make test`) pass in dom0 of a Qubes install

- [ ] I have added/removed files, and have updated packaging logic in MANIFEST.in and rpm-build/SPECS/securedrop-workstation-dom0-config.spec


emkll:
- Added `securedrop-admin` logic to staging and prod scenarios (the /usr/bin/ file is brought in exclusively by the RPM), reusing existing logic from the `scripts/` directory.
- Added conditional logic to install securedrop-workstation-dom0-config for staging and prod environments
- Added a test-only staging scenario, which should mirror prod, but use test-specific URLs.
- Fixed the tests

Note that the staging environment should be used on test hardware only, and provisions the workstation VMs and their environment based on the config that's in the RPM itself (initial dom0 logic is ensured by the Makefile target). Dom0 is configured by installing the package, and the package will replace the files in /srv/salt in place, during the run. 

To prevent a race condition, once an RPM is is uploaded with changes from this branch (with the securedrop-admin wrapper and script it place, we should apply the following diff, for staging and prod targets):
```
diff --git a/Makefile b/Makefile
index 9e83f0611..585e783f7 100644
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,8 @@ staging: ## Configures a STAGING install. To be used on test hardware ONLY
        ./scripts/configure-environment --env staging
        $(MAKE) validate
        $(MAKE) prep-salt
-       ./scripts/provision-all
+       $(MAKE) prep-dom0
+       /usr/bin/securedrop-admin --apply
 
 dom0-rpm: ## Builds rpm package to be installed on dom0
        @./scripts/build-dom0-rpm
```

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -45,12 +45,14 @@ install -m 755 -d %{buildroot}/usr/share/%{name}/scripts
...
 install -m 644 dom0/*.top %{buildroot}/srv/salt/
 install -m 644 dom0/*.j2 %{buildroot}/srv/salt/
+install -m 644 dom0/*.yml %{buildroot}/srv/salt/
 install -m 644 dom0/securedrop-update %{buildroot}/srv/salt/
 install -m 644 dom0/securedrop-login %{buildroot}/srv/salt/
 install -m 644 dom0/securedrop-launcher.desktop %{buildroot}/srv/salt/
 install -m 655 dom0/securedrop-handle-upgrade %{buildroot}/srv/salt/
 # The next file should get installed via RPM not via salt
 install -m 755 dom0/securedrop-update %{buildroot}/srv/salt/securedrop-update
+install -m 755 dom0/securedrop-admin %{buildroot}/%{_bindir}/securedrop-admin
```
conorsch: Instead of:

```
install -m 755 dom0/securedrop-admin %{buildroot}/%{_bindir}/securedrop-admin
```

why not

```
install -m 755 scripts/securedrop-admin.py %{buildroot}/%{_bindir}/securedrop-admin
```

? N.B. Haven't tested, mostly questioning the need for the wrapper script indirection, as above.

scripts/securedrop-admin.py
```diff
@@ -0,0 +1,91 @@
...
+    try:
+        subprocess.check_call(
+            ["sudo", "cp", os.path.join(SCRIPTS_PATH, "config.json"), SALT_PATH]
+        )
+        subprocess.check_call(
+            ["sudo", "cp", os.path.join(SCRIPTS_PATH, "sd-journalist.sec"), SALT_PATH]
+        )
+    except subprocess.CalledProcessError as e:
+        print(str(e))
+        sys.exit(1)
```
conorsch: Why catch, then sys.exit? If we simply allow the exception to be raised, we'll still have the script halt execution, as well as exit non-zero. Does the syntax here prevent preventing the traceback? (Even if it does, not sure it's worth the extra lines.)

scripts/securedrop-admin.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def provision_all():
+    """
+    Runs provision-all to apply the salt state.highstate on dom0 and all VMs
+    """
+    try:
+        subprocess.check_call([os.path.join("./scripts/provision-all")])
```
conorsch: Indirection, oh my!! Nit: shouldn't there be a SCRIPTS_PATH in the .join()? Would prefer that we integrate provision-all directly into securedrop-admin, but now is not the time... looking forward to circling back and converting to Python Qubes API wherever possible.

dom0/securedrop-admin
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e
+set -u
+set -o pipefail
+
+python3 /usr/share/securedrop-workstation-dom0-config/scripts/securedrop-admin.py "$@"
```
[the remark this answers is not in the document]
emkll: You are correct, indirection is not required here: the first prep-dom0 pass will install the packages (and drop the securedrop-admin script in place). I will update this

conorsch:
Looking good with the conditional logic. The latest changes show a minor test failure, related to the updated prod/test paths for keyfiles:

```
======================================================================
ERROR: test_rpm_repo_public_key (test_dom0_rpm_repo.SD_Dom0_Rpm_Repo_Tests)
  File "/home/user/securedrop-workstation/tests/test_dom0_rpm_repo.py", line 33, in test_rpm_repo_public_key
    with open(self.pubkey_actual, "r") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation'

======================================================================
FAIL: test_rpm_repo_config (test_dom0_rpm_repo.SD_Dom0_Rpm_Repo_Tests)
  File "/home/user/securedrop-workstation/tests/test_dom0_rpm_repo.py", line 54, in test_rpm_repo_config
    self.assertEqual(found_lines, wanted_lines)
AssertionError: Lists differ: ['[se[104 chars]ation-prod', 'file:///etc/pki/rpm-gpg/RPM-GPG-[152 chars]epo'] != ['[se[104 chars]ation', 'enabled=1', 'baseurl=https://yum-test[80 chars]epo']

First differing element 2:
'gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation-prod'
'gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation'

First list contains 1 additional elements.
First extra element 6:
'name=SecureDrop Workstation Qubes dom0 repo'

  ['[securedrop-workstation-dom0]',
   'gpgcheck=1',
-  'gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation-prod',
?                                                                    -----

+  'gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation',
-  'file:///etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation-test',
   'enabled=1',
   'baseurl=https://yum-test.securedrop.org/workstation/dom0/f25',
   'name=SecureDrop Workstation Qubes dom0 repo']

Ran 49 tests in 237.949s

FAILED (failures=1, errors=1)
Makefile:114: recipe for target 'test' failed
make: *** [test] Error 1
```

I believe all that's necessary is a few finishing touches on the salt logic. The conditionals for test/prod state look solid on visual review, it's mostly the corresponding filepaths that should be unique across the board, effectively overturning the simplification in https://github.com/freedomofpress/securedrop-workstation/pull/432/commits/c378758adfdb6836d15207260ce2cef9a5aea1b7 in favor of using explicit prod/test filepaths.

emkll:
@conorsch I have removed the logic for trusting the prod key in non prod environments. In the changes you've reviewed, I have also omitted changes to sys-firewall. After implementing changes to sys-firewall, it seemed to me like the logic was overly complicated for little benefit (templating the script in  `/rw/config/` for example, handling the clean-up case), and may introduce unnecessary risk. If we have a compelling reason to use it in the future, we could revisit this.

I think, before merging this, we should focus on the prod-first-install story. Can we get away with not using git to ship software to the dev vm / cloned to dom0?

emkll:
Rebased on latest `master` and added docs for the initial provisioning case. This should now be ready for final review

conorsch:
Latest changes look great, particularly the docs additions explaining the first-run steps for fetching, verifying, and installing the RPM. I'm completely satisfied with the changes here, @emkll, so over to you for a formal approval when you're ready!

PR #433, opened 2020-01-28
Change JSON key "last_updated" in `sdw-update-flag` to "last_status_update"

eloquence:
The use of `last_updated` here is ambiguous with with the usage in the separate file `sdw-last-udpated`, which contains the time of the last successful update of the SecureDrop Workstation.

The `last_updated` key in `sdw-last-updated` contains the timestamp of the last time the status itself was written to disk, regardless of success or failure.

I have also renamed the file itself to `sdw-update-status` for clarity, since that is the purpose of the file.

## Status

Ready for review

## Testing

- Tests should still pass
- Verify that the status file is written as expected to `sdw-update-status` when running the updater
- Verify that `make clean` removes file under old name

### DIFF
--- a/dom0/sd-clean-all.sls
+++ b/dom0/sd-clean-all.sls
@@ -21,7 +21,7 @@ remove-dom0-sdw-config-files:
       - /etc/qubes-rpc/policy/securedrop.Log
       - /etc/qubes-rpc/policy/securedrop.Proxy
       - /home/{{ gui_user }}/Desktop/securedrop-launcher.desktop
-      - /home/{{ gui_user }}/.securedrop_launcher/sdw-update-flag
+      - /home/{{ gui_user }}/.securedrop_launcher
 
 sd-cleanup-sys-firewall:
   cmd.run:
--- a/dom0/sd-clean-all.sls
+++ b/dom0/sd-clean-all.sls
@@ -21,6 +21,7 @@ remove-dom0-sdw-config-files:
       - /etc/qubes-rpc/policy/securedrop.Log
       - /etc/qubes-rpc/policy/securedrop.Proxy
       - /home/{{ gui_user }}/Desktop/securedrop-launcher.desktop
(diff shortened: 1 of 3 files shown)


eloquence:
Personally I think `sdw-update-status` would be a more self-explanatory name for the flag file, unless we're future-proofing it in some way by calling it "flag", and I'd be happy to rename it if that sounds reasonable (it's not used for anything yet), but for now focused on just addressing the dueling use of `last_updated` which I found very confusing, and which @emkll agreed would be reasonable to tweak.

eloquence:
Per brief chat at standup, renamed the file as well; leaving in draft for now until I give it one quick run-through in Qubes.

eloquence:
Tested in Qubes, works as expected.

dom0/sd-clean-all.sls
```diff
@@ -21,6 +21,7 @@ remove-dom0-sdw-config-files:
       - /etc/qubes-rpc/policy/securedrop.Log
       - /etc/qubes-rpc/policy/securedrop.Proxy
       - /home/{{ gui_user }}/Desktop/securedrop-launcher.desktop
+      - /home/{{ gui_user }}/.securedrop_launcher/sdw-update-flag
```
emkll: We should at least remove the other flags as well (the current, working flags) and consider deleting the entire folder (which include the logs)

eloquence: I was wondering about that; I thought you maybe intentionally omitted the folder. But I agree, it should clean up everything.

eloquence:
> The reboot_required case should be handled at this point and show the reboot button 

I'll file a separate issue for that, thanks for the clear STR

dom0/sd-clean-all.sls
eloquence: Done and tested. If you prefer to preserve the logs we can list individual files instead, but it would feel a bit wrong to me to leave this directory hanging around.

Issue #402, opened 2020-01-09
Skip VM updates if last check was performed <X hours ago


eloquence:
The launcher app currently checks for updates every single time, which can take a while. It would be good to skip this step if it has been performed very recently.

Open questions:
- Should the launcher GUI be shown to the user at all in this case, or be bypassed entirely?
- What is the maximum amount of time that may pass without checking for updates again?

eloquence:
>  Should the launcher GUI be shown to the user at all in this case, or be bypassed entirely?

I would vote to bypass the GUI entirely in this case, and to immediately launch the client instead.

eloquence:
Nominating as beta requirement as we don't want users to have to wait several minutes if they need to restart the client for any reason during a single session.

eloquence:
In addition to the "client restart" scenario, we currently run the "check for updates" on reboots as well, which means that the dom0 updates take especially long to go through.

zenmonkeykstop:
So reqs for this are:
- If updater has run within N hours, don't show GUI but go straight to client
- If system has been rebooted by updater, don't show GUI but go straight to client
- Otherwise, show GUI

Also adding:
- In either of the first two cases, log the no-GUI start.

emkll:
Yes that seems correct @zenmonkeykstop:

- read timestamp in `sdw_last_updated` flag from `~/.securedrop-launcher`
- if  `sdw_last_updated` < 8h and `last_required_reboot_performed()` [1]: start client
- else show launcher

[1] https://github.com/freedomofpress/securedrop-workstation/blob/master/launcher/sdw_updater_gui/Updater.py#L292

zenmonkeykstop:
OK, adding an update_skip_delta value (in sec) in config.json, and a "--update-delta" (positive) integer arg to the launcher with Salt setting it appropriately.

Behavior will be:
- if reload flag is set, clear reload flag and skip update check
- if reload flag not set, `--update_delta` not set or = 0, skip update check
- if reload flag not set, `--update_delta` is set and > 0,  and if `sdw_last_updated - now > update_delta`, do update check

PR #445, opened 2020-02-07
Add security notification script sdw-notify

### LINKED ISSUE #236: Display system alert recommending shutdown after 5 days of activity

eloquence:
This script checks whether updates have been applied recently enough using the preflight updater (AKA launcher) that is part of the SecureDrop Workstation. "Recently enough" is defined as follows:

- If the updater has never successfully run, uptime does not exceed a warning threshold.
- If the updater has run successfully, it is not longer ago than the  warning threshold. However, we allow for a grace period.

In this PR, the warning threshold is set to 5 days, and the grace period to 30 minutes.

The script runs as a cron job and displays a PyQt warning if warranted according to the above criteria. Lock files stored in `/run/user` are used to avoid showing the warning multiple times, or while the preflight updater is running.

As part of this work, code from `sdw-launcher` has been reorganized, to allow for shared use in a utility library by other system components like `sdw-notify`.

**NOTE:** Due to packaging changes, this PR bumps the version of `securedrop-workstation`. The packaging changes address parts of #449 (the creation of `*.pyc` and `*.pyo` files) by disabling implicit Python bytecode compilation.

## Status

Ready for initial review. Draft until re-verification in Qubes and RPM updates. 

## Description of Changes

Fixes #236

## Testing

Installation:
- [ ] Check out this PR in your development VM
- [ ] `make clone && make prep-dom0` in `dom0`

Regression testing:
- [ ] Test that the preflight updater is still working as expected by clicking the "SecureDrop" desktop icon
- [ ] Test that the preflight updater still creates logs in `~/.securedrop_launcher/launcher.log`

Testing new preflight updater functionality:
- [ ] Test that you cannot run another preflight updater while it is already running

Testing new notification functionality:

Before running any commands, I recommend running `watch tail -n 25  ~/.securedrop_launcher/logs/sdw-notify.log` in a separate terminal. All notable events below should produce log entries.

- [ ] Test that running `/opt/securedrop/launcher/sdw-notify.py` has the expected results for your current workstation state per the [behavior design doc](https://docs.google.com/document/d/1cCpcpLRiKC0tm_vsD4k4Q1Fr5_R34htCb8FNTV8mRCI/edit#)
- [ ] Test that setting the timestamp written to `~/.securedrop_launcher/sdw-last-updated` to a date well in the past and running  `/opt/securedrop/launcher/sdw-notify.py` causes the notification to appear, if the system has been up for more than 30 minutes (uptime grace period).
   - You can manually edit `UPTIME_GRACE_PERIOD` in  `/opt/securedrop/launcher/sdw_notify/Notify.py` to force a run.
- [ ] Test that setting the timestamp written to `~/.securedrop_launcher/sdw-last-updated` to today and  running  `/opt/securedrop/launcher/sdw-notify.py` causes the notification to _not_ appear.
   - Ensure that the uptime grace period has elapsed or is set to 0.
- [ ] Test that removing the timestamp `~/.securedrop_launcher/sdw-last-updated` causes the notification to appear. **(Updated per discussion below.)**
- [ ] Test that creating a timestamp file `~/.securedrop_launcher/sdw-last-updated` with a bad value causes the notification to appear. **(Updated per discussion below.)**
- [ ] Test that all of the above produces log entries.
- [ ] Test that the cron job runs as expected (easiest by modifying the `0` to `*` in `/etc/crontab` to change it from "every hour" to "every minute")

## Checklist
- [x] Linting (`make flake8`) passes
- [ ] Tests (`make test`) pass in dom0 of a Qubes install
   - I'm seeing unrelated test failures related to kernel versions and the RPM key in sys-firewall.

## Authorship

This PR represents a collaboration between @pierwill and @eloquence, with a lot of helpful input from @emkll.


launcher/tests/test_notify.py
```diff
@@ -0,0 +1,247 @@
...
+import subprocess
+
+from unittest import mock
+from importlib.machinery import SourceFileLoader
+from multiprocessing import Pool
+from tempfile import TemporaryDirectory
+
+relpath_notify = "../sdw_notify/Notify.py"
+path_to_notify = os.path.join(os.path.dirname(os.path.abspath(__file__)), relpath_notify)
+notify = SourceFileLoader("Notify", path_to_notify).load_module()
```
eloquence: This is the pattern for relative imports that was already used in `sdw-launcher`, updated here to avoid the deprecation warning caused by the use of `imp`. However, I'm sure a cleaner relative import is possible and would be happy to refactor both files - pointers appreciated!

launcher/tests/test_notify.py
```diff
@@ -0,0 +1,247 @@
...
+        updater.LOCK_FILE = os.path.join(tmpdir, "sdw-launcher.lock")
+        lh = updater.obtain_lock()  # noqa: F841
+
+        # We're in the same process, so obtaining an additional lock would
+        # always succeed. We use the multiprocessing module to run the function
+        # as a separate process.
+        p = Pool(processes=1)
+        lock_result = p.apply(notify.can_obtain_updater_lock)
+        p.close()
+        assert lock_result is False
```
eloquence: I've struggled to find a clean way to provoke a lock conflict in a test that works well across platforms. Pointers appreciated here as well. This method does appear to work, but as far as I can tell, it causes the code coverage for those lines to be misreported.

.circleci/config.yml
```diff
@@ -26,6 +26,7 @@ jobs:
             virtualenv .venv
             source .venv/bin/activate
             pip install --require-hashes -r test-requirements.txt
+            sudo apt install lsof
```
eloquence: `lsof` is available on Mac and under Linux, and therefore seemed like the best fit to perform a cross-platform test for exclusive lock access. It is, however, not installed in this particular CircleCI container by default.

dom0/sd-dom0-crontab.sls
```diff
@@ -0,0 +1,20 @@
...
+# SecureDrop preflight updater has not run for longer than a defined
+# warning threshold.
+dom0-crontab-update-notify:
+  file.blockreplace:
+    - name: /etc/crontab
+    - append_if_not_found: True
+    - marker_start: "### BEGIN securedrop-workstation ###"
+    - marker_end: "### END securedrop-workstation ###"
+    - content: |
+        0 * * * * {{gui_user}} DISPLAY=:0 /opt/securedrop/launcher/sdw-notify.py
```
eloquence: The `DISPLAY=:0` environment varialbe is necessary to display a graphical warning from a cron job without a display.

dom0/sd-dom0-crontab.sls
```diff
@@ -0,0 +1,20 @@
...
+
+# Identify the GUI user by group membership
+{% set gui_user = salt['cmd.shell']('groupmems -l -g qubes') %}
+
+# Add an hourly job, run as the GUI user, to display a warning if the
+# SecureDrop preflight updater has not run for longer than a defined
+# warning threshold.
+dom0-crontab-update-notify:
+  file.blockreplace:
+    - name: /etc/crontab
```
eloquence: `/cron.hourly` will run as root, so modifying `/etc/crontab` appears to be the cleanest way to run as the gui_user, which is necessary in order to run with the same permissions as the preflight updater.

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -65,6 +67,8 @@ install -m 644 Makefile %{buildroot}/usr/share/%{name}/Makefile
 install -m 755 scripts/* %{buildroot}/usr/share/%{name}/scripts/
 install -m 644 launcher/*.py %{buildroot}/opt/securedrop/launcher/
 install -m 644 launcher/sdw_updater_gui/*.py %{buildroot}/opt/securedrop/launcher/sdw_updater_gui/
+install -m 644 launcher/sdw_notify/*.py %{buildroot}/opt/securedrop/launcher/sdw_notify/
+install -m 644 launcher/sdw_util/*.py %{buildroot}/opt/securedrop/launcher/sdw_util/
```
eloquence: Note we're currently not marking `sdw-notify.py` or `sdw-launcher.py` as executable here; that's enforced in the dom0 Salt state.

launcher/sdw_util/Util.py
```diff
@@ -0,0 +1,35 @@
+"""
```
emkll: This is great, thanks for doing this. There uptime, lockfile, but I think this definitely sufficient for merge in this iteration. If we were to move further (non-logging) logic here, we should have test coverage here

launcher/sdw-notify.py
```diff
@@ -0,0 +1,62 @@
...
+
+    warning_should_be_shown = Notify.is_update_check_necessary()
+    if warning_should_be_shown is None:
+        # Data integrity issue with update timestamp. Logged.
+        sys.exit(1)
+    elif warning_should_be_shown is True:
+        show_update_warning()
+
+
+def show_update_warning():
```
emkll: Given how straightforward the QT logic is here, I think it's fine to leave as is, but if ever we introduce further complexity/stying here, I would recommend we decouple the main from the QT code, and simply instanciate the QT app object in the main, and handle all other QT tasks in a separate file/class

launcher/sdw_notify/Notify.py
```diff
@@ -0,0 +1,174 @@
...
+from datetime import datetime
+
+sdlog = logging.getLogger(__name__)
+
+# The directory where status files and logs are stored
+BASE_DIRECTORY = os.path.join(os.path.expanduser("~"), ".securedrop_launcher")
+
+# The file and format that contains the timestamp of the last successful update
+LAST_UPDATED_FILE = os.path.join(BASE_DIRECTORY, "sdw-last-updated")
+LAST_UPDATED_FORMAT = "%Y-%m-%d %H:%M:%S"
```
emkll: nit: using the variable from Updater.py or storing in Utils.py could help reduce duplication here

launcher/tests/test_notify.py
emkll: I think here it would be fine to mock `open` or `lockf` and have `IOException` (not tested, but as example, using the following decorators :

```@mock.patch("builtins.open", side_effect=IOError())```
and 
```@mock.patch("fcntl.lockf", side_effect=IOError())``` 

depending on which part of the lock logic you want to test

File: launcher/tests/test_updater.py [hunk 10]
```diff
@@ -12,7 +12,7 @@
 path_to_script = os.path.join(
     os.path.dirname(os.path.abspath(__file__)), relpath_updater_script
 )
-updater = imp.load_source("Updater", path_to_script)
+updater = SourceFileLoader("Updater", path_to_script).load_module()
```
emkll: thanks for squashing this deprecation warning :+1:

eloquence:
Thanks for the detailed review! A couple of quick responses:

> In other words, for 3 and 4, I think that the absence or the invalidity of a timestamp should trigger the notification/alert. 

That makes sense, let's just show the warning immediately in those two cases without a grace period, because they should not occur on a normal system, and we don't want to delay letting the user know that something is wrong.

I really like your suggestion of populating `sdw-last-updated` during the installation, that would avoid showing the warning if the user waits a little bit after the install, and would also ensure the updater doesn't have to run immediately again after the install once #402 is merged. I think that can be done as a separate and independent change.

> The only option presented to a user in this notification is to acknowledge via the OK button. In order to better guide the user into the next steps, it may be a good idea to allow the user to open the launcher/updater directly from this notification.

I completely agree; I deliberately kept the notification super simple for this first iteration, but I think we should tweak it a bit:
1) vary the text of the notification based on the actual system state
2) give the user an option to launch the preflight updater if the client is not already running

The additional logic needed for 2) makes me think this is a good change for a follow-up PR (plus I'm not versed in PyQt yet, so it may be faster for someone else to take this on).

eloquence:
> I really like your suggestion of populating `sdw-last-updated` during the installation, that would avoid showing the warning if the user waits a little bit after the install

Thinking this through a bit more, there are a couple of issues with that approach:
1) Right now we're not enforcing updates to `dom0` during the provisioning (it's just recommended in the README).
2) I think to be able to trust the state of `sdw-last-updated` it makes sense that it should only ever be written by a single source of truth (preflight updater), so we don't potentially have to deal with inconsistencies.
3) Even if we write this during the provisioning, the notification might pop up if the cronjob happens to run first.
4) It's probably a good idea to recommend to anyone installing the workstation that they run the preflight updater immediately after the installation, since it's such a crucial system component. So I'm not sure we _want_ to suppress it by marking the system as up-to-date during provisioning.

This is IMO beyond the scope of this PR, but what do y'all think about storing the installation date in  a file like `sdw-install-date` as a final step in a `make all` run / `securedrop-admin --apply`? Then the notifier could stay silent if that file doesn't exist (because it indicates that the installation is still in progress, or broken). Recording this date could also be useful for debugging. Unless there's an easy way to infer that the installation is complete?

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
eloquence: Done, but that required changing the wildcard pattern under `%files` (to avoid duplication warnings), which in turn exposed the issue described by @conorsch in #449. We'll need to decide whether to explicitly include the `*.pyc`/`*.pyo` files (previously they were _implicitly_ included by wildcard), explicitly _exclude_ them via `%exclude`, or disable the build root policy that generates them in the first place; see discussion in #449.

eloquence:
Re: mocking the lock (lock-mocking):

> I think here it would be fine to mock open or lockf and have IOException (not tested, but as example, using the following decorators :

That's a great point, I think it's reasonable to treat `fcntl` itself as reliable and focus on testing the business logic here. Done for e95e4c4d9adabac254ac2a0c819f482fe9dc9493 for the existing test.

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -22,17 +22,23 @@ SecureDrop Workstation project. The package should be installed
 in dom0, or AdminVM, context, in order to manage updates to the VM
 configuration over time.
 
+# To ensure forward-compatibility of RPMs regardless of updates to the system
+# Python, we disable the creation of bytecode at build time via the build
+# root policy.
+%undefine py_auto_byte_compile
```
eloquence: This is the change discussed in https://github.com/freedomofpress/securedrop-workstation/issues/449#issuecomment-584484190 .

launcher/tests/test_notify.py
```diff
@@ -0,0 +1,322 @@
...
+def test_notify_lock(
+        mocked_info, mocked_warning, mocked_error
+):
+    """
+    Test whether we can successfully obtain an exclusive lock for the notifier
+    script
+    """
+    with TemporaryDirectory() as tmpdir, \
+            mock.patch("Notify.LOCK_FILE_NOTIFIER",
+                       os.path.join(tmpdir, "sdw-notify.lock")):
```
eloquence: FYI: I'm using this pattern to patch the global variable, to avoid unintended side effects either across tests or in a single test; please let me know if there's a preferable pattern for the same result.

launcher/sdw_util/Util.py
eloquence: I'll take a stab at refactoring the lockfile logic into the util library to further reduce duplication and increase testability.

eloquence: This is now done.

eloquence:
Recap of recent changes:
- We're at 100% test coverage for Updater, Util and Notify
- RPM packaging now handles permissions for executables and disables BRP that implicitly generates *.pyc and *.pyo files
- Util now includes all lockfile handling
- We (should) now show a warning when there's no timestamp file
- We (should) now show a warning when the timestamp file is in the wrong format.

I've not fully re-tested yet, but you're welcome to take a spin through it again @emkll; I'll officially move it to "Ready" again once I've done another test through myself. Thanks for all the comments so far!

eloquence:
> Prior to merge, could you please run the black formatter on launcher/ subdirectory

Done. Will re-test now, as well, and await Conor's visual review.

eloquence:
(Test plan and spec updated per agreed upon revised behavior.)

eloquence:
(Rebased to latest upstream `master`.)

eloquence:
Changed text of warning slightly, to "This computer has not been checked for security updates recently." This is so it's accurate under all circumstances when the warning is shown.

(As discussed above, additional logic for launching the updater or showing more specific messages can be handled in follow-up changes.)

eloquence:
I've retested this in Qubes and it works as expected. @conorsch, now would be a good time for a visual review (I don't think you need to do a full functional test since both @emkll and I have done so extensively, but up to you of course).

Prior to merge, I will squash the changes.

launcher/tests/test_notify.py
conorsch: The path insertions aren't ideal. Since we have an RPM package now, we can install the Python modules at the system level and import them directly. Let's not do that now, especially with #450 still pending review. Not critical pre-pilot, but would make the updater/notifier code much more maintainable.

launcher/tests/test_notify.py
```diff
@@ -0,0 +1,171 @@
...
+
+# Regex for warning log if the last-updated timestamp does not exist (updater
+# has never run)
+NO_TIMESTAMP_REGEX = r"Timestamp file '.*' does not exist."
+
+# Regex for warning log if we've updated too long ago, and grace period has elapsed
+UPDATER_WARNING_REGEX = (
+    r"^Last successful update \(.* hours ago\) is above warning threshold "
+)
+r"\(.* hours\). Uptime grace period of .* hours has elapsed (uptime: .* hours)."
```
conorsch: Looks like L30 & L31 should be swapped. As written, `UPDATER_WARNING_REGEX` is resolving to the first half of the pattern. The problem of strings-not-concatenating occurs throughout `test_notify.py`, will comment in-line.

launcher/tests/test_notify.py
```diff
@@ -0,0 +1,171 @@
...
+
+# Regex for info log if we've updated too long ago, but grace period still ticking
+GRACE_PERIOD_REGEX = r"Last successful update \(.* hours ago\) is above "
+r"warning threshold \(.* hours\). Uptime grace period of .* hours has not elapsed "
+r"yet \(uptime: .* hours\)."
```
conorsch: String parts aren't concatenating, try:

```
# Regex for info log if we've updated too long ago, but grace period still ticking
GRACE_PERIOD_REGEX = (
    r"Last successful update \(.* hours ago\) is above "
    r"warning threshold \(.* hours\). Uptime grace period of .* hours has not elapsed "
    r"yet \(uptime: .* hours\)."
)
```

launcher/tests/test_notify.py
```diff
@@ -0,0 +1,171 @@
...
+
+# Regex for info log if we've updated recently enough
+NO_WARNING_REGEX = (
+    r"Last successful update \(.* hours ago\) is below the warning threshold "
+)
+r"\(.* hours\)."
```
conorsch: As above, strings are not concatenating; try:

```
# Regex for info log if we've updated recently enough
NO_WARNING_REGEX = (
    r"Last successful update \(.* hours ago\) is below the warning threshold "
    r"\(.* hours\)."
)
```

launcher/tests/test_notify.py
eloquence: Yikes, great flag. `black` helpfully inserted the parentheses, which should have made it obvious to me that the concatenation was wrong. 10003bd31707da6776820006cd3af55f787790c1 should fix these concatenations throughout.

Sometimes Python makes me miss semicolons. ;-)

launcher/tests/test_notify.py
eloquence: Yeah I agree. Note that this is just copying the approach that was already in place (slightly revised to avoid a deprecation warning for `imp`), but we should definitely clean the launcher organization up a bit.

Issue #424, opened 2020-01-23
Rebuild qubes-template-securedrop-workstation for production


emkll:
The test apt key is baked into our base image [1]. In order to prepare for the beta, we will need to rebuild the template using a production apt server.

This requires either:
- Deploy certain packages to the prod apt server prior to running the build
- Adding some build-only logic to use a pre-prod server to build the template

[1] https://github.com/freedomofpress/qubes-template-securedrop-workstation/blob/master/securedrop-workstation/04_install_qubes_post.sh#L39

emkll:
The rationale here is that we want to use a single template in all environments. As discussed in standup today, we can sign the resulting .rpm artifact with both test and prod keys: After a build, they will be signed with the test key and pushed to the test apt server. Once testing is successful, we can sign this artifact with the release key, and push to the prod apt server for release.

I would recommend we proceed as follows:
1.  (optional) Build new kernel, test locally
2.  Replace logic in https://github.com/freedomofpress/qubes-template-securedrop-workstation/blob/master/securedrop-workstation/04_install_qubes_post.sh#L42 to point to production apt server
3.  Build template

For 2., we can either (keeping in mind the goal would be to reduce (eliminate) the possibility of a prod install using non-prod repos/artifacts) :
a. Upload the kernel images (+ metapackage and config package) to the prod server, and use the prod server to install the kernel images
b. Use the apt-test server for bootstrap, and then remove the apt-test server *after* installing the kernel (replacing with prod apt server + keys). We will need to carefully promote the kernel images (+metapackage and config packages as well)

While method a. provides consistency at the expense of  complexity, I would be inclined to go for option b., for the beta. 
@conorsch @redshiftzero  any preferences here ?

conorsch:
Great discussion, @emkll. For now, let's proceed with installing "test" packages in the template build logic, and signing the resulting template RPM with the test key (since CI-built artifacts were used). That'll keep the wheels turning for the current sprint.

We should aim to provide _production_ packages in advance of beta QA in the following order:

1. Tag suitable SDW deb packages
2. Build prod-ready SDW deb packages, upload to _prod_ apt server
3. Update template build logic to use _prod_ config by default (although a separate target for "make staging-template" or similar would be useful!)
4. Sign _prod-only_ template RPM with prod key
5. Upload to prod endpoints

[...]

PR #450, opened 2020-02-11
Skips running the updater if last update was good, a reboot isn't needed, and a config-specified delta isn't defined.

### LINKED ISSUE #402: Skip VM updates if last check was performed <X hours ago

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes #402.
- updates the SecureDrop desktop shortcut executable to include a parameter `--skip-delta=<update_skip_delta>` 
- updates the Securedrop launcher  to add the optional `--skip-delta` parameter above. If the last update was successful and less than `update_skip_delta` seconds ago, then the launcher runs the SecureDrop client directly. Otherwise, it runs the updater. if `--skip-delta` is omitted, a default interval of 8 hours is assumed. If `--skip-delta` is 0, the updater always runs.

## Testing
On this branch:
- install the SecureDrop Workstation (remembering to set the `update_skip_delta` field in config.json)
- confirm that the SecureDrop desktop shortcut's command is now `/opt/securedrop/launcher/sdw-launcher.py --skip-delta=28800`
- delete the dom0 directory `~/.securedrop_launcher` if it exists

### testing initial and second run
- double-click the SecureDrop shortcut
  - [x] confirm that the updater is launched
  - [x] confirm that the logfile at `~/.securedrop_launcher/logs/launcher.log` contains the line:
  ```
  INFO: Update status not available, launching updater.
  ```
- complete the update check, applying updates and rebooting if necessary. If a reboot is not necessary, click **Continue** to launch the client, then close the client
- double-click the SecureDrop shortcut
  - [x] confirm that the client is launched directly without the updater starting
  - [x] confirm that the logfile above contains the line:
  ```
  INFO: Updates OK and interval not expired, launching client.
  ```
- close the client

### testing updates_required status
- update the dom0 JSON file `~/.securedrop_launcher/sdw-update-status`, changing the value of the `status` field to `"1"`.
- double-click the SecureDrop shortcut
  - [x] confirm that the updater is launched
  - [x] confirm that the logfile at `~/.securedrop_launcher/logs/launcher.log` contains the line:
  ```
  INFO: Update status is 1, launching updater.
  ```
- after updates are complete, click **Continue**, then close the client

### testing reboot_required status (without reboot having been performed)
- update the dom0 JSON file `~/.securedrop_launcher/sdw-update-status`, changing the value of the `status` field to `"2"`.
- double-click the SecureDrop shortcut
  - [x] confirm that the updater is launched
  - [x] confirm that the logfile at `~/.securedrop_launcher/logs/launcher.log` contains the line:
  ```
  INFO: Required reboot pending, launching updater
  ```
- after updates are complete, click **Continue**, then close the client

### testing updates_failed status
- update the dom0 JSON file `~/.securedrop_launcher/sdw-update-status`, changing the value of the `status` field to `"3"`.
- double-click the SecureDrop shortcut
  - [x] confirm that the updater is launched
  - [x] confirm that the logfile at `~/.securedrop_launcher/logs/launcher.log` contains the line:
  ```
  INFO: Update status is 3, launching updater.
  ```
- after updates are complete, click **Reboot**

### testing reboot_required status (with required reboot)
- update the dom0 JSON file `~/.securedrop_launcher/sdw-update-status`, changing the value of the `status` field to `"2"`.
- reboot Qubes.
- double-click the SecureDrop shortcut
  - [x] confirm that the client is launched
  - [x] confirm that the logfile at `~/.securedrop_launcher/logs/launcher.log` contains the line:
  ```
  INFO: Required reboot performed, updating status and launching client..
  ```
- [x] confirm that the `sdw-update-status` file has been updated with status `0` and a new timestamp
- close the client

### testing interval expiry
- open a dom0 Terminal
- run the command `/opt/securedrop/launcher/sdw-launcher.py`
  - [x] confirm that the client starts without the updater running
- run the command `/opt/securedrop/launcher/sdw-launcher.py --skip-delta=0`
  - [x] confirm that the updater runs directly
- once the update completes, click **Continue**, then close the client
- in the Terminal, run the command `/opt/securedrop/launcher/sdw-launcher.py --skip-delta=600`
  - [x] confirm that the client starts without the updater running
- wait a little over 10 seconds, then, in the Terminal, run the command `/opt/securedrop/launcher/sdw-launcher.py --skip-delta=10`
  - [x] confirm that the updater runs.

- [x] Linting (`make flake8`) and tests (`make test`) pass in dom0 of a Qubes install
**`make test` fails an unrelated qubes-rpc test due to system setup**


launcher/sdw-launcher.py
```diff
@@ -1,42 +1,33 @@
...
+from sdw_updater_gui.Updater import should_launch_updater
+from PyQt4 import QtGui
+from logging.handlers import TimedRotatingFileHandler
 import logging
 import os
 import sys
+import argparse
 
 DEFAULT_HOME = os.path.join(os.path.expanduser("~"), ".securedrop_launcher")
-logger = ""
```
eloquence: Just a note that this logging code is refactored into `sdw_util` in #445. No strong preference on sequencing though; I'll try to address remaining comments on #445 today so perhaps we can aim to land that one first.

conorsch:
Now that's what I call a test plan! Stepping through on review, will update as I go...

config.json.example
```diff
@@ -4,7 +4,8 @@
     "hostname": "avgfxawdn6c3coe3.onion",
     "key": "Il8Xas7uf6rjtc0LxYwhrx"
   },
-  "environment": "prod",
+ "environment": "prod",
+ "update_skip_delta": 28800,
```
emkll: What's the rationale to have this value user-facing? Perhaps we can _override_ in config.json, but I think it would be more prudent to have the default be specified in `Updater.py`. Suppose that we want to change this default to a shorter period in the future, this would require an admin to configure their `config.json` and not be applied automatically.

If we ultimately decide that this config is necessary and useful, we should also validate the config.json value in https://github.com/freedomofpress/securedrop-workstation/blob/master/scripts/validate_config.py

zenmonkeykstop: The rationale for doing so is to make it tweakable, per-instance and without a code push. The default is set in `sdw-launcher.py`, and is 0 sec, as the safest approach to updates is to run the update check by default. But I get your point that we might want to make it optional, and enforce a reasonable default in its absence. Happy to make the change if others agree. Will also add validation once we decide to keep it.

zenmonkeykstop:
@conorsch - heads-up that as per offline chat, I removed the config.json option and associated build and salt configs. The only test plan change is to the last section - running the launcher script without any arguments will now cause it to test for a default update interval of 8 hours.

launcher/sdw_updater_gui/Updater.py
```diff
@@ -425,6 +425,63 @@ def _safely_start_vm(vm):
...
+    if _valid_status(status):
+        if _interval_expired(interval, status):
+            sdlog.info("Update interval expired: launching updater.")
+            return True
+        else:
+            if _status_ok_or_rebooted(status):
+                sdlog.info("Update interval not expired, launching client.")
+                return False
+            else:
+                sdlog.info("Updates or reboot required, launching updater.")
```
conorsch: As shown in the test plan, conditions can lead to "Updates or reboot required, launching updater". It might be helpful (to admins & support) to log the actual status code, and possibly timestamp.

launcher/tests/test_updater.py
```diff
@@ -840,3 +840,79 @@ def test_last_required_reboot_performed_not_required(
...
+                        (UpdateStatus.UPDATES_OK, True, True),
+                        (UpdateStatus.UPDATES_REQUIRED, True, True),
+                        (UpdateStatus.REBOOT_REQUIRED, True, True),
+                        (UpdateStatus.UPDATES_FAILED, True, True),
+                        (UpdateStatus.UPDATES_OK, True, False),
+                        (UpdateStatus.UPDATES_REQUIRED, True, False),
+                        (UpdateStatus.REBOOT_REQUIRED, True, False),
+                        (UpdateStatus.UPDATES_FAILED, True, False)
+])
+# @mock.patch("Updater.last_required_reboot_performed", return_value=True)
```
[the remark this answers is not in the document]
zenmonkeykstop: Removed, mocked in body of test. done.

launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -19,7 +32,7 @@ def __init__(self, parent=None):
         self.setupUi(self)
         self.clientOpenButton.setEnabled(False)
         self.clientOpenButton.hide()
-        self.clientOpenButton.clicked.connect(self.launch_securedrop_client)
+        self.clientOpenButton.clicked.connect(launch_securedrop_client)
```
emkll: Since you've moved this as a function to L14, is the method in `UpdaterApp` class now dead code? https://github.com/freedomofpress/securedrop-workstation/blob/0cbe309ccecd0a00c140f5e58af8a588262b63b0/launcher/sdw_updater_gui/UpdaterApp.py#L176

zenmonkeykstop: It is! deleted.

launcher/sdw_updater_gui/Updater.py
zenmonkeykstop: logger is configured to prepend a timestamp automatically. Added status code in message.

zenmonkeykstop:
@conorsch the argument handling that caused the problem you spotted has been fixed. The test plan should now describe expected behavior.

eloquence:
@zenmonkeykstop 

From reading the code, it looks like the `sdw-update-status` and `sdw-last-updated` files won't get updated if we bypass the updater due to the new logic introduced here. Is that correct? If so, in the event of a reboot, that would cause to no longer reflect the state of the machine correctly, due to the skip.

zenmonkeykstop:
@eloquence correct. The update time doesn't change, because the the updater hasn't run, and the status doesn't change until the updater runs next. if the `reboot_required` status is set, the launcher checks if a reboot took place and acts accordingly.

eloquence:
> The update time doesn't change, because the the updater hasn't run, and the status doesn't change until the updater runs next. 

I don't think this is the behavior we want. A stale `sdw-last-updated` will cause security notifications introduced in #445 to continue to appear; a stale `sdw-update-status` could have other unintended side effects for any future uses we make of this file (i.e. it should not indicate REBOOT_REQUIRED when a reboot is no longer required).

I would propose the following:
- Write `sdw-last-updated` to disk when all updates are successfully performed, regardless of reboot status. We want to capture the time of the update, not of the reboot, for the purposes of security checks.

- Update `sdw-update-status` (including timestamp) to reflect changes to the reboot status anytime the launcher runs, irrespective of the business logic introduced here -- so we get the file to the correct status as soon as possible. My understanding of the date in `sdw-update-status` is that it is _the date the status has changed_, so that date _should_ be refreshed.

Does that make sense, @zenmonkeykstop, @emkll & others?

zenmonkeykstop:
Seems reasonable @eloquence.

zenmonkeykstop:
@eloquence your requested changes are in. Minor changes to the test plan as some log messages have changed but otherwise unaffected.

eloquence:
@zenmonkeykstop Thank you! :) To be clear, it was just a suggestion, and I defer to @emkll on whether this is the right approach. Thanks for updating the test plan as well!

zenmonkeykstop:
Yup, if he dissents we can back it out, but it makes sense to me.

conorsch:
@zenmonkeykstop #445 is merged, which has introduced conflicts here. Mind rebasing and pinging for re-review?

conorsch:
> @zenmonkeykstop #445 is merged, which has introduced conflicts here. Mind rebasing and pinging for re-review?

Also, please bump the `securedrop-workstation-dom0-config` version `0.1.3` -> `0.1.4`, updating the changelog. We'll get nightlies uploaded automatically for use with the `staging` env. As for prod packages, let's wait until both this PR (#450) and #424 are resolved. If RPM changes here are stable, then we can make a prod-sig release of the config RPM.

zenmonkeykstop:
rebase and a quick spin through the test plan in progress.

zenmonkeykstop:
Ready for re-review!

eloquence:
Since I'm fairly familiar with this code and the expected behavior, I'll take a spin through this as well (can't approve, but can give a functional sign-off from my end).

launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -249,7 +249,7 @@ def run(self):
         # write the flags to disk
         run_results = Updater.overall_update_status(results)
         Updater._write_updates_status_flag_to_disk(run_results)
-        if run_results == UpdateStatus.UPDATES_OK:
+        if run_results in {UpdateStatus.UPDATES_OK, UpdateStatus.REBOOT_REQUIRED}:
```
eloquence: I would recommend adding a code comment here explaining the behavior.

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -94,6 +94,9 @@ find /srv/salt -maxdepth 1 -type f -iname '*.top' \
     | xargs qubesctl top.enable > /dev/null
 
 %changelog
+* Fri Feb 14 2020 Kevin O Gorman <kog@freedom.press> - 0.1.4
```
eloquence: Per @conorsch the recommendation for future changelog entries is to use "SecureDrop Team <securedrop@freedom.press>" consistently

conorsch: Citation: https://github.com/freedomofpress/securedrop-debian-packaging/issues/108

launcher/sdw_updater_gui/UpdaterApp.py
zenmonkeykstop: :heavy_check_mark:

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
zenmonkeykstop: :heavy_check_mark:

launcher/tests/test_updater.py
```diff
@@ -840,3 +833,90 @@ def test_last_required_reboot_performed_not_required(
...
+        (UpdateStatus.REBOOT_REQUIRED, True, True),
+        (UpdateStatus.UPDATES_FAILED, True, True),
+        (UpdateStatus.UPDATES_OK, True, False),
+        (UpdateStatus.UPDATES_REQUIRED, True, False),
+        (UpdateStatus.REBOOT_REQUIRED, True, False),
+        (UpdateStatus.UPDATES_FAILED, True, False),
+    ],
+)
+@mock.patch("Updater._write_updates_status_flag_to_disk")
+def test_should_run_updater_status_interval_expired(
```
eloquence: This and `test_should_run_updater_status_interval_not_expired` are your most complex tests, and I think they would benefit from a few code comments/test descriptions to explain the test behavior. That said, after starting at them for 15 minutes or so, the behavior seems reasonable to me.

My personal intuition would be to put the `expected` value rightmost in each list, to make the lists a bit more readable (kind of like a truth table), but I don't know if there's a convention you're following here that goes another way.

eloquence:
Testing now. :) So we understand more clearly in future what was merged, could you update the PR description to reflect the current behavior and default interval (post `config.json` removal)?

launcher/sdw_updater_gui/Updater.py
```diff
@@ -430,6 +430,65 @@ def _safely_start_vm(vm):
...
+                        "Required reboot performed, updating status and launching client."
+                    )
+                    _write_updates_status_flag_to_disk(UpdateStatus.UPDATES_OK)
+                    return False
+                else:
+                    sdlog.info("Required reboot pending, launching updater")
+                    return True
+            else:
+                sdlog.info(
+                    "Update status is {}, launching updater.".format(
```
eloquence: This results in log lines like "Update status is 1, launching updater" which is of course true, but is only useful to anyone familiar with the implementation internals -- I think for states we know the updater can enter, we should have corresponding human-readable log lines.

launcher/sdw-launcher.py
```diff
@@ -3,12 +3,33 @@
...
 from sdw_util import Util
 from sdw_updater_gui import Updater
-
+from sdw_updater_gui.UpdaterApp import launch_securedrop_client
+from sdw_updater_gui.Updater import should_launch_updater
 import logging
 import sys
+import argparse
+
+DEFAULT_INTERVAL = 28800
```
eloquence: I always make a habit of annotating seconds values, i.e. `# 8 hours` in this case

launcher/sdw_updater_gui/Updater.py
zenmonkeykstop: :heavy_check_mark:

launcher/sdw-launcher.py
zenmonkeykstop: :heavy_check_mark:

launcher/tests/test_updater.py
zenmonkeykstop: :heavy_check_mark:

eloquence:
> testing reboot_required status (without reboot having been performed)
> after updates are complete, click Continue, then close the client

I did not get the option to launch the client in this scenario, even no updates were applied. Instead it prompted me to reboot. I think that is the expected behavior currently.

I'm wondering if we should even check for updates if the user runs the updater _before_ rebooting. It seems more reasonable to immediately advance to the "Reboot required" screen. I don't think that needs to be done in the same PR, however, as we might want to think through the UX for that case a bit more (e.g., maybe the screen should be a bit different).

zenmonkeykstop:
Ypu that is the desired behaviour, will update test plan. As for the second part, I agree that it's desirable, but also that it's probably out of scope.

launcher/sdw_updater_gui/Updater.py
```diff
@@ -430,6 +430,65 @@ def _safely_start_vm(vm):
         sdlog.error(str(e))
 
 
+def should_launch_updater(interval):
+    sdlog.info("Starting SecureDrop Launcher")
```
eloquence: This results in a duplicate log line (since the same message is logged from `main`), is that intentional)

eloquence:
Excellent test plan, and everything working as expected (I ticked the missing boxes); all my comments are minor except for that behavior issue re: reboot case, which I'll file a new issue for.

Given that we're also altering the `sdw-last-updated` behavior with https://github.com/freedomofpress/securedrop-workstation/pull/450/files#diff-507ed828f43696d2e69142a9ea70d217R252 I'll note that I have also performed the following test:

- [x] Run updater with `REBOOT_REQUIRED` set in `sdw-update-status`, and confirm that `sdw-last-updated` gets written to disk after the update check is completed.

Note that for the purposes of this test, no actual updates were available, and I modified `sdw-update-status` _before_ the test. Are there additional tests we should run here to ensure the behavior is what we expect it to be?

launcher/sdw_updater_gui/Updater.py
zenmonkeykstop: nope, merge cruft. will delete.

zenmonkeykstop:
Updated the unit tests to also check for the sdw-update-status file being written (as there's precisely one case (REBOOT_REQUIRED + reboot performed) where that happens, so there's that.

The code writing the sdw-last-updated flags is in UpdaterApp.py, which currently does not have a coverage target being all UIy and threaded, tho @emkll  has expressed an interest in getting test coverage there as well. If that gets some testing love in the future, that would be the place to test.

launcher/sdw-launcher.py
```diff
@@ -17,11 +38,24 @@ def main():
...
         sys.exit(1)
     sdlog.info("Starting SecureDrop Launcher")
-    app = QtGui.QApplication(sys.argv)
-    form = UpdaterApp()
-    form.show()
-    sys.exit(app.exec_())
+
+    args = parse_argv(argv)
+
+    try:
```
emkll: Untested, but perhaps something like `interval = int(getattr(args, 'skip_delta', DEFAULT_INTERVAL))` would help simplify the logic here

conorsch:
Latest diff is looking good. @emkll has given a detailed functional review. There are a few opportunities to prune codepaths, but these files are quite hot now given the approach of the pilot, so merging to keep moving.

PR #454, opened 2020-02-14
Detect process conflicts with sdw-notify

### LINKED ISSUE #453: Don't fire sdw-notify during provisioning
Follow-up to #445.

#445 adds an hourly cron job that shows a security notice if the system hasn't been updated in a while, or we have no/bad data about when it's been updated. There's no point in showing this warning during the somewhat lengthy provisioning process, and the cron job may fire while provisioning is still underway. Let's add logic to suppress it from firing if we detect that an installation is underway.

eloquence:
## Status

Ready for review

## Description of Changes

Resolves #453.

We check for a defined list of processes using `pgrep` to ensure that `sdw-notify` does not run when provisioning or similar system updates are already underway.

## Test plan

Setup:
1. Check out this branch and run `make prep-dom0`
2. Ensure that `~/.securedrop_launcher/sdw-last-updated` does not exist, to force the notification to always appear.

Tests:
- [ ] Run `/opt/securedrop/launcher/sdw-notify.py`. The notification _should_ show up.
- [ ] Run `make test` and run `/opt/securedrop/launcher/sdw-notify.py` while it is doing its thing. The notification should _not_ appear, and you should see a log entry in `~/.securedrop_launcher/logs/sdw-notify.log` explaining why.
- [ ]  (optional, for good measure) Run the Qubes GUI updater and kick off a graphical update process (`qubesctl` must be running in the process list). Run `sdw-notify` while that's underway. Result should be the same as above.

## Checklist
- [x] Tested in Qubes
- [x] `make flake8` passes
- [x] `make test` passes
- [x] no packaging changes
- [x] made with love

### DIFF
--- a/launcher/sdw-notify.py
+++ b/launcher/sdw-notify.py
@@ -22,6 +22,11 @@ def main():
     """
 
     Util.configure_logging(Notify.LOG_FILE)
+
+    if Util.is_conflicting_process_running(Notify.CONFLICTING_PROCESSES):
+        # Conflicting system process may be running in dom0. Logged.
+        sys.exit(1)
+
     if Util.can_obtain_lock(Updater.LOCK_FILE) is False:
         # Preflight updater is already running. Logged.
         sys.exit(1)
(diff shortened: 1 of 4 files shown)


eloquence:
We will likely want to use this utility function in `sdw-launcher.py` as well, but I think that's best handled separately, as the preflight updater is an interactive application, and we want to think carefully about the interaction flow in this case (e.g., we probably don't want to just skip updates and launch the client, but if we error out, we have to clearly explain why).

PR #468, opened 2020-02-24
Fix initial beta release docs and manifest, update version to 0.2.1

emkll:
## Status

Ready for review

## Description of Changes
- Clarifies docs to install securedrop-workstation-dom0-config in production context
- Add `dom0/*.conf` to manifest and rpmspec (this was not caught in the review of #447)
- add launcher directory to /srv/salt 

## Testing
- In a dev environment, `make clean` completes without error (and removes the launcher dir from /srv/salt)
- [ ] We should rely on salt provisioning no-ops rather than complicating split logic (dev vs staging/prod)
- Build RPM on this branch
- Follow production provisioning docs
- [ ] Install completes without error (`securedrop-admin --apply`)

- [x] All tests (`make test`) pass in `dom0` of a Qubes install


README.md
```diff
@@ -214,7 +222,7 @@ name=SecureDrop Workstation Qubes initial install bootstrap
 
 3. Download the RPM package
 ```
-[user@work ~]$ sudo dnf download securedrop-workstation-dom0-config
+[user@work ~]$ dnf download securedrop-workstation-dom0-config
 ```
 
 The RPM file will be downloaded to your current working directory.
```
redshiftzero: probably worth fixing this while you're updating the prod install instructions: I think the step below (step 4) also needs a modification: `[dom0]$ qvm-run --pass-io work '/home/user/securedrop-workstation-dom0-config-x.y.z-1.fc25.noarch.rpm' > securedrop-workstation.rpm` -> `[dom0]$ qvm-run --pass-io work 'cat /home/user/securedrop-workstation-dom0-config-x.y.z-1.fc25.noarch.rpm' > securedrop-workstation.rpm`

emkll: good catch, fixed

emkll:
Tested this again on a fresh Qubes install with a .rpm built on this branch, and works as expected, this PR is not ready for review. Given the time it takes to go through testing, I would suggest the reviewer to simply review the diff visually, but will defer to their judgement.
As part of the review, I would like a critical look on how the dev/non-dev cases of file copies in dom0 are handled, with regard to the launcher and other dom0 files (shipping them via rpm in place, and then re-applied via salt, no-op). Should we continue copying these files to two locations? Should we instead have more branching in the code to avoid these cases?

Issue #178, opened 2018-10-30
Force hard system shutdown instead of suspend/hibernate


eloquence:
For security reasons (e.g., full-disk encryption state, application of automatic system updates, etc.), we should ensure that the system is shut down instead of suspended/hibernated. At minimum, this should occur when closing the laptop lid; if we can disable (at least at the UI level) suspend/hibernate altogether, that would be preferable.

conorsch:
> this should occur when closing the laptop lid

That we can do via `/etc/systemd/logind.conf`, setting `HandleLidClose=poweroff` (then bounce `systemd-networkd`) in `dom0`. Tested it and it does indeed power off the machine when the lid is closed. 

We'll need something else for disabling suspend altogether. A third case: when the machine is left idle for many hours, probably worth a shutdown, as well.

eloquence:
> That we can do via `/etc/systemd/logind.conf`, setting `HandleLidClose=poweroff`

Nit: the setting is named `HandleLidSwitch`. I can confirm that changing it to `poweroff` and running `sudo systemctl restart systemd-logind` alters the behavior on my X1 Carbon as expected.

emkll:
Closed via https://github.com/freedomofpress/securedrop-workstation/pull/475 and https://github.com/freedomofpress/securedrop-workstation/pull/473

PR #473, opened 2020-02-28
Fully power off workstation after lid close

eloquence:
## Description

In production or staging environment, fully powers off the workstation instead of merely suspending to disk if the laptop lid is closed, in order to protect full-disk encryption key. 

## Status

Ready for review. Towards #178, but does not resolve as "suspend" is still an option in the logout menu.

## Test plan

Since we've not issued a new prod release yet, I suggest testing this PR by running the Salt state on its own.

### Testing production config
1. `make clone` this branch into dom0
2. Ensure you have a valid `config.json`. Run `scripts/configure-environment  --environment prod` to switch it to production. (Note: That script currently messes up JSON formatting, make sure you have a copy if you care about that.)
3. `make prep-salt` to deploy salt config in `dom0`
4. Open a new `dom0` terminal and follow the logind logs with `journalctl -f -u systemd-logind`
5. In the original terminal, deploy the salt state added in this PR with `sudo qubesctl --show-output --targets dom0 state.sls sd-dom0-systemd`
6. - [ ] Observe in the journal that `systemd-logind` was restarted
7. - [ ] Observe that `/etc/systemd/logind.conf` now contains the `HandleLidSwitch=poweroff` directive.
8. Detach any external displays.
9. - [ ] Close the laptop lid and observe that the system is ultimately powered off (this can take up to a minute or so).
10. Reboot (sorry!).

### Testing cleanup
1. Follow the logind logs again in a separate terminal.
2. In a `dom0` terminal, run `sudo qubesctl --show-output --targets dom0 state.sls sd-clean-all` to undo the changes from the previous run. (Note: This will remove other files in `dom0` and force you to re-run `make all` for a working environment.)
3. - [ ] Observe in the journal that `systemd-logind` is restarted.
4. - [ ] Observe that `/etc/systemd/logind.conf` no longer contains the line in question.
5. Close your laptop lid.
6. - [ ] Observe that the system is again suspended, if that was your previous configuration.

### Testing that change has no impact in dev env
1. Set your `config.json` to `dev` using the same method as before.
2. Re-deploy it with `make prep-salt`.
3. Re-run the salt state `sd-dom0-systemd` as before.
4. - [ ] Observe that  `/etc/systemd/logind.conf` has _not_ been modified and `systemd-logind` has _not_ been restarted.

## Checklist

- [x] Linter (`make flake8`) passes
- [ ] `make test` not re-run yet
- [x] Adds files to RPM contents implicitly through existing wildcard rules
- [ ] Does not bump RPM version

### DIFF
--- a/dom0/sd-clean-all.sls
+++ b/dom0/sd-clean-all.sls
@@ -29,9 +29,11 @@ remove-dom0-sdw-config-files:
       - /home/{{ gui_user }}/Desktop/securedrop-launcher.desktop
       - /home/{{ gui_user }}/.securedrop_launcher
 
-sd-cleanup-crontab:
+sd-cleanup-etc-changes:
   file.replace:
-    - name: /etc/crontab
+    - names:
+      - /etc/crontab
+      - /etc/systemd/logind.conf
     - pattern: '### BEGIN securedrop-workstation ###.*### END securedrop-workstation ###\s*'
     - flags:
       - MULTILINE
--- a/dom0/sd-clean-all.sls
+++ b/dom0/sd-clean-all.sls
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 # vim: set syntax=yaml ts=2 sw=2 sts=2 et :
 
+{% import_json "sd/config.json" as d %}
+
 set-fedora-as-default-dispvm:
   cmd.run:
     - name: qvm-check fedora-30-dvm && qubes-prefs default_dispvm fedora-30-dvm || qubes-prefs default_dispvm ''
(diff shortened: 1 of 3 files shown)


eloquence:
There is in fact an easy way to disable the buttons as well, using `xfconf-query` (see [here](http://linux.chrissweeney.co.uk/topic.php?t=53)); haven't verified if it's sticky yet. Moving this back to development for a bit. Might indeed be best to use the xfce4-settings for all of this, since its power manager offers additional options (e.g., changing the behavior of the power button).

eloquence:
Unfortunately there's no way to force a shutdown on lid close using xfce's power management settings, so we're back to the logind method. But we _can_ hide the options we don't want. I'll make that part of a separate PR to tweak XFCE settings, including the icon size. We can land this one if it looks good.

kushaldas:
Updated the clean-up test command step.

PR #479, opened 2020-03-03
Fixes #15 removes our logging plugin for sd-log

kushaldas:
## Status

Ready for review

## Description of Changes

This PR removes the sdlog.conf for rsyslog via rc.local file
for the sd-log vm.

Fixes https://github.com/freedomofpress/securedrop-log/issues/15

## Testing

- [ ] `make all`
- [ ] `sd-log` vm should not have any error at `/var/log/syslog` file.
- [ ] `sd-log` vm `/rw/config/rc.local` file has the following lines:

```
# Add sd-rsyslog.conf file for syslog
rm -f /etc/rsyslog.d/sdlog.conf
systemctl restart rsyslog
```

- [x] All tests (`make test`) pass in `dom0` of a Qubes install

- [x] This PR adds/removes files, and includes required updates to the packaging

### DIFF
--- a/dom0/sd-log-disable-plugin.sls
+++ b/dom0/sd-log-disable-plugin.sls
@@ -3,6 +3,9 @@
 
 ##
 # Disable securedrop-log rsyslog plugin in sd-log vm
+# Fixes https://github.com/freedomofpress/securedrop-log/issues/15
+# Due to a single Debian package for both logging and also sd-log vm
+# we need to do the following step.
 ##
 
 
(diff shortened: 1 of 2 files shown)


dom0/sd-log-disable-plugin.sls
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# vim: set syntax=yaml ts=2 sw=2 sts=2 et :
+
+##
+# Disable securedrop-log rsyslog plugin in sd-log vm
```
emkll: Could we also link to the issue in the comments here? It would be useful to note, for future maintainers, that this is the side effect of using the same package for both the client and server component of the log: we want to make sure that sd-log does not send logs.

dom0/sd-log-disable-plugin.sls
```diff
@@ -0,0 +1,20 @@
...
+
+
+sd-log-remove-rsyslog-qubes-plugin:
+  file.blockreplace:
+    - name: /rw/config/rc.local
+    - append_if_not_found: True
+    - marker_start: "### BEGIN securedrop-workstation ###"
+    - marker_end: "### END securedrop-workstation ###"
+    - content: |
+        # Add sd-rsyslog.conf file for syslog
```
emkll: we are removing here, no?

kushaldas: Yes, I will update the comments.

Issue #459, opened 2020-02-18
Investigate updater performance bottlenecks


eloquence:
During testing, we have found that preflight updates can take as long as an hour, with the `fedora-30` VM in particular often taken a long time to update (@zenmonkeykstop reported 26 minutes just for Fedora updates). That's prohibitively slow and negates much of the speed advantage of using Qubes instead of an air-gap.

We need to understand a bit better what's causing updates to be slow, and in particular, if there are specific bottleneck issues that cause individual updater runs to be dramatically slower than they otherwise would be. For a systematic investigation, we need to:

- run the updater repeatedly under comparable conditions (i.e., ideally reset an updated VM to a pre-update state after the updater runs);
- investigate logs, CPU utilization, memory usage, network access, and disk access during update runs;
- conduct small experiments (e.g., allocate higher memory to VMs that are being updated; change the update servers used for an update; increase concurrency).

This ticket is intended to track findings from an initial investigatory spike, which may then lead to follow-up tickets, or the re-prioritization of existing ones (e.g., #403).

eloquence:
@kushaldas noted that the Fedora updater sometimes picks a very slow mirror to download from, which can cause Fedora updates to be prohibitively slow.

@conorsch notes a known upstream issue where VMs are sometimes started with insufficient memory (https://github.com/QubesOS/qubes-issues/issues/4890), and recommends carefully observing the memory assigned to VMs that are being updated / testing whether increasing that allocation makes a difference.

conorsch:
We've got a few options for improving the updater story. Listed in approximate order of preference, heavily prioritizing near-term feasibility:

  1. Replace login background tasks with graphical updater (#415)
  2. Consider running `dpkg --configure -a` (or similar) as part of preflight updates (#428)
  3. Ensuring VMs have adequate resources to update quickly
  4. Parallelize updates so multiple VMs are checked concurrently (#403)
  5. Reducing number of TemplateVMs to update, by consolidating/standardizing

Speaking entirely from my own experience running the updater (as well as the `make all` dev workflow) repeatedly, the most significant factor is _did the qmemmem service fall over_. If it did, my updates are likely going to take a very long time.

In terms of diagnosing the problem, the most reliable method I've come up with to determine "has qmemman crashed" is scripted, see below.

<details>

<summary>
check-qmemman.sh
</summary>

```
#!/bin/bash
# Utility script to check whether Qubes memory balancing
# service has failed. Compares the timestamps of the last
[...]

PR #486, opened 2020-03-04
Optimize Fedora Template updates

emkll:
## Status

Ready for review

## Description of Changes

Towards #459 

- fedora-30 updates no longer trigger workstation reboot: fedora-based AppVMs will be rebooted after the upgrade completes.
- Assumes that fedora-30 updates are required, given the frequency of updates to fedora-30

## Testing

- [ ] This is a meaningful optimization for end-users and will reduce the total time to use the workstation, without impacting the patch level of VMs.
- `make clone` and `make prep-dom0`
- [ ] Run the updater (delete `~/.securedrop_launcher/sdw-update-status` if required)
- [ ] Rebooting of VMs occurs without issue (`dnf downgrade zlib` in `fedora-30` to trigger updates)
- [ ] Updates complete successfully
- [ ] delete `~/.securedrop_launcher/sdw-update-status` 
- [ ] Run updater again, whonix-based vms properly update (doesn't take a long time, as sys-whonix is functional)
- [ ] Test coverage of Updater.py is 100%

### DIFF
--- a/launcher/sdw_updater_gui/Updater.py
+++ b/launcher/sdw_updater_gui/Updater.py
@@ -433,7 +433,7 @@ def shutdown_and_start_vms():
     for vm in sys_vms_in_order:
         _safely_start_vm(vm)
 
-    vms_in_order = ["sd-proxy", "sd-whonix", "sd-app", "sd-gpg", "sd-log"]
+    vms_in_order = ["sys-whonix", "sd-proxy", "sd-whonix", "sd-app", "sd-gpg", "sd-log"]
     sdlog.info("Rebooting all vms for updates")
     for vm in vms_in_order:
         _safely_shutdown_vm(vm)
--- a/launcher/sdw_updater_gui/Updater.py
+++ b/launcher/sdw_updater_gui/Updater.py
@@ -108,27 +108,12 @@ def _check_updates_dom0():
 
 def _check_updates_fedora():
     """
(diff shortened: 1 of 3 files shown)


launcher/sdw_updater_gui/Updater.py
```diff
@@ -108,27 +108,12 @@ def _check_updates_dom0():
...
-        sdlog.error(str(e))
-        return UpdateStatus.UPDATES_REQUIRED
-    finally:
-        reboot_status = _safely_shutdown_vm(current_templates["fedora"])
-        if reboot_status == UpdateStatus.UPDATES_FAILED:
-            return reboot_status
-
-    sdlog.info("{} is up to date".format(current_templates["fedora"]))
-    return UpdateStatus.UPDATES_OK
+    return UpdateStatus.UPDATES_REQUIRED
```
redshiftzero: Rebooting the VMs instead of the entire workstation as you are doing below makes a lot of sense to me. I was initially surprised by removing this check, but I see that it improves the average case performance: if we almost always need to apply these updates, we can save the time of running the check by instead just trying to apply them. We only lose out for the rare (only rare for Fedora) best case where no updates are needed.

redshiftzero:
Holding off on merge for a bit since @conorsch mentioned he also wanted to take a look at this pre-merge

conorsch:
Nothing to add, diff looks grand, all review comments addressed. :+1:

PR #496, opened 2020-03-11
Securedrop-workstation 0.2.3

emkll:
## Status

Ready for review 
## Description

Bumps version strings, provides rpm changelog
## Test plan

- [ ] CI is passing
- [ ] Version strings are correctly updated
- [ ] Changelog is accurate


rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -104,6 +104,11 @@ find /srv/salt -maxdepth 1 -type f -iname '*.top' \
     | xargs qubesctl top.enable > /dev/null
 
 %changelog
+* Wed Mar 11 2020 SecureDrop Team <securedrop@freedom.press> - 0.2.3
+- Aggregate logs for both TemplateVMs and AppVMs
+- Add securedrop-admin --uninstall
+- Optimize Fedora Template updates
```
redshiftzero: ah, we should also mention moving `sd-proxy` to a `securedrop-workstation-buster` based template here I think

emkll: good catch, agreed. Should now be fixed

Issue #498, opened 2020-03-19
[0.2.3-rpm] libxenlight failed to create new domain `sd-log` 


eloquence:
In my prod install, I have now seen this error message (displayed as a sticky Qubes notification) twice towards the end of updates. `sd-log` does end up running just fine.

`/var/log/qubes/qubes.log` shows this output:

<details>
<summary>Click to expand</summary>

```
2020-03-18 18:00:23,561 Setting Qubes DB info for the VM
2020-03-18 18:00:23,563 Starting Qubes DB
2020-03-18 18:00:23,633 Activating the sys-whonix VM
2020-03-18 18:00:28,747 Starting sd-proxy
2020-03-18 18:00:28,754 Starting sd-whonix
2020-03-18 18:00:32,529 Setting Qubes DB info for the VM
2020-03-18 18:00:32,529 Starting Qubes DB
2020-03-18 18:00:32,595 Activating the sd-whonix VM
2020-03-18 18:00:46,144 Setting Qubes DB info for the VM
2020-03-18 18:00:46,145 Starting Qubes DB
2020-03-18 18:00:46,179 Starting sd-log
2020-03-18 18:00:46,216 Activating the sd-proxy VM
2020-03-18 18:00:49,778 Start failed: internal error: libxenlight failed to create new domain 'sd-log'
2020-03-18 18:00:54,378 Starting sd-log
2020-03-18 18:00:58,560 Setting Qubes DB info for the VM
2020-03-18 18:00:58,560 Starting Qubes DB
2020-03-18 18:00:58,640 Activating the sd-log VM
2020-03-18 18:01:18,881 Starting sd-app
2020-03-18 18:01:26,525 Setting Qubes DB info for the VM
2020-03-18 18:01:26,526 Starting Qubes DB
2020-03-18 18:01:26,687 Activating the sd-app VM
2020-03-18 18:01:55,294 Starting sd-gpg
2020-03-18 18:02:01,086 Setting Qubes DB info for the VM
2020-03-18 18:02:01,088 Starting Qubes DB
2020-03-18 18:02:01,378 Activating the sd-gpg VM
```

</details>

eloquence:
See #495 for another `sd-log` error during a prod install reported by Kev last week, which may or may not be related.

eloquence:
Again seeing this during an update run today.

eloquence:
[...]

PR #516, opened 2020-03-28
Improves updater reboot handling

conorsch:
## Status

Ready for review.

## Description of Changes

Refs #498 

Adjusts the reboot logic within the Workstation graphical updater, such that: 

* SDW VMs are shut down _before_ the sys-* VMs (in an effort to reduce problems caused by `qvm-kill`ing the sys-* VMs while other machines, such as the SDW VMs, are attached to them)
* Boots all halted VMs in the reverse order they were halted in 

## Testing

First, it's important to be able to reproduce the VM-failed-to-start issue described in #498. @eloquence has had good luck reproducing, and tested vigorously while this branch was under development.

### For dev env
- `make all` to apply these changes

### For prod env
You'll have to patch the relevant file in both `/opt/` and `/srv/salt/`, since the dom0 state change as part of the updater will undo your changes. 

### For all envs

In order to ensure that a reboot is required in the graphical updater, we'll intentionally downgrade a package:

```
qvm-run fedora-30 'sudo dnf downgrade -y zlib'
qvm-shutdown --wait fedora-30
/opt/securedrop/launcher/sdw-launcher.py --skip-delta 0
```

Let the upgrade step run to completion, and ensure no VMs fail to start. You'll likely want to do this whole process 2-5 times, to ensure it's really resolved. 


eloquence:
Before this was a PR, I followed this process 4 times in my `0.2.3-rpm` prod install with the patch applied, and was no longer able to reproduce #498. Without this patch I was pretty much getting it all the time for `sd-proxy` at the end of the update process.

launcher/sdw_updater_gui/Updater.py
```diff
@@ -406,24 +406,33 @@ def shutdown_and_start_vms():
     in case they are being used by another non-workstation VM.
     """
 
+    sdw_vms_in_order = [
+        "sys-whonix",
+        "sd-proxy",
+        "sd-whonix",
+        "sd-app",
```
emkll: `sd-app` is dependant on `sd-proxy`, but not set as vm, but using a qrexec service. Its relative positioning is identical to the previous version, and didn't seem to pose a problem then, so fine as-is

eloquence:
I ran 5 updates successfully without the error, but on update number 6 just now (after a reboot) again failed with `libxenlight` error on `sd-proxy`. :( It's still possible this mitigates the issue significantly as I was able to get it on _every_ update previously.

conorsch:
Roger that, @eloquence. We've marked this PR as "referencing" #498, since we don't have enough data to confirm it's resolved, and your findings support that. Looking forward to digging more deeply into #498, but comfortable shipping these changes as a modest improvement in the interim.

launcher/sdw_updater_gui/Updater.py
conorsch: Good point, and given @eloquence's results, definitely worth revisiting. Not going to append any further changes here, since we've tested extensively on this diff, but will highlight your suggestion in #498 for follow-up.

Issue #527, opened 2020-04-08
Install failure: "Recurse failed: none of the specified sources were found"


eloquence:
This was originally noticed [once](https://github.com/freedomofpress/securedrop-workstation/issues/499#issuecomment-602941427) in QA but resolved after `dom0` updates, and has now been noticed once in a production install. During `securedrop-admin --apply`, the following error may occur:

```

          ID: dom0-securedrop-launcher-directory
    Function: file.recurse
        Name: /opt/securedrop/launcher
      Result: False
     Comment: Recurse failed: none of the specified sources were found
     Started: 20:52:46.766870
    Duration: 2.371 ms
     Changes:   

```

eloquence:
`qubesctl` provides a `saltutil.sync_all` command, the [documentation](https://github.com/QubesOS/qubes-mgmt-salt/blob/master/README.rst) for which states:

> Sync all modules. If a problem exists, one may remove the salt cache directory (`rm -r /var/cache/salt`) and re-sync the modules

In a production install, removing the salt cache directory resolved the issue above. @emkll, could you document which commands you ran to resolve, i.e. did you just remove the cache directory, or also run the sync command?

eloquence:
Mickael clarified that syncing alone did not resolve, but cleaning cache & syncing did. To attempt to resolve the issue, we've agreed to add these steps (clear cache & sync) to `securedrop-admin --apply`. I'll take a stab at this so we can test it.

eloquence:
We've not observed this since #530 was merged, so closing for now.

Issue #505, opened 2020-03-23
Warnings on uninstall, package not removed


eloquence:
Seeing the following warnings after running `securedrop-admin --uninstall`, notably:

> securedrop-workstation-dom0-config-0.2.3-1.fc25.noarch was supposed to be removed but is not!

<details>
<summary>
Click to expand
</summary>

```
Purging Salt config...
Uninstalling Template
warning: file /var/lib/qubes/vm-templates/securedrop-workstation-buster/whitelisted-appmenus.list: remove failed: No such file or directory
warning: file /var/lib/qubes/vm-templates/securedrop-workstation-buster/vm-whitelisted-appmenus.list: remove failed: No such file or directory
warning: file /var/lib/qubes/vm-templates/securedrop-workstation-buster/root.img.part.03: remove failed: No such file or directory
warning: file /var/lib/qubes/vm-templates/securedrop-workstation-buster/root.img.part.02: remove failed: No such file or directory
warning: file /var/lib/qubes/vm-templates/securedrop-workstation-buster/root.img.part.01: remove failed: No such file or directory
warning: file /var/lib/qubes/vm-templates/securedrop-workstation-buster/root.img.part.00: remove failed: No such file or directory
warning: file /var/lib/qubes/vm-templates/securedrop-workstation-buster/netvm-whitelisted-appmenus.list: remove failed: No such file or directory
warning: file /var/lib/qubes/vm-templates/securedrop-workstation-buster/icon.png: remove failed: No such file or directory
warning: file /var/lib/qubes/vm-templates/securedrop-workstation-buster/clean-volatile.img.tar: remove failed: No such file or directory
warning: file /var/lib/qubes/vm-templates/securedrop-workstation-buster/apps.templates: remove failed: No such file or directory
warning: file /var/lib/qubes/vm-templates/securedrop-workstation-buster/apps.tempicons: remove failed: No such file or directory
warning: file /var/lib/qubes/vm-templates/securedrop-workstation-buster/apps: remove failed: No such file or directory
warning: file /var/lib/qubes/vm-templates/securedrop-workstation-buster: remove failed: No such file or directory
Uninstalling dom0 config package
warning: file /srv/salt/securedrop-login: remove failed: No such file or directory
warning: file /srv/salt/securedrop-launcher.desktop: remove failed: No such file or directory
warning: file /srv/salt/securedrop-handle-upgrade: remove failed: No such file or directory
[...]

PR #530, opened 2020-04-09
Clear Salt cache and synchronize Salt before installing/uninstalling

eloquence:
Towards #527

## Status

Ready for review

## Test plan

- [ ] With the changes from this PR locally applied, attempt a clean production or staging install using `securedrop-admin` and observe whether you're able to reproduce #527 
- [ ] With the changes from this PR locally applied, attempt an uninstall using `securedrop-admin`, and observe whether you see any new issues not already tracked in #505

## If you have made changes to the provisioning logic

- [ ] All tests (`make test`) pass in `dom0` of a Qubes install (not run yet)

- [x] No files added/removed

### DIFF
--- a/scripts/securedrop-admin.py
+++ b/scripts/securedrop-admin.py
@@ -79,6 +79,22 @@ def validate_config(path):
         raise SDAdminException("Error while validating configuration")
 
 
+def refresh_salt():
+    """
+    Cleans the Salt cache and synchronizes Salt to ensure we are applying states
+    from the currently installed version
+    """
+    try:
+        subprocess.check_call(["sudo", "rm", "-rf", "/var/cache/salt"])
+    except subprocess.CalledProcessError:
+        raise SDAdminException("Error while clearing Salt cache")
+
+    try:
+        subprocess.check_call(["sudo", "qubesctl", "saltutil.sync_all", "refresh=true"])
+    except subprocess.CalledProcessError:
+        raise SDAdminException("Error while synchronizing Salt")
+
+
 def perform_uninstall():
 
     try:
@@ -128,6 +144,7 @@ def main():
         print("Applying configuration...")
         validate_config(SCRIPTS_PATH)
         copy_config()
+        refresh_salt()
         provision_all()
     elif args.uninstall:
         print(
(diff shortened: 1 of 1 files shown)


eloquence:
While we're at it, it may be reasonable to add the package cache clean operation discussed in #505 to this PR, since both changes require fairly lengthy test runs to verify. If so, happy to investigate or add a commit with a preferred command.

scripts/securedrop-admin.py
```diff
@@ -79,6 +79,22 @@ def validate_config(path):
         raise SDAdminException("Error while validating configuration")
 
 
+def refresh_salt():
+    """
+    Cleans the Salt cache and synchronizes Salt to ensure we are applying states
+    from the currently installed version
+    """
+    try:
+        subprocess.check_call(["sudo", "rm", "-rf", "/var/cache/salt"])
```
eloquence, line 88: It's worth noting there's also `qubesctl saltutil.clear_cache`, which does leave some scaffolding in place, but is a bit less drastic. The removal is recommended by the `qubes-mgtm-salt` README [here](https://github.com/QubesOS/qubes-mgmt-salt/blob/master/README.rst).

conorsch:
> With the changes from this PR locally applied, attempt a clean production or staging install

As part of review, I'll transition a test machine from "prod" to "staging" env, running the `--uninstall` action first.

scripts/securedrop-admin.py
conorsch: Comparing the state after running each command:

* `rm -rf /var/cache/salt` removes _everything_, as you might expect
* `qubesctl saltutil.clear_cache` removes _only files_, but leaves the directories. that sounds acceptable to me, but given that the qubes docs are still technically recommending an rm, let's do that. We also have a recent report that these commands did indeed resolve an issue https://github.com/freedomofpress/securedrop-workstation/issues/527#issuecomment-611709172

conorsch:
The introduced of these additional calls doesn't show any problems for me. It's worth nothing that there's now an error message displayed:

```
[CRITICAL] Specific ext_pillar interface qvm_prefs is unavailable
```

but that message doesn't interfere with the install or uninstall actions. I suspect it's related to an ordering problem in how the upstream qubes modules (specifically `qvm_prefs`) are merged into the site salt config. Given the quest for stability in the install flow, I'd prefer to merge the changes proposed here, and we can circle back to redirect stderr to dev/null if desired.

PR #532, opened 2020-04-10
Safely shut down sys-usb; tweak logging

### LINKED ISSUE #531: Stop killing sys-usb in updater reboots
In the GUI updater, we use `qvm-kill` to force-stop the `sys-*` VMs, given that a graceful `qvm-shutdown` call will fail, since other VMs depend on `sys-net` and `sys-firewall`. That's not the case for `sys-usb`, though: we can gracefully reboot it without any need to kill, since it has no net clients.

Several times while testing the updater, the force reset of sys-usb has left sys-usb in a broken state, requiring a reboot of the host machine in order to resolve.

eloquence:
## Status

Ready for review

## Description of Changes

Resolves #531. Makes logging slightly more verbose to help diagnose issues with specific VMs.

`sys-whonix` is now treated as a "safe to shut down" system VM along with `sys-usb` (`sd-whonix` does not depend on it running).

### Preparatory steps
(All testing should be done with an existing Qubes install at least at `0.2.4-rpm` - dev, prod or staging should make no difference)

1. Apply the changes in this PR to the launcher versions in `/opt/securedrop/launcher` and `/srv/salt/launcher` (if only the `/opt` copy is overwritten, the updater itself will replace it on the next run).
2. Downgrade `zlib` (`dnf downgrade zlib`) in `fedora-30` to ensure a realistic update scenario. Power off `fedora-30`.

### Scenario: fedora-30 update

1. Run  `/opt/securedrop/launcher/sdw-launcher.py --skip-delta 0`. This forces an updater run.
2. - [ ] Observe that the update completes successfully and that the `zlib` package in `fedora-30` is now at the latest version
3. - [ ] Observe that no unexpected errors are logged in `~/.securedrop_launcher/logs/launcher.log`
4. - [ ] Observe that `sys-usb` is running as expected and that the `sd-devices` workflow (automatic attachment of a LUKS-encrypted USB block device or USB printer to `sd-devices`) still works.

You may still encounter #498 due to the continued use of `qvm-kill`.

## Checklist

- [ ] All tests (`make test`) pass in `dom0` of a Qubes install (N/A, updater has its own test suite which passes)
- [x] No package updates required

### DIFF
--- a/launcher/sdw_updater_gui/Updater.py
+++ b/launcher/sdw_updater_gui/Updater.py
@@ -402,14 +402,14 @@ def shutdown_and_start_vms():
     correct order of operations, as sd-whonix cannot shutdown if sd-proxy is powered
     on, for example.
 
-    System AppVMs(sys-net, sys-firewall and sys-usb) will need to be killed and restarted
-    in case they are being used by another non-workstation VM.
+    All system AppVMs (sys-net, sys-firewall and sys-usb) need to be restarted.
+    We use qvm-kill for sys-firewall and sys-net, because a shutdown may fail
+    if they are currently in use as NetVMs by any of the user's other VMs.
     """
 
     sdw_vms_in_order = [
         "sd-app",
         "sd-proxy",
-        "sys-whonix",
         "sd-whonix",
         "sd-gpg",
         "sd-log",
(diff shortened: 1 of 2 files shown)


conorsch:
Have not reviewed yet, but did take the liberty of rebasing on latest master (8bfd1cfd9e8f4c4e8a82516484b471040b2cb9a3). I've had the sys-usb problem bite me on updater runs recently, so looking forward to reviewing, although likely not till next week.

eloquence:
Tested in Qubes including with `sys-whonix` change in 4e54c89, did not encounter our friend #498 but I expect it will only be completely gone once we get rid of `qvm-kill` (and maybe even poll for successful VM startup given the HVM boot time).

Issue #529, opened 2020-04-09
Provide method for sharing logs


eloquence:
Since all of our support during the SecureDrop Workstation pilot is remote, it's important that we have an agreed upon methods for sharing logs. Methods that will currently work:

- Copying to `dom0` using verbose `--pass-io` commands, then copying from there
- Use of a USB drive in `sd-log` VM

Methods that currently don't work:

- Copying directly from `sd-log` (restricted by us)
- Copy/paste (restricted by us)

Let's discuss if loosening any of those restrictions is appropriate, if any other methods make sense, or if not, which of the working methods we want to recommend for now.

redshiftzero:
I think we’d want to add a command in dom0 that would tar up and encrypt the logs to FPF's communications key (analogous to the existing `securedrop-admin logs` in core), and then we’d want to export to a network connected VM for onionsharing/emailing/other networked method (which would require a Filecopy permissions update). 

The change in #533 to add `sd-log @tag:sd-receive-logs ask` seems like it is a step along the path to the above, where in the interim we can instruct admins to apply the tag to an existing networked VM.

zenmonkeykstop:
Other stuff to consider based on support experiences: 
- we may want to specify a subset of VMs to collect logs from
- we may want to restrict logs submitted by keyword/regexp, or truncate logs sent
- we could consider setting up onionshare as a receiver on our side, to allow users to send logs unattended.

eloquence:
This was partially addressed via #533, which enables users to copy files from `sd-log` to another VM.

zenmonkeykstop:
This would be a requirement for broader adoption- adding a `support` tag to flag this and similar issues.

PR #533, opened 2020-04-10
Permit whitelisting VMs for copy/paste & copying logs via tags

### LINKED ISSUE #526: Consider supporting limited copy/paste
Currently we [prevent](https://github.com/freedomofpress/securedrop-workstation/blob/master/dom0/sd-dom0-qvm-rpc.sls#L13-L21) copy/paste in all VMs that are part of the project. As we enter real-world usage, we can anticipate an increasing need for copy/paste for a number of situations, e.g.:

- pasting passphrases into the login window from a `vault` VM
- copying messages from and to the conversation view (e.g., to another secure messaging app)
- copying logs and error messages.

Through RPC policies, we can control this permission with directionality for each VM. We should carefully consider changes to clipboard policy and behavior in the context of our overall threat model. 

In particular:
- Is it reasonable to loosen clipboard policies without any additional mitigations for certain VM<sub>1</sub>->VM<sub>2</sub> operations?
- Are there mitigations that we can put in place within SecureDrop Workstation VMs that would permit us to further responsibly loosen policies (e.g., can we detect clipboard events in `sd-app` and purge the clipboard after a paste)?
- Are there mitigations that we should coordinate with upstream (e.g., more fine-grained RPC policies to control purge behavior, a meaningful "ask" policy, clipboard previews, other UX improvements), and which would permit us to further loosen policies?

eloquence:
Adds RPC policies that let tagged VMs send clipboard pastes to `sd-app`, receive clipboard pastes from `sd-app`, and receive files from `sd-log`.

## Status

Ready for review; docs in progress at https://github.com/freedomofpress/securedrop-workstation-docs/pull/35 and will need to be kept in sync with changes here.

Adds RPC policies for the following tags:
- `sd-receive-app-clipboard` -- VMs with this tag will be able to receive clipboard contents from `sd-app`
- `sd-send-app-clipboard` -- VMs with this tag will be able to send clipboard contents to `sd-app`
- `sd-receive-logs` -- VMs with this tags will be able to receive any file sent via `qvm-copy` from `sd-log`

Adds a script in `scripts/remove-tags` that removes all tags, and runs it for `make clean` / `securedrop-admin --uninstall` operations.

## Description of Changes

Fixes #526
Towards #529

#### Prerequisites

1. Check out the version of `sd-dom0-qvm-rpc.sls` from this branch and apply it to your `/srv/salt` directory in an up-to-date production or staging install, or provision a dev environment using `make dev` from this branch.
2. For production or staging environments, run `securedrop-admin --apply` to apply the change.

#### Test
1. Add the ability to receive clipboard contents from `sd-app` to a *target VM*, e.g. `qvm-tags work add sd-receive-app-clipboard`  (`qvm-tags <target VM name> add sd-receive-app-clipboard`)
2. Add the ability to send clipboard contents from a single *source VM* to `sd-app`, e.g. `qvm-tags vault add sd-send-app-clipboard` (`qvm-tags <source VM name> add sd-send-app-clipboard`).
3. Add the ability to receive logs to a *target VM*, e.g., `qvm-tags work add sd-receive-logs`
4. Launch the client
5. - [ ] Observe that you are able to copy clipboard contents from the SecureDrop Client reply box to the *target VM* specified in step 1
6. - [ ] Observe that you are able to copy clipboard contents from the *source VM* specified in step 2 to an input field in the SecureDrop Client
7. - [ ] Observe that you are able to copy files from `sd-log` to the *target VM* specified in step 3 using `qvm-copy-to-vm` and `qvm-copy`.
8. - [ ] Observe that you are not able to perform copy/paste from any SecureDrop VM to any other VM, except from `sd-app` to *target VM* you configured in step 1
9. - [ ] Observe that you are not able to perform copy/paste to any SecureDrop VM from any VM except from the *source VM* you configured in step 4 to `sd-app`.
10. - [ ] Observe that you are not able to perform qvm-copy *to* any SecureDrop VMs except from `sd-proxy` to `sd-app`.
11. - [ ] Observe that you are not able to perform qvm-copy *from* any SecureDrop VMs except for `sd-log` as permitted by the `sd-log` rules you configured.
12. Remove the tags by running `scripts/remove-tags` from this branch.
13. - [ ] Observe via `qvm-tags <name of VM> ls` that tags were correctly removed.
14. - [ ] Observe that the copy & paste rules you configured have been disabled.
15. - [ ] Observe that the `sd-log` rules you configured have been disabled.
16. - [ ] Spot check that the rules otherwise still behave as before.

#### Prerequisites
1. You must have a working staging or prod install you're willing to lose :cry: 
2. Re-tag select VMs as per the test plan above.
3. Make sure that the changes in this PR are applied to the `/bin` version of `securedrop-admin`, and to the files in `/srv/salt` touched by this PR (create `remove-tags`, update `sd-clean-all.sls`)
#### Test
1. Uninstall via `securedrop-admin --uninstall`.
2. - [ ] Observe that the tags you added are correctly removed as part of the uninstall.

#### Prerequisites

1. You must have a working dev install you are willing to lose :cry: 
2. Re-tag select VMs as per the test plan above.
3. Ensure that your `securedrop-workstation` checkout directory in `dom0` is using this branch.

#### Test
1. Uninstall via `make clean`
2. - [ ] Observe that the tags are correctly removed as part of the uninstall.

## Checklist

- [x] RPM updated to include new files (release not bumped yet)
- [ ] `make test` runs fine in `dom0` (not tested yet; test for RPC policies has been updated, however).


eloquence:
The use of `qvm-copy-to-vm` as indicated in the test plan works for me, but is deprecated; I was not able to use `qvm-copy` to copy files from `sd-log`, even after setting the policy to "allow" instead of "ask" -- I always get "Request refused". Not sure if the RPC policies require a refresh or if there's another reason those requests are still blocked.

eloquence:
> `I was not able to use qvm-copy to copy files from sd-log`

This appears to be due to the somewhat confusing behavior with empty selection prompts described in https://github.com/QubesOS/qubes-issues/issues/4403 ; the `@default` rule in 846105ee6e6bb0045d6edd1224f2042adeeb65c7 should resolve (and does not expand the scope of permissions).

dom0/sd-dom0-qvm-rpc.sls
```diff
@@ -17,6 +17,8 @@ dom0-rpc-qubes.ClipboardPaste:
     - marker_start: "### BEGIN securedrop-workstation ###"
     - marker_end: "### END securedrop-workstation ###"
     - content: |
+        @tag:send-clipboard-to-sd @tag:sd-workstation ask
+        @tag:sd-workstation @tag:receive-clipboard-from-sd ask
```
conorsch, line 21: Pleased to report that the directionality appears to work as intended: it's possible to allow clipboard pastes _into_ e.g. `sd-app`, while still forbidding pastes out. Testing is critical here, will likely take the bulk of review time. Fleshing out the test plan with specific targets, even more verbose than already written, is likely to help coordinate testing.

dom0/sd-dom0-qvm-rpc.sls
```diff
@@ -35,6 +37,8 @@ dom0-rpc-qubes.Filecopy:
     - marker_start: "### BEGIN securedrop-workstation ###"
     - marker_end: "### END securedrop-workstation ###"
     - content: |
+        sd-log @default ask
+        sd-log @tag:receive-sd-logs ask
```
conorsch, line 41: Let's make sure all the tags added by the SDW config begin with `sd-`, for consistency. So: `receive-sd-logs` -> `sd-receive-logs`, etc. Will make it easier over time to reason about collections and interactions between RPC policies.

tests/vars/qubes-rpc.yml
```diff
@@ -1,6 +1,8 @@
 - policy: qubes.ClipboardPaste
   starts_with: |-
     ### BEGIN securedrop-workstation ###
+    @tag:send-clipboard-to-sd @tag:sd-workstation ask
+    @tag:sd-workstation @tag:receive-clipboard-from-sd ask
```
conorsch, line 5: Good to have the tests updated. We also need a clean action, though: uninstalling should handle removal of these custom SDW-related tags. For example:

```
qvm-ls --tags sd-receive-logs --raw-list | xargs --no-run-if-empty -I {} qvm-tags {} del sd-receive-logs
```

for each custom tag.

dom0/sd-dom0-qvm-rpc.sls
conorsch: The requirement to include `@default ask` here is far from intuitive, but the citation you provided was informative, and functional testing locally does indeed line up with what's intended.

eloquence:
Thanks @conorsch! Let's discuss a bit more if/how we would ship those policies, which I think will influence both the naming & uninstall policy:

- Do we want the user to use existing VMs like `vault`?
- Do we want to provide instructions (or a dom0 utility) for setting up new VMs like `sd-signal` or `sd-vault` using our hardened base template?
- Do we want to wait with making RPC changes until we've agreed what additional VMs may be needed?

I think it's worth considering putting together a small helper script for managing the creation of custom VMs, e.g.:

```
$ sd-vm create sd-signal --enable=receive-logs,custom-software,network
Created TemplateVM "sd-signal-buster-template". Install custom software here.
Created AppVM "sd-signal" with network access using "sys-firewall".
Permission "receive logs" has been granted for VM "sd-signal".

$ sd-vm update sd-signal --disable=receive-clipboard
Permission "receive clipboard" has been revoked for VM "sd-signal".

$ sd-vm create shiny-passwords 
Error: Cannot create VM "shiny-passwords".
VMs added to SecureDrop Workstation must begin with "sd-" prefix.
```

This would give us a supported customization path with some guardrails, abstracting tags and templates away from the user, and allowing us to add business logic as needed. In this model, I agree that the uninstall scenario should also remove the tags, as part of the managed configuration of the system.

(The goal would be to keep this tool as thin of an abstraction as possible, of course.)

On the other hand, if we ask users to create/tag VMs using standard Qubes tooling, I'm not sure our uninstall script should remove those tags.

Either way, I think some customization path is necessary so that we can use the pilot to determine which additional VMs we may want to ship by default.

eloquence:
Talked to @emkll about this today. He's generally in favor of this approach during the pilot, until we have more data about what default VMs (and potential default policies for `vault` or others) we may want to add.

Main changes required:
- Thorough documentation
- `sd-` prefix for all tags
- Cleanup logic

Once all that is done, I'll promote this PR from draft.

For the cleanup logic, I'll just add a note to the docs and maybe the `--uninstall` command that custom tags will be removed.

dom0/sd-dom0-qvm-rpc.sls
eloquence: (This is done.)

eloquence:
I've tested the new tag removal logic via `make clean` and `securedrop-admin --uninstall`, and it seems to work as intended; quick visual feedback on the [remove-tags](https://github.com/freedomofpress/securedrop-workstation/pull/533/files#diff-3dc9c9971efd501d45f60f4d9da84e77) script is appreciated.

I'll now add a parallel docs PR, flesh out the test plan, and then mark it as formally ready for review.

eloquence:
I've significantly updated the test plan. I still want to add a section to the docs for managing the clipboard policies, but this should IMO be done in parallel with a first round review, just in case we make significant changes to this approach. The initial docs [here](https://github.com/freedomofpress/securedrop-workstation-docs/pull/35) should suffice as a starting point during review.

tests/vars/qubes-rpc.yml
eloquence: Ended up adding a Python script leveraging `qubesadmin` for this purpose.

dom0/sd-dom0-qvm-rpc.sls
eloquence: Test plan is plenty verbose now :)

scripts/remove-tags
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python3
```
emkll, line 1: Is there a reason we aren't using salt? It would minimize the amount of various scripts that need to be run (including as part of `sd-clean-all.sls`). We use salt for setting some workstation tags, for example: https://github.com/freedomofpress/securedrop-workstation/blob/5482f2e9c0e0658e0d44123e938bba49d40a0c04/dom0/sd-gpg.sls#L26-L28

Makefile
```diff
@@ -110,6 +110,7 @@ clean: assert-dom0 prep-salt ## Destroys all SD VMs
 	$(MAKE) destroy-all
 	sudo qubesctl --show-output --skip-dom0 --targets whonix-gw-15 state.sls sd-clean-whonix
 	sudo qubesctl --show-output state.sls sd-clean-all
+	./scripts/remove-tags
```
emkll, line 113: do you think it would make sense to invoke this script/cleanup via `sd-clean-all.sls`? The challenge is that the path in dev environment will be different than the one in staging and production environments

dom0/sd-dom0-qvm-rpc.sls
```diff
@@ -17,6 +17,8 @@ dom0-rpc-qubes.ClipboardPaste:
     - marker_start: "### BEGIN securedrop-workstation ###"
     - marker_end: "### END securedrop-workstation ###"
     - content: |
+        @tag:sd-send-clipboard @tag:sd-workstation ask
```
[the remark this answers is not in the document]
eloquence: I was thinking about this as well a bit over the weekend. One use case is copying text _out_ of disposable VMs used for viewing submissions. Right now this just works (with these changes) because they have the `sd-workstation` tag. If we want to narrow the policies and still support that use case, we could perhaps use `sd-client` (two-directional copy/paste in and out) and `sd-viewer-vm` (one-directional copy/paste out).

Makefile
eloquence: I agree that seems like the better way to do it, will poke.

dom0/sd-dom0-qvm-rpc.sls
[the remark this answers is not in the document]
redshiftzero: Echoing @emkll, I recommend we scope to `@tag:sd-client` (i.e. exclude disposable VMs), see this [note](https://www.qubes-os.org/doc/copy-paste/#on-copypaste-security) in the qubes documentation discouraging copy/paste from less trusted to more trusted domains: 

> one should keep in mind that performing a copy and paste operation from less trusted to more trusted domain can always be potentially insecure, because the data that we insert might potentially try to exploit some hypothetical bug in the destination VM (e.g. the seemingly innocent link that we copy from untrusted domain, might turn out to be, in fact, a large buffer of junk that, when pasted into the destination VM’s word processor could exploit a hypothetical bug in the undo buffer). This is a general problem and applies to any data transfer between less trusted to more trusted domains.... So, you should always copy clipboard and data only from more trusted to less trusted domains.

However, if we are finding (or otherwise think that) data transfer from disposable VMs into trusted VMs is an important workflow to support, we can file a followup issue for further investigation to either determine how to decrease risk when copy/pasting data from `sd-viewer` to `sd-client` or to provide an alternate method to handle that use case (some of the workflows we need to support for removing identifying metadata could handle this use case, e.g. adding an OCR step after we convert documents to trusted PDFs).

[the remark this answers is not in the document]
eloquence: Sounds good; for now I'll call the permission `sd-send-client-clipboard` and `sd-receive-client-clipboard` to reflect its narrow scope.

Makefile
conorsch: Definitely fold into `clean-salt`, as @emkll suggests. If we use a separate script, then the RPM spec will need to be updated, as well. Consider using the less elegant, but more simply `xargs`-based approach outlined above.

eloquence: The RPM spec already picks up the script by wildcard.

dom0/sd-dom0-qvm-rpc.sls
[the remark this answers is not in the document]
eloquence: This is done in 956cc10c1db1e6720c5ebd69d4ea600cc4706ad2 (I used `sd-app` rather than a tag since that seems the best way to ensure we have at most a many-to-one/one-to-many relationship here).

Makefile
eloquence: What I would propose is this:
1. Move `remove-tags` from `scripts` to `dom0`
2. Add it to RPM (I think that's what you meant, Conor?)
3. Call it from `sd-clean-all.sls` as we do for `update-xfce-settings` resets.
4. Remove it from `securedrop-admin` and `Makefile`.

Does that make sense?

conorsch: @eloquence Yes, those steps make sense. Moving the invocation to the salt clean-all state is sound at this point. 

Ideally we'd call it from the scripts dir in /usr/share/, but before we can easily do that, we'll need to use the RPM in all environments, including make-dev, as described in https://github.com/freedomofpress/securedrop-workstation/issues/505#issuecomment-602814288

eloquence: Done in 49916a21b7c163904d8c770092ea483a348f62ba, not tested yet.

eloquence: Tested now, seems to work as intended (this is via `securedrop-admin`):

```
          ID: remove-rpc-policy-tags
    Function: cmd.script
        Name: salt://remove-tags
      Result: True
     Comment: Command 'salt://remove-tags' run
     Started: 17:49:47.430389
    Duration: 267.274 ms
     Changes:   
              ----------
              pid:
                  29831
              retcode:
                  0
              stderr:
              stdout:
                  Removing tag 'sd-send-app-clipboard' from VM 'work'.
                  Removing tag 'sd-receive-app-clipboard' from VM 'work'.
----------
```

tests/vars/qubes-rpc.yml
```diff
@@ -1,6 +1,8 @@
 - policy: qubes.ClipboardPaste
   starts_with: |-
     ### BEGIN securedrop-workstation ###
+    @tag:sd-send-app-clipboard sd-app ask
```
emkll, line 4: I also think using `sd-app` here makes sense: the intent of the `sd-client` tag is to support dev environment for opening in dispVM and using split GPG (see https://github.com/freedomofpress/securedrop-workstation/commit/784076da56d25f2c089864ae9cff2aafb0a756a0)

dom0/sd-clean-all.sls
```diff
@@ -46,13 +46,20 @@ remove-dom0-sdw-config-files:
...
+
+# Remove any custom RPC policy tags added to non-SecureDrop VMs by the user
+remove-rpc-policy-tags:
+  cmd.script:
+    - name: salt://remove-tags
+
 # Removes files that are provisioned by the dom0 RPM, only for the development
 # environment, since dnf takes care of those provisioned in the RPM
 {% if d.environment == "dev" %}
 remove-dom0-sdw-config-files-dev:
```
emkll, line 58: We may want to require the `remove-rpc-policy-tags` here to avoid potential race conditions or precedence issues when applying this state on uninstall (though I did not observe any errors in local testing, this might happen if we add tasks in the future)

eloquence: Good point, done in a04589ab56e8aad6a833e06c602eeca508a25af4 (I added the XFCE script because the same logic seems to apply to it); haven't re-tested yet. Can test in dev tomorrow if nobody beats me to it.

eloquence: Re-tested in dev env w/ require logic, relevant output in https://gist.github.com/eloquence/9f9f8586034ccc4a57fe7bda900108cb, can confirm that files continue to be removed as expected.

tests/vars/qubes-rpc.yml
emkll: We can revisit this later, should this be an issue in developer's environment. It should be a small diff that is easily applied on new and existing workstations

Issue #544, opened 2020-05-04
Update Fedora-30 based VMs to Fedora 31


emkll:
Fedora 30 TemplateVMs go end-of-life at the end of this month[1] . We should ensure the Fedora 31 template is added and all workstation-relevant VMs use the Fedora 31 Template. 

[1] : https://www.qubes-os.org/news/2020/04/30/fedora-31-template-available/

eloquence:
Is this a change that we expect will be fully automatic from the admin's perspective (shipped via RPM & updater enforcing dom0 state), or do we already know that additional manual steps will be required?

emkll:
It seems like most of the logic in place from the Fedora 29 to Fedora 30 transition is working properly here for the move to Fedora 31.

With a one line diff to the salt logic (and test fixes), I was able to provision a workstation using fedora-31 system VMs.

We would need to enforce this `sd-sys-vms` state as part of the graphical updater flow, and ensure the logging output would allow us to debug potential errors.

This does not update unmanaged fedora-30 VMs (work, vault or other), and does not remove the fedora-30 template from the system.

Otherwise, this seems like a change we could fully automate as part of the regular update flow.

```
diff --git a/dom0/sd-sys-vms.sls b/dom0/sd-sys-vms.sls
index 5599f82..d1ae401 100644
--- a/dom0/sd-sys-vms.sls
+++ b/dom0/sd-sys-vms.sls
@@ -9,7 +9,7 @@ include:
   # is created.
   - qvm.default-dispvm
 
-{% set sd_supported_fedora_version = 'fedora-30' %}
+{% set sd_supported_fedora_version = 'fedora-31' %}
 
 # Install latest templates required for SDW VMs.
 dom0-install-fedora-template:
diff --git a/tests/test_qubes_vms.py b/tests/test_qubes_vms.py
index e103196..820404c 100644
--- a/tests/test_qubes_vms.py
+++ b/tests/test_qubes_vms.py
@@ -3,7 +3,7 @@ import unittest
 from qubesadmin import Qubes
 
 
-CURRENT_FEDORA_VERSION = "30"
+CURRENT_FEDORA_VERSION = "31"
 CURRENT_WHONIX_VERSION = "15"
 
 
user@dev:~/src/securedrop-workstation$ git diff
diff --git a/dom0/sd-sys-vms.sls b/dom0/sd-sys-vms.sls
index 5599f82..d1ae401 100644
--- a/dom0/sd-sys-vms.sls
[...]

PR #554, opened 2020-05-19
Updates rc naming convention in RPM spec

conorsch:
## Status

Ready for review.

## Description of Changes

Refs #540. Supersedes and therefore closes #552. 

Changes proposed in this pull request:

Following the conventions in [0], we now update both the "Version" and
"Release" fields in order to handle RPMs. Some additional munging of the
"Source0" and "%setup" lines is warranted, as well, to build the correct
filepaths when using a hyphen in the full version spec, e.g. 0.3.0-rc1.

[0] https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_prerelease_versions

The "version" hasn't yet been bumped in this PR, after testing locally, please append a commit with results of `./update-version.sh 0.3.0-rc1`.

## Testing
- [ ] Versioning scheme makes sense (see discussion in #540, cf. #552)
- [ ] Changes from #550 reverted cleanly 
- [ ] Changes to `update_version.sh` work as expected:
  - [ ] `./update_version.sh 0.3.0-rc1` increments and proper sdist naming for tarball and source location, and the package builds using `make dom0-rpm`
  - [ ] `./update_version.sh 0.3.0-rc2` increments the version correctly, and package builds using `make dom0-rpm`
  - [ ] `./update_version.sh 0.3.0` increments the version correctly, and package builds using `make dom0-rpm`
- [ ] Ensure `0.3.0` > `0.3.0-rc1`: Create two rpms and observe the output of `dnf install $RPM` and observe the upgrade/downgrade message at the confirmation prompt.
- [ ] The resulting rpm on this branch works in dom0 (sudo dnf install and `./securedrop-admin --apply`) 

### If you have made code changes

- [ ] Linter (`make flake8`) passes in the development environment (this box may
      be left unchecked, as `flake8` also runs in CI)

### If you have made changes to the provisioning logic

- [ ] All tests (`make test`) pass in `dom0` of a Qubes install

- [ ] This PR adds/removes files, and includes required updates to the packaging
      logic in `MANIFEST.in` and `rpm-build/SPECS/securedrop-workstation-dom0-config.spec`

Issue #539, opened 2020-04-21
Call qubesctl from securedrop-admin Python code


conorsch:
We currently use the "provision-all" bash script to serialize calls to `qubesctl` to configure the SDW VMs. In dev contexts, provision-all is invoked directly. In prod, however, provision-all is called from the `securedrop-admin` Python script. If an error occurs somewhere in provision-all, the Python subprocess call will swallow that specific error, and report a generalized failure of a non-zero exit code for the script.

Let's move the `qubesctl` calls into a def within the python script, so that we'll at least know which specific qubesctl call failed in the event of an error. Note that we cannot (yet) invoke the salt management commands via a python interface: the qubessalt code is still python2, so we must use subprocess.

zenmonkeykstop:
Keeping open, this matches intended direction for provisioning

zenmonkeykstop:
Should be an easy fix that would vastly improve troubleshooting. Let's do it!

PR #555, opened 2020-05-19
Converts F30 -> F31 in dev env

conorsch:
## Status

Ready for review.

## Description of Changes

Refs #544 

These changes ensure that fedora-31 is preferred everywhere. Running a
`make dev` will ensure that the handle-upgrade logic is run, restarting
the VMs as required. What's *not* included in this change is a method
for handling automated upgrades in staging/prod. In order to accomplish
unattended transition from f30 -> f31, we'll have to update the dom0
state management logic to "include" the handle-upgrade script.

## Testing

Make sure you use a `dev` environment (i.e. `config.json` shows "environment": "dev"). 

```
make clone
make all
make test
```

All tests should pass. On my machine, the only test that failed was the dnf update check, so after updating the packages in `fedora-31`, the test suite was passing completely. 

Also run `qvm-ls | grep fedora` and ensure that `sys-net`, `sys-usb`, and `sys-firewall` are all based on `fedora-31`. (The tests check this, but best to be sure.) I've not yet confirmed the usb export flow to be working, so validation on that front is required, as well. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ As of March 2020, the production environment is in beta. The staging environment
 
 **IMPORTANT: THE STAGING ENVIRONMENT SHOULD NEVER BE USED FOR PRODUCTION PURPOSES. IT SHOULD BE USED ON TEST MACHINES ONLY.**
 
-#### Update `dom0`, `fedora-30`, `whonix-gw-15` and `whonix-ws-15` templates
+#### Update `dom0`, `fedora-31`, `whonix-gw-15` and `whonix-ws-15` templates
 Updates to these VMs will be provided by the installer and updater, but to ensure they are up to date prior to install, it will be easier to debug, should something go wrong.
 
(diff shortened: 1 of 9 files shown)


conorsch:
> What's not included in this change is a method for handling automated upgrades in staging/prod. 

On second thought, these changes may also be sufficient for staging/prod. Either way, let's defer testing of staging/prod for now, and focus on the dev story.

We haven't explicitly discussed whether these changes will be included in the 0.3.0 release. Given merge of https://github.com/freedomofpress/securedrop-workstation/pull/554, if this PR gets merged before the 0.3.0 is final, let's make sure to keep the changelog accurate.

conorsch:
Thanks for the detailed review notes, @rmol! 

>  will look into whether we should remove halted management DispVMs at the start

Sounds like a durability measure we'd want to accommodate, but it sounds like even the Qubes upstream updater would be affected. Assuming the disp-mgmt-* VMs were hanging around from a previously interrupted upgrade, we don't currently have a good recovery story, so let's aim to evalute whether our logging would clearly identify that error—I expect the call-python-from-bash approach currently used (cf. #539) means we'd miss the QubesVMInUseErrors.

Issue #576, opened 2020-06-16
Makefile cleanup and culling


eloquence:
The `Makefile` in this repo could use a bit of love:
- `PHONY` and prereqs are not used consistently; for example, if you `touch prod`, the `prod` target will no longer run.
- There are probably some targets that can be culled altogether or made more generic/DRY

Nice task for anyone who wants to get a bit more familiar with `Makefile` syntax.

rocodes:
Notes from backlog pruning:
- Makefile can be cleaned up (correct use of `PHONY`)
- Many make targets that we may want to review to see if we're still using them, including deprecated targets
- Review of more makefiles in other projects as well

PR #569, opened 2020-06-06
Add black formatter check

eloquence:
Resolves #568 

- Adds `black` formatter with 100 character line length per our coding convention
- Adjusts `flake8` settings accordingly
- Reinstates `dev-requirements.in` and `dev-requirements.txt` for managing these dependencies
- Generalizes and improves upon the auto-discovery logic for Python files in this repo
- Adds a `venv` Makefile target for convenience, following @rmol's example in https://github.com/freedomofpress/securedrop-sdk/pull/124
- Stops using the machine type executor in CI for lint jobs and running linters in Docker locally

## Status
Ready for review

### Preparatory steps
- Check out this branch in your development environment
- Prepare a virtualenv using `make venv` and activate it using `source .venv/bin/activate`
- To test revision exclusion, ensure you are running at least `git` version 2.23.x.

###  `flake8` and `black` targets

- Run `make black`. 
  - [ ] Observe the output "All done! 35 files would be left unchanged."
  -  [ ] Observe that the exit code is 0 (`echo $?`)
- Run `make flake8`
  - [ ] Observe that there is no output
  - [ ] Observe that the exit code is 0 (`echo $?`)
- Modify a Python file in this repository that ends with `.py` and introduce an issue that would be flagged by both `flake8` and `black`, e.g., an [E251](https://www.flake8rules.com/rules/E251.html) violation
- Modify a Python file in this repository that has no file extension (e.g., in the `scripts` directory) and introduce a similar issue. 
- Re-run `make flake8` and `make black`
  - [ ] Observe that the issues you introduced are reported for both files.
  - [ ] Observe a that the exit code is non-zero (`echo $?`) for both commands.
- Undo your modifications.

I would recommend additional investigation of the auto-discovery logic introduced in `scripts/lint-all`. We want to make sure that it covers all Python files in this repository, including likely future additions.

### Excluding formatter revisions

- Run the command `git blame scripts/validate_config.py`
- [ ] Observe that several lines are attributed to commit 782a04e5, which made formatting changes.
- Run the command `git blame --ignore-revs-file .git-blame-ignore-revs scripts/validate_config.py`
- [ ] Observe that these same lines are attributed to their original authors, and that commit 782a04e5 is excluded.

PR #577, opened 2020-06-16
Deprecate use of `make all`, display notice instead

### LINKED ISSUE #575: Deprecate `make all`, point to `make dev`, `make prod`, `make staging`
The `make all` target changes your configuration to `dev` without any output or warnings to that effect, which can lead to unexpected build states as observed in #574. As far as I can tell, there's no particular reason to preserve `make all` at this point -- I would recommend we get rid of it (we can output a deprecation notice for now) and just support `make dev`, `make prod` and `make staging`.

eloquence:
## Status

Ready for review

## Description of Changes

Resolves #575
Small step towards #576 

## Testing

- [ ] Clone this branch into `dom0`
- [ ] Verify that all modified `make` targets still run (it may be fine to not let them finish!)
- [ ] Run `make test` to confirm `dom0` tests are still passing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,30 +6,28 @@ ifneq ($(HOST),dom0)
 	exit 1
 endif
 
-all: ## Builds and provisions all VMs required for testing workstation
-	$(MAKE) assert-dom0
+all: assert-dom0
+	@echo "Please run one of the following targets:"
+	@echo
+	@echo "make dev"
+	@echo "make staging"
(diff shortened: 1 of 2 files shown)


conorsch:
@eloquence Conflict due to merge of #569, needs a rebase!

eloquence:
(Rebased; moved `black` target down since we're not really alphabetical and it makes more sense next to `flake8`)

Issue #580, opened 2020-06-26
Release SecureDrop Workstation 0.4.0


eloquence:
This is an issue to track the production release of the SecureDrop Workstation 0.4.0 RPM package, which ships the updater changes in #528. 

## RC preparation
- [x] Prepare 0.4.0-rc1 version
- [x] Upload 0.4.0-rc1 packages to yum-test

## Test plan

All testing should be done with an existing Qubes install at least at `0.4.0-rc1`, dev or staging. **IMPORTANT:** do not attempt to install completely from scratch (without the base template being installed) until #579 is resolved.

### Scenario 1: `fedora-31` case

1. Downgrade a package in `fedora-31` and reboot the VM to force an actual upgrade to applied (e.g., `sudo dnf downgrade zlib`).
2. Run  `/opt/securedrop/launcher/sdw-launcher.py --skip-delta 0`. This forces an updater run.
3. - [ ] Observe that you are able to perform a successful update run with no new regressions (you may still encounter #498).
4. - [ ] Observe that you are able to launch the client, without a reboot warning.
5. - [ ] Observe that `sdw-update-status` and `sdw-last-updated` in `~./securedrop_launcher` contain the expected state (both should include the current timestamp, and `sdw-update-status` should contain status `0` for success)

### Scenario 2: `dom0` case

1. Downgrade a package in `dom0` (see instructions below).
2. Run  `/opt/securedrop/launcher/sdw-launcher.py --skip-delta 0`.
3. - [ ] Observe that you are able to perform a successful update run with no new regressions (you may still encounter #498).
4. - [ ] Observe that you are prompted to reboot after the update. Keep the window open for now.
5. - [ ] Observe that `sdw-update-status` and `sdw-last-updated` contain the expected state (both should include the current timestamp, and `sdw-update-status` should contain status `2` for "reboot required").
6. Click the reboot button and wait for the system to reboot.
7. Log back in and run  `/opt/securedrop/launcher/sdw-launcher.py --skip-delta 0`.
8. After updates, observe that you are able to run the client without a reboot.

### Downgrading a package in `dom0`

The method I used for downgrading (all commands require `sudo` or root shell):
1. Look at `dnf history` in `dom0` to view recently updated packages.
2. Inspect a history entry in detail with `dnf history info <id>`.
3. Pick the older version of a recently updated package using `qubes-dom0-update $package-$version`, e.g. `qubes-dom0-update python3-qubesimgconverter-4.0.27`

[...]

PR #581, opened 2020-06-30
SecureDrop Workstation 0.4.0-rc1

eloquence:
## Status

Ready for review

## Description of Changes

Towards #580

Bumps version to 0.4.0-rc1

## Testing

- [x]  Changelog is accurate


conorsch:
To aid in review, and guide on the rc packaging steps, I've created a new wiki page here: https://github.com/freedomofpress/securedrop-workstation/wiki/Building-securedrop-workstation-dom0-config-RPM-package

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -106,6 +106,10 @@ find /srv/salt -maxdepth 1 -type f -iname '*.top' \
     | xargs qubesctl top.enable > /dev/null
 
 %changelog
+* Mon Jun 29 2020 SecureDrop Team <securedrop@freedom.press> - 0.4.0-rc1
+- Consolidates updates from two stages into one
```
creviera, line 110: should we add pr or issue numbers here?

eloquence: good question. so far we haven't in this changelog. since this is a build artifact that can exist independently of the repo that those IDs reference and that will not be linked to from the outside, my inclination is to avoid those references here

PR #589, opened 2020-07-13
Add sd-log to data flow diagram

emkll:
## Status

Ready for review

## Description of Changes
Adds sd-log to Data Flow Diagram

## Testing

- [ ] Should we add sd-log to the diagram?
- [ ] Is the DFD accurate?

## Checklist

Docs/images only


eloquence:
Looks good. In my `staging` env I do see logs for `sd-gpg` in `sd-log` and I do see `rsyslog` running in `sd-log`, but it's not denoted as such here. Is that intentional?

emkll:
@eloquence that is an omission in the diagram, will update

eloquence:
LGTM!

Issue #590, opened 2020-07-16
Latest kernel shows problems on some hardware


conorsch:
After we shipped 4.14.186 kernels as part of #546, we received a report of distorted graphics after upgrading. The behavior described was quite similar to that documented in https://github.com/freedomofpress/securedrop-workstation/issues/308#issuecomment-521288132

While we didn't catch the issue during QA, I was able to reproduce on test-only hardware. After reviewing logs, the problem seems to correlate with this event in syslog:

```
FATAL: Module u2mfn not found in directory /lib/modules/4.14.186-grsec-workstation
```

It appears that the `dkms autoinstall` line in the postinst for `securedrop-workstation-grsec` https://github.com/freedomofpress/securedrop-debian-packaging/blob/e7d5bea3f2eb6bbbc7ad76772ec42b4610830916/securedrop-workstation-grsec/debian/postinst#L42 is failing, but still exiting zero—so apt/dpkg didn't consider it an error. As a workaround, I was able to rebuild the dkms projects after bouncing paxctld and the situation was resolved. Let's try updating the postinst logic to restart the paxctld service _before_ running `dkms autoinstall`. 

<details>
<summary>Detailed dom0 logs </summary>

```
[user@dom0 ~]$ qvm-run sdw-backup gnome-terminal # cloned from problematic securedrop-workstation-buster
Running 'gnome-terminal' on sdw-backup
[user@dom0 ~]$ # confirmed garbled terminal
[user@dom0 ~]$ qvm-run -p sdw-backup 'sudo dpkg-reconfigure --frontend=noninteractive securedrop-workstation-grsec'

Kernel preparation unnecessary for this kernel.  Skipping...

Building module:
cleaning build area...
make -j2 KERNELRELEASE=4.14.186-grsec-workstation -C /lib/modules/4.14.186-grsec-workstation/build M=/var/lib/dkms/u2mfn/4.0.30/build...
cleaning build area...

DKMS: build completed.

u2mfn.ko:
Running module version sanity check.
 - Original module
   - No original module exists within this kernel
 - Installation
   - Installing to /lib/modules/4.14.186-grsec-workstation/updates/dkms/

depmod....
Job for systemd-modules-load.service failed because the control process exited with error code.

DKMS: install completed.
See "systemctl status systemd-modules-load.service" and "journalctl -xe" for details.
Synchronizing state of paxctld.service with SysV service script with /lib/systemd/systemd-sysv-install.
[...]

Issue #583, opened 2020-07-06
sd-log sometimes ingests logs as `host`


eloquence:
As reported [here](https://github.com/freedomofpress/securedrop-workstation/pull/563#issuecomment-642108426) and [here](https://github.com/freedomofpress/securedrop-workstation/pull/582#issuecomment-654514611), logs are sometimes ingested under the `host` directory instead of the VM's actual name, potentially due to a race condition. This may be limited to the `sd-whonix` VM.

eloquence:
(Bumping priority as these test failures have caused significant confusion.)

eloquence:
I'll take a 4 hour timebox against this issue during the 9/17-10/1 sprint; self-assigning accordingly.

eloquence:
Started on a timebox today. In my staging env, I found a `host` dir that was last written to on September 11, but currently `sd-whonix` correctly logs into its directory on `sd-log`. Repeatedly restarting so far has not yielded a repro. Could be an issue at provisioning-time where, once correctly provisioned, it works as expected.

conorsch:
> Could be an issue at provisioning-time where, once correctly provisioned, it works as expected.

That's right, it'll be a first-run problem, specific to the Whonix Gateway VM. If you inspect the logs available in `sd-log:QubesIncomingLogs/host/*`, you should see tor-related messaging. More research required to sort out whether it's `whonix-gw-15` (TemplateVM) or `sd-whonix` (AppVM) that's briefly misconfigured at install time.

eloquence:
I reprovisioned my staging environment and observed the logs carefully. The `host` log dir was getting populated with `sd-whonix` logs during the "Provision all SecureDrop Workstation VMs" stage. However, I see `host` getting written to _after every restart_ of `sd-whonix` as well, well after provisioning, regardless of the template boot status. More on that below.

`whonix-gw-15` does not get logged at all (running `logger` in the template does not result in any logs being written in `sd-log`). It's not tagged with `sd-workstation`, but our RPC policies require this tag to be set:

https://github.com/freedomofpress/securedrop-workstation/blob/main/dom0/sd-log.sls#L50-L51

I'm assuming that's the desired behavior given that we generally are more interested in AppVM logs than TemplateVM logs. As far as I can tell, it's only the RPC policies that prevent logs from being sent to `sd-log`, otherwise it would be happily ingesting those logs as `host` as well.

[...]

PR #604, opened 2020-08-26
Adds config test for required kernel modules

conorsch:
## Status
Ready for review.

## Description of Changes

These changes add new config tests for VM state, ensuring that the u2mfn kernel module, required for graphical environments in Qubes, was built accurately. Without these tests, the graphical display problems reported in https://github.com/freedomofpress/securedrop-workstation/issues/590 wouild not result in a test suite failure. Now the tests should fail accordingly. 

## Testing

See detailed test plan in https://github.com/freedomofpress/securedrop-debian-packaging/pull/189 to validate the logic changes in the new metapackage. For reviewing this PR, a fully passing `make test` is sufficient. Note that you may see a failing test such as https://github.com/freedomofpress/securedrop-workstation/issues/583 , which is unrelated.

Additionally, it's probably worth confirming that removing the filepath `/usr/lib/modules/4.14.186-grsec-workstation/updates/dkms/u2mfn.ko` from a TemplateVM causes the new test to fail, although be aware that removing that file will break the GUI as described in https://github.com/freedomofpress/securedrop-workstation/issues/590

### DIFF
--- a/tests/test_vms_exist.py
+++ b/tests/test_vms_exist.py
@@ -36,6 +36,12 @@ class SD_VM_Tests(unittest.TestCase):
         assert kernel_version.endswith("-grsec-workstation")
         assert kernel_version == EXPECTED_KERNEL_VERSION
 
+        u2mfn_filepath = "/usr/lib/modules/{}/updates/dkms/u2mfn.ko".format(EXPECTED_KERNEL_VERSION)
+        # cmd will raise exception if file not found
+        stdout, stderr = vm.run("sudo test -f {}".format(u2mfn_filepath))
+        assert stdout == b""
+        assert stderr == b""
+
     def _check_service_running(self, vm, service):
         """
         Ensures a given service is running inside a given VM.


emkll:
Testing for this pr is currently blocked on https://github.com/freedomofpress/securedrop-workstation/issues/606 , the new kernel metapackage being tested here is not yet on apt-test.freedom.press

conorsch:
@emkll You're right that https://github.com/freedomofpress/securedrop-workstation/issues/606 deserves immediate attention (looking into it now) but technically it needn't block here. The ko filepath being tested here has always been present—for each kernel version, and the kernel version has not changed recently—and its absence will result in a non-working GUI. I've amended the plan to point this out. 

Fine to wait on review here, I'd just like to make sure that if the GUI is broken in a way we can observe, we have a corresponding test—otherwise, the config tests say "Everything's fine!" when the AppVMs are unusable.

conorsch:
https://github.com/freedomofpress/securedrop-workstation/pull/606 is re-closed, so ready for testing here.

tests/test_vms_exist.py
```diff
@@ -36,6 +36,12 @@ def _check_kernel(self, vm):
         assert kernel_version.endswith("-grsec-workstation")
         assert kernel_version == EXPECTED_KERNEL_VERSION
 
+        u2mfn_filepath = "/usr/lib/modules/{}/updates/dkms/u2mfn.ko".format(EXPECTED_KERNEL_VERSION)
+        # cmd will raise exception if file not found
+        stdout, stderr = vm.run("sudo test -f {}".format(u2mfn_filepath))
+        assert stdout == ""
```
emkll, line 42: Looks like adding `.decode("utf-8")` to stdout/stderr resolves the tests for me locally

emkll (changes requested):
Thanks @conorsch for the tests! Looks like the tests are failing for me locally to to string/bytes comparison (see inline).

tests/test_vms_exist.py
conorsch: Thanks, you're right! Updated the string checks `""` -> `b""`, now passing for me.

emkll (approved):
lgtm !

PR #617, opened 2020-10-01
Set hostname at template level for whonix-gw-15

### LINKED ISSUE #583: sd-log sometimes ingests logs as `host`

eloquence:
## Status

Ready for initial feedback. Additional testing & test plan to come.

## Description of Changes

Resolves #583. Sets hostname at template level to ensure that logs are ingested as `sd-whonix` at boot time.

### DIFF
--- a/dom0/sd-logging-setup.sls
+++ b/dom0/sd-logging-setup.sls
@@ -61,42 +61,35 @@ sd-gpg-remove-rsyslog-qubes-plugin:
     - name: /rw/config/rc.local
     - require:
       - file: sd-gpg-remove-rsyslog-qubes-plugin
-
-{% elif grains['id'] == "sd-whonix" %}
-# We can not place the file on the template under /etc/rsyslog.d/ because of whonix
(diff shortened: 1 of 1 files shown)


eloquence:
(In early testing this seems to work as expected, but I've not attempted a fresh provisioning run with this change yet.)

eloquence:
Preliminary testing indicates that a fix in `securedrop-log` (using `qubesdb-read /name` instead of `gethostname()`) will work better than this approach with no need to special-case `sd-whonix`. It'll also let us enable logging for the template, if we want it. Closing this PR; will revisit if we run into issues.

PR #618, opened 2020-10-02
Remove special casing for sd-whonix

eloquence:
Part of #583

## Description

Removes Salt logic for custom handling of `whonix-gw-15` integration with `sd-log` and instead treats it like any other TemplateVM.

Removes the `vmname` testing from `dom0` tests. The functionality remains in `securedrop-log`, but is only likely to be used during development.

## Status

Ready for review. 

### Prerequisites 
- The changes in https://github.com/freedomofpress/securedrop-log/pull/18 have been merged and have landed in the latest nightly on https://apt-test.freedom.press/pool/main/s/securedrop-log/
- You have a working staging environment built from latest `main` of `securedrop-workstation`.

### Instructions
1. Remove  `~/QubesIncomingLogs/host` directory on `sd-log` if it exists from previous runs
2. `make clone` from this branch.
3. `make staging` with a valid config. This will apply the Salt states from this PR.
4. Ensure the `sd-whonix` VM, the `sd-app` VM, and the `sd-log` VM are all booted.
- [ ] On `sd-whonix`: Observe that `/rw/config/rc.local` no longer contains any customizations
- [ ] On `sd-whonix`: Observe that `/rw/config/sdlog.conf` no longer exists.
- [ ] On `sd-log`: Observe that no files exist under `host` in `~/QubesIncomingLogs`
- [ ] On `sd-log`: Observe that logs are correctly ingested under `sd-whonix`.
- [ ] On `sd-log`: Observe that logs are correctly ingested under `sd-app`.
5. `make clean` in `dom0`.
- [ ] On `whonix-gw-15`: Observe that `/etc/rsyslog.d/sdlog.conf` has been removed

### DIFF
--- a/dom0/sd-clean-whonix.sls
+++ b/dom0/sd-clean-whonix.sls
@@ -14,7 +14,6 @@ remove-securedrop-log-package-from-whonix:
 sd-cleanup-whonix-gw-15:
   cmd.run:
     - names:
-      - sudo rm -f /etc/rsyslog.d/sdlog.conf
       - sudo rm -f /etc/apt/sources.list.d/securedrop_workstation.list
       - sudo systemctl restart rsyslog
       - sudo apt-key del 4ED79CC3362D7D12837046024A3BE4A92211B03C
--- a/dom0/sd-logging-setup.sls
+++ b/dom0/sd-logging-setup.sls
@@ -62,41 +62,33 @@ sd-gpg-remove-rsyslog-qubes-plugin:
     - require:
       - file: sd-gpg-remove-rsyslog-qubes-plugin
 
-{% elif grains['id'] == "sd-whonix" %}
-# We can not place the file on the template under /etc/rsyslog.d/ because of whonix
-# template. This sdlog.conf file is the same from the securedrop-log package, to
-# make sure that rsyslogd use our logging plugin.
-sd-rsyslog-sdlog-conf-for-sd-whonix:
-  file.managed:
-    - name: /rw/config/sdlog.conf
-    - source: "salt://sdlog.conf"
-
-# Because whonix-gw-15 template is not allowing to create the config file on
-# package install time, we do it via rc.local call.
-sd-rc-enable-logging-for-sd-whonix:
-  file.blockreplace:
-    - name: /rw/config/rc.local
-    - append_if_not_found: True
-    - marker_start: "### BEGIN securedrop-workstation ###"
-    - marker_end: "### END securedrop-workstation ###"
-    - content: |
-        # Add sd-rsyslog.conf file for syslog
-        ln -sf /rw/config/sdlog.conf /etc/rsyslog.d/sdlog.conf
-        cat <<EOF > /etc/sd-rsyslog.conf
-        [sd-rsyslog]
-        remotevm = sd-log
-        localvm = {{ grains['id'] }}
-        EOF
-        systemctl restart rsyslog
-  cmd.run:
-    - name: /rw/config/rc.local
-    - require:
-      - file: sd-rc-enable-logging-for-sd-whonix
-
 {% else %}
 # For all other VMs, configure to send to sd-log
 configure-rsyslog-for-sd:
   file.managed:
     - name: /etc/sd-rsyslog.conf
     - source: "salt://sd-rsyslog.conf.j2"
+{% endif %}
+
+# Remove outdated configuration that was previously used to configure the
+# sd-whonix VM name for logging purposes, see:
+# https://github.com/freedomofpress/securedrop-workstation/issues/583
+#
+# Can be removed in a future release once all production workstations have
+# been updated.
+{% if grains['id'] == "sd-whonix" %}
+sd-whonix-cleanup-rc-local:
+  file.replace:
+    - names:
+      - /rw/config/rc.local
+    - pattern: '### BEGIN securedrop-workstation ###.*### END securedrop-workstation ###\s*'
+    - flags:
+      - MULTILINE
+      - DOTALL
+    - repl: ''
+    - backup: no
+
+sd-whonix-cleanup-sdlog-conf:
+  file.absent:
+    - name: /rw/config/sdlog.conf
 {% endif %}
--- a/tests/test_sd_whonix.py
+++ b/tests/test_sd_whonix.py
@@ -55,7 +55,7 @@ class SD_Whonix_Tests(SD_VM_Local_Test):
         assert self._fileExists(self.whonix_apt_list)
 
     def test_logging_configured(self):
-        self.logging_configured(vmname=True)
+        self.logging_configured()
 
     def test_sd_whonix_verify_tor_config(self):
         # User must be debian-tor for v3 Onion, due to restrictive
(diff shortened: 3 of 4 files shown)


dom0/sd-logging-setup.sls
```diff
@@ -62,41 +62,33 @@ sd-gpg-remove-rsyslog-qubes-plugin:
...
-# We can not place the file on the template under /etc/rsyslog.d/ because of whonix
-# template. This sdlog.conf file is the same from the securedrop-log package, to
-# make sure that rsyslogd use our logging plugin.
-sd-rsyslog-sdlog-conf-for-sd-whonix:
-  file.managed:
-    - name: /rw/config/sdlog.conf
-    - source: "salt://sdlog.conf"
-
-# Because whonix-gw-15 template is not allowing to create the config file on
-# package install time, we do it via rc.local call.
```
eloquence, line 75: For the record, this was added in 232c56f44a34d393b89b0578364e955da62f1576. @kushaldas, can you elaborate on what you meant at the time with "whonix-gw-15 template is not allowing to create the config file on package install time"? As far as I can tell, the config file `/etc/rsyslog.d/sdlog.conf` is installed just fine from the package without it, and I've not observed any behavior specific to `whonix-gw-15` that would prevent this from being the case.

dom0/sd-clean-whonix.sls
```diff
@@ -14,7 +14,6 @@ remove-securedrop-log-package-from-whonix:
 sd-cleanup-whonix-gw-15:
   cmd.run:
     - names:
-      - sudo rm -f /etc/rsyslog.d/sdlog.conf
```
eloquence, line 17: Note that we use `pkg.purged` to remove the `securedrop-log` package, which [removes config files](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.pkg.html):

> Verify that a package is not installed, calling pkg.purge if necessary to purge the package. All configuration files are also removed.

Removing a configuration file on a Debian system by other means can have unpredictable consequences: the package manager will not reinstate it , because it will assume that the user removed it intentionally. For this reason, I think it's best to rely on `pkg.purged` alone here.

eloquence:
I've added a test plan. I would recommend the following order of operations:
1) Initial visual review by @kushaldas of the logic changes here and in https://github.com/freedomofpress/securedrop-log/pull/18. Specifically, if there' s anything I'm overlooking that makes the use of `/rw/config` necessary for `sd-whonix`, that would be good to know (see comment here: https://github.com/freedomofpress/securedrop-workstation/pull/618#discussion_r499981376)
2) Review and merge of https://github.com/freedomofpress/securedrop-log/pull/18
3) Review and merge of this PR.

dom0/sd-logging-setup.sls
kushaldas: https://github.com/freedomofpress/securedrop-workstation/pull/447/ this is the PR and as I can see in my logs, the file was still missing even after the package was installed.

Here is my note from that work:

```
* https://github.com/freedomofpress/securedrop-workstation/pull/447  howt to remove the files?
  - how to remove gpg key? sudo apt-key del KEY_ID
  - `make clean`
  - remove works properly 
  - `make all` agagin.
  - most of the vms decided not to reply to Qubes calls :(
  - cleaned, and `make all` again 
  - Someday Qubes and Salt will play nicely together. Question is in which year?
  - this try, the package is there, but, no logs. 
  - because the sdlog.conf file is missing, WHY WHY WHY
  - I this before, i commented on the issue, dpkg did not create the sdlog.conf file at all on the system.
  - one idea, if the file is missing, just copy it over via rc.local file
  - testing again 
  - testing again, adding sdlog.conf only for sd-whonix if missing.
  - testing once more, things look better now
  - meeting with Mickael
  - need to test some more changes
  - make clean and make all again at super late night :)
```

eloquence: Hi @kushaldas, I experienced this once during testing with `dpkg`, and in my experience it happens when you've previously removed the file by hand and uninstalled (rather than purged) the package - regardless of the order of the two operations. It's not specific to Whonix, and is just how Debian handles configuration files. Because we purge the file correctly with our uninstall logic, I don't believe it should be an issue on production systems.

conorsch:
See comment in https://github.com/freedomofpress/securedrop-log/pull/18#pullrequestreview-505242284: we must also update the `tests/test_sd_whonix.py` checks here in order to get all tests passing!

eloquence:
> we must also update the tests/test_sd_whonix.py checks here in order to get all tests passing!

Roger that, will update the tests on Monday and then promote to ready. :)

eloquence:
This is ready for review, but to take it for a spin with the associated securedrop-log changes, it would IMO be best to kick off a nightly build of that package; nightlies seem to have gotten stuck around October 5.

tests/test_sd_whonix.py
```diff
@@ -54,9 +54,6 @@ def test_sd_whonix_repo_enabled(self):
         """
         assert self._fileExists(self.whonix_apt_list)
 
-    def test_logging_configured(self):
-        self.logging_configured(vmname=True)
-
```
conorsch, line 59: The `test_logging_configured` shouldn't be removed, it should just have its call updated to pass no args (matching the modified test logic).

conorsch:
>  nightlies seem to have gotten stuck around October 5.

You're right! Resolved out of band. Latest version of `securedrop-log` is now served on apt-test, so testers should not see a `sd-log:~/QubesIncomingLogs/host/` dir created any more. 

Started, but did not finish review. Left one comment about the test logic—to any future reviewers, e.g. @zenmonkeykstop, please tack on a test if you agree. Also make sure to run `make test` as part of review, since we should see 100% of tests passing now.

tests/test_sd_whonix.py
eloquence: Fixed (I misread the inheritance logic).

conorsch:
Reviewing now!

rmol (approved):
:+1: Ran through the test plan, and other than `sd-app` not being started by `make staging`, the changes specific to this PR look good and worked.

1. Remove  `~/QubesIncomingLogs/host` directory on `sd-log` if it exists from previous runs
2. `make clone` from this branch.
3. `make staging` with a valid config. This will apply the Salt states from this PR.
4. Ensure the `sd-whonix` VM, the `sd-app` VM, and the `sd-log` VM are all booted.

:warning: `sd-app` was _not_ booted after `make staging` completed.

- [x] On `sd-whonix`: Observe that `/rw/config/rc.local` no longer contains any customizations
- [x] On `sd-whonix`: Observe that `/rw/config/sdlog.conf` no longer exists.
- [x] On `sd-log`: Observe that no files exist under `host` in `~/QubesIncomingLogs`
- [x] On `sd-log`: Observe that logs are correctly ingested under `sd-whonix`.
- [x] On `sd-log`: Observe that logs are correctly ingested under `sd-app`.
5. `make clean` in `dom0`.
- [x] On `whonix-gw-15`: Observe that `/etc/rsyslog.d/sdlog.conf` has been removed

conorsch:
> sd-app was not booted after make staging completed.

That makes sense: we don't actually enforce that state (although we should). So, my read of the test plan is "manually start those VMs before you proceed to the next step."

PR #623, opened 2020-10-14
Fix profile.d tests for sd-app

### LINKED ISSUE #621: Many test failures in dom0 due to recent sd-app changes
## Steps to reproduce

Run `make test` in your `dom0` checkout with a recently provisioned SecureDrop Workstation using latest nightlies.

## Expected behavior

Tests pass

## Actual behavior

Lots of test failures, mainly due to the string "Setting /usr/bin/securedrop-client to enforce mode" appearing in command output. Example: 

```
FAIL: test_mimeapps (test_app.SD_App_Tests)
  File "/home/erik/securedrop-workstation/tests/test_app.py", line 30, in test_mimeapps
    self.assertEqual(results, expected_results)
AssertionError: 'Setting /usr/bin/securedrop-client to enf[46 chars]top;' != '2 \n    295 open-in-dvm.desktop;'
+ 2 
- Setting /usr/bin/securedrop-client to enforce mode.
-       2 
      295 open-in-dvm.desktop;
```

## Explanation

Nightlies broke on October 5 and were just fixed today, so we did not notice this earlier. https://github.com/freedomofpress/securedrop-client/pull/1153 changed the logic for updating the `QUBES_GPG_DOMAIN` to also invoke `sudo aa-enforce /usr/bin/securedrop-client`. That command produces the aforementioned output. Because this is done in `profile.d`, the command runs on every `qvm-run sd-app` command from `dom0`, which breaks the `sd-app` tests.

We can redirect the `aa-enforce` output to `/dev/null`, which would fix most of the tests, but that assumes we really want to run `aa-enforce` this frequently. In any event, [this test](https://github.com/freedomofpress/securedrop-workstation/blob/main/tests/test_app.py#L12-L15) directly reads the contents of the file that's been modified as part of https://github.com/freedomofpress/securedrop-client/pull/1153, and will still need to be updated.

emkll:
## Status

Ready for review
This PR should be reviewed at the same time as https://github.com/freedomofpress/securedrop-client/pull/1159 and https://github.com/freedomofpress/securedrop-debian-packaging/pull/202. 

## Description of Changes

Fixes #621
`QUBES_GPG_DOMAIN` is conditionally set based on the running VM in order to support template consolidation. See https://github.com/freedomofpress/securedrop-client/issues/1141

This will also help testing both client and packaging PRs that resolve these test failures.

## Testing

Build a client package based on the following branches/PRs:
- client: https://github.com/freedomofpress/securedrop-client/pull/1159
- packaging: https://github.com/freedomofpress/securedrop-debian-packaging/pull/202
- [ ] Client builds successfully

Apply the diff below to install the sd-app package locally instead of through apt server
```
diff --git a/dom0/sd-app-files.sls b/dom0/sd-app-files.sls
index 9828d4e..55b3b18 100644
--- a/dom0/sd-app-files.sls
+++ b/dom0/sd-app-files.sls
@@ -14,8 +14,9 @@ include:
 
 # FPF repo is setup in "securedrop-workstation" template
 install-securedrop-client-package:
-  pkg.installed:
-    - pkgs:
-      - securedrop-client
-    - require:
-      - sls: fpf-apt-test-repo
+  file.managed:
+   - name: /opt/securedrop-client.deb
+   - source: salt://sd/sd-workstation/securedrop-client_0.2.1+buster_all.deb
+   - mode: 644
+  cmd.run:
+   - name: apt install --allow-downgrades -y /opt/securedrop-client.deb
```
- [ ] `make dev` should complete successfully
- [ ] `make test` should complete successfully  (except sd-app being up-to-date, this is because the client version needs to be bumped)

- [x] All tests (`make test`) pass in `dom0` of a Qubes install (except the upgrades for sd-app (the client version needs to be bumped)

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -9,10 +9,8 @@ class SD_App_Tests(SD_VM_Local_Test):
         self.vm_name = "sd-app"
         super(SD_App_Tests, self).setUp()
 
-    def test_decrypt_sd_user_profile(self):
-        contents = self._get_file_contents("/etc/profile.d/sd-app-qubes-gpg-domain.sh")
-        expected_content = 'export QUBES_GPG_DOMAIN="sd-gpg"\n'
-        self.assertEqual(contents, expected_content)
+    def test_gpg_domain_configured(self):
+        self.qubes_gpg_domain_configured(self.vm_name)
 
     def test_open_in_dvm_desktop(self):
         contents = self._get_file_contents("/usr/share/applications/open-in-dvm.desktop")
--- a/tests/base.py
+++ b/tests/base.py
@@ -125,3 +125,27 @@ remotevm = sd-log
         # so let's investigate further.
         # cmd_output = self._run("sudo grep -F \"action 'action-0-omprog' suspended (module 'omprog')\" /var/log/syslog | wc -l").strip()  # noqa
         # self.assertTrue(cmd_output == "0")
+
+    def qubes_gpg_domain_configured(self, vmname=False):
+        """
+        Ensure the QUBES_GPG_DOMAIN is properly set for a given AppVM. This
+        var is set by a script /etc/profile.d.
+        sd-app should have it set to sd-gpg.
+        All other AppVMs should not have this configured.
+        """
+        expected_profile_app = 'if [ "$(qubesdb-read /name)" = "sd-app" ]; then export QUBES_GPG_DOMAIN="sd-gpg"; fi\n'  # noqa: E501
+        expected_env_app = "sd-gpg"
+        expected_env_all = ""
+        profile_path = "/etc/profile.d/sd-app-qubes-gpg-domain.sh"
+
+        env_cmd = 'echo "$QUBES_GPG_DOMAIN"'
+        env_contents = self._run(env_cmd)
+
+        if vmname == "sd-app":
+            self.assertTrue(self._fileExists(profile_path))
+            profile_contents = self._get_file_contents(profile_path)
+            self.assertEqual(profile_contents, expected_profile_app)
+            self.assertEqual(env_contents, expected_env_app)
+        else:
+            self.assertFalse(self._fileExists(profile_path))
+            self.assertEqual(env_contents, expected_env_all)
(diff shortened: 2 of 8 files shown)


tests/test_app.py
```diff
@@ -11,7 +11,7 @@ def setUp(self):
 
     def test_decrypt_sd_user_profile(self):
         contents = self._get_file_contents("/etc/profile.d/sd-app-qubes-gpg-domain.sh")
-        expected_content = 'export QUBES_GPG_DOMAIN="sd-gpg"\n'
+        expected_content = 'if [ "$(qubesdb-read /name)" = "sd-app" ]; then export QUBES_GPG_DOMAIN="sd-gpg"; fi\n'  # noqa: E501
```
eloquence, line 14: This'll work, but to make this more of a test of behavior rather than implementation I wonder if it might be better to check if the output of `qvm-run -p sd-whonix "echo \$QUBES_GPG_DOMAIN"` is the expected `sd-gpg` -- that way we could also run the inverse test in the other AppVMs, and would be more independent of the implementation.

emkll: good call, updated the test in f65b62c

eloquence: Nice! In case you didn't already consider and reject that in this case, note that have a `base.py` we can use to avoid duplication of test code -- see `logging_configured` as an example. Because these helper functions don't start with `test`, they won't be run as part of the test suite unless explicitly invoked with `test_*` function in child classes, as we do for the logging test.

zenmonkeykstop (dismissed):
Built client and patched as per test plan.
- [x] `make dev` completed happily
- [x] `make test` completed successfully with the exception of `test_all_sd_vms_uptodate`, which only failed because the securedrop-client nightly is available as an upgradable option

LGTM once it's flipped out of "in dev" status - suggested test refactoring would be nice to have but I don't think it should block merge.

emkll:
Updated tests based on feedback and rebased on latest `main`

conorsch (approved):
Works well. Confirming all tests passing, sans one, which is an artifact of the manual deb installation, noted in the test plan.

PR #636, opened 2020-11-05
ensure default config backed up by qa switcher script

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Minor fix to switcher script - it wasn't backing up the default config and was using template-consolidation pool if old version of switcher script was present.

QA-only.


conorsch:
Can you also add references to the qa-switcher paths to the clean state? As it is, these files will linger even after the dev environment is purged via `make clean`.

conorsch (approved):

PR #654, opened 2021-02-04
Update CODE_OF_CONDUCT.md

creviera:
fix typos


conorsch (approved):
Thanks, @creviera! I'm optimistic we can start using https://github.com/freedomofpress/.github/pull/1 soon, but in the meantime, let's definitely keep this copy up to date.

Issue #657, opened 2021-02-05
Preflight updater tests have started failing in CI


eloquence:
Can be seen in https://app.circleci.com/pipelines/github/freedomofpress/securedrop-workstation/308/workflows/c6fbf290-5585-4908-a442-eecf77fee31e/jobs/3483 and on recent PRs, e.g. #654 and #656.

eloquence:
I investigated this a bit. This test is failing in CI, but not locally, and not if I run it in a Docker container matching our CircleCI config:

https://github.com/freedomofpress/securedrop-workstation/blob/main/launcher/tests/test_util.py#L53

As you can see in the debug output here, the `lsof` command actually does not produce the output we expect:

https://app.circleci.com/pipelines/github/freedomofpress/securedrop-workstation/311/workflows/4d64914e-c1df-492d-a1f1-b9397e1e4d50/jobs/3498

For whatever reason, in CI, the test process does not successfully obtain an exclusive lock, but it also does not error out. I'm wondering if this is a result of CI-only caching behavior. Will do some more testing but if folks have any specific hypotheses why we're seeing this behavior only in CI, I'm all ears.

eloquence:
I'm not even able to reproduce the issue using [CircleCI's CLI](https://circleci.com/docs/2.0/local-cli/) -- if I run `circleci local execute --job launcher-tests-buster` locally on `main`, it works just fine.

#658 is a proof-of-concept for unbreaking the tests by switching this job to a machine executor. This does require a PyQt bump for our development version and the use of Python 3.9.1, unless we do additional work to install the Python version we want in the image.

Of course, this may be overkill for this one test, though it's nice to verify that the lockfiles work as expected because any breakage could cause bugs that are confusing and difficult to notice.

eloquence:
We didn't discuss in sprint plan today, but I'll poke a bit more at this and will discuss with @conorsch if I get stuck; should be resolved before next SDW release.

PR #658, opened 2021-02-09
Proof of concept for unbreaking launcher tests via machine type

eloquence:
Refs #657

Note: This repository is configured to only run CI on PRs, so we have to open PRs to compare output. 

This draft PR demonstrates getting the `launcher-tests-buster` job to succeed by switching to the machine executor type. However, only Python 2.7.18 and 3.9.1 are available with the Ubuntu 20.04 image. To use Python 3.9.1, we also have to bump to a more recent version of PyQt.


eloquence:
(Closing for now)

eloquence:
Closing in favor of #659, which does not come with the performance penalty of switching to the machine type.

Issue #598, opened 2020-08-05
make dom-rpm failing on Fedora


kushaldas:
It failed with the following error

```
Please provide a source image with `from` prior to commit
```

## Steps to reproduce

- [ ] checkout the code in a Fedora Vm
- [ ] Have docker installed and running (`systemstl start docker`)
- [ ] `make dom0-rpm`

conorsch:
As of https://github.com/freedomofpress/securedrop-workstation/pull/666, this isn't an issue anymore: `make dom0-rpm` appears to work just fine under Fedora. The `make install-deps` still assumes Debian, which is a minor inconvenience, but overall I believe we're covered here.

eloquence:
(Closing per above)

PR #666, opened 2021-03-13
Builds dom0 RPM reproducibly, removes docker requirement for building RPMs

### LINKED ISSUE #481: Update Fedora docker image
The Fedora docker image [1] used to build packages is based on Fedora-29. We should update it to a later version, as Fedora 29 is now end-of-life [2]

[1] https://github.com/freedomofpress/securedrop-workstation/blob/master/docker/CreateRepoRPM/Dockerfile#L2
[2] https://fedoraproject.org/wiki/End_of_life

### LINKED ISSUE #628: Make dom0 rpm (securedrop-workstation-dom0-config) reproducible
Currently, the artifact produced by the `make dom0-rpm` is not reproducible, each time the command is run, an artifact with a different digest is created. Given that it contains mostly some salt definitions and Python, it may be possible to obtain an identical artifact when running the command.

conorsch:
## Status

Ready for review  

## Description of Changes

Fixes #628. Supersedes and therefore closes #481. Refs #598, as well, although I should comment other there separately.

Updates the build logic for the `securedrop-workstation-dom0-config` package so that builds are fully reproducible. Generates RPMs for F25 & F32 (that is, Qubes 4.0.x & 4.1.x, respectively), same as before. Both are reproducible. Removes the custom Dockerfile logic, and now performs builds inside a Debian stable environment (Debian 10 Buster, at time of writing). Updates the CircleCI logic to include use of reprotest.

### Confirming reproducibility
This is the easy part! 

- [x] `make install-deps` && `make dom0-rpm` builds the packages
- [x] Take note of the checksums, then delete the `.rpm` files that were created.
- [x] Re-run `make dom0-rpm`, and compare the checksums manually. They should be the same!
- [x] `make reprotest` reports identical checksums, which is essentially a wrapper for the above. N.B. if you don't frequently use `reprotest`, you may see failures related to system-wide configs. Try `make reprotest-ci`, which has fewer variations, and that should certainly pass for you.

### Confirm package contents are the same as prior to reproducibility
This is the part that requires more careful review. We want to be sure that in the quest for reproducibility, we haven't inadvertently altered the contents and logic of the package that we're shipping. To be clear, there are minor changes in the package contents, but they are quite minor and I believe warranted.

<details>
<summary>Full list of removed files</summary>

```
├── content
│ ├── file list
│ │ @@ -81,30 +81,19 @@
│ │  ./srv/salt/sdlog.conf
│ │  ./srv/salt/securedrop-check-migration
│ │  ./srv/salt/securedrop-handle-upgrade
│ │  ./srv/salt/securedrop-launcher.desktop
│ │  ./srv/salt/securedrop-login
│ │  ./srv/salt/update-xfce-settings
│ │  ./usr/bin/sdw-admin
│ │ -./usr/lib/python3.5/site-packages/securedrop_workstation_dom0_config-0.5.3-py3.5.egg-info
│ │ -./usr/lib/python3.5/site-packages/securedrop_workstation_dom0_config-0.5.3-py3.5.egg-info/PKG-INFO
│ │ -./usr/lib/python3.5/site-packages/securedrop_workstation_dom0_config-0.5.3-py3.5.egg-info/SOURCES.txt
│ │ -./usr/lib/python3.5/site-packages/securedrop_workstation_dom0_config-0.5.3-py3.5.egg-info/dependency_links.txt
│ │ -./usr/lib/python3.5/site-packages/securedrop_workstation_dom0_config-0.5.3-py3.5.egg-info/top_level.txt
│ │ -./usr/share/doc/securedrop-workstation-dom0-config
│ │  ./usr/share/doc/securedrop-workstation-dom0-config/LICENSE
│ │  ./usr/share/doc/securedrop-workstation-dom0-config/README.md
│ │  ./usr/share/securedrop-workstation-dom0-config
│ │ -./usr/share/securedrop-workstation-dom0-config/Makefile
│ │  ./usr/share/securedrop-workstation-dom0-config/config.json.example
│ │  ./usr/share/securedrop-workstation-dom0-config/scripts
│ │ -./usr/share/securedrop-workstation-dom0-config/scripts/build-dom0-rpm
│ │  ./usr/share/securedrop-workstation-dom0-config/scripts/clean-salt
│ │ -./usr/share/securedrop-workstation-dom0-config/scripts/clone-to-dom0
│ │  ./usr/share/securedrop-workstation-dom0-config/scripts/configure-environment
│ │  ./usr/share/securedrop-workstation-dom0-config/scripts/destroy-vm
│ │  ./usr/share/securedrop-workstation-dom0-config/scripts/lint-all
│ │ -./usr/share/securedrop-workstation-dom0-config/scripts/prep-dev
│ │  ./usr/share/securedrop-workstation-dom0-config/scripts/provision-all
│ │ -./usr/share/securedrop-workstation-dom0-config/scripts/rpmbuilder.Dockerfile
│ │  ./usr/share/securedrop-workstation-dom0-config/scripts/sdw-admin.py
│ │  ./usr/share/securedrop-workstation-dom0-config/scripts/validate_config.py
```
</details>

The handful of removed files were never required in a prod or staging dom0 installation, anyway. To double-check the results shown above, step through these commands on your local system:

```
# run these commands in your dev VM, e.g. sd-dev
make install-deps
mkdir -p /tmp/reprotest/{main,new}
git checkout main
make dom0-rpm
cp rpm-build/RPMS/noarch/securedrop-workstation-dom0-config-0.5.3-1.fc25.noarch.rpm /tmp/reprotest/main/
git checkout - # back to this branch
make dom0-rpm 
cp rpm-build/RPMS/noarch/securedrop-workstation-dom0-config-0.5.3-1.fc25.noarch.rpm /tmp/reprotest/new/
diffoscope --exclude-directory-metadata=recursive /tmp/reprotest/{main,new}/securedrop-workstation-dom0-config-0.5.3-1.fc25.noarch.rpm
```

The `--exclude-directory-metadata=recursive` flag will omit any diff output based in timestamps. Since the RPMs built on main are _not_ reproducible, every file in the package will show as differing from this branch's artifact, due at the very least to the timestamps. Excluding that info lets you focus on content. 

### Confirm it all works
Finally, let's make sure that despite the minimal cosmetic differences, all the tooling updates still result in a working system.
```
# run these commands in dom0
make clone
make clean
make dev
make test
```

Log in to the client, and perform some basic testing to ensure connectivity, DispVM access, etc.


rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -1,16 +1,28 @@
-Name:		securedrop-workstation-dom0-config
-Version:	0.5.3
+%global srcname securedrop-workstation-dom0-config
+%global version 0.5.3
+%global __python3 /usr/bin/python3
+%global python3_sitelib /usr/lib/%{_python_version}/site-packages
+# For reproducible builds
```
emkll, line 5: if there's a source or description for these changes in a comment or commit message, it may help future maintainers.

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -1,16 +1,28 @@
...
 URL:		https://github.com/freedomofpress/securedrop-workstation
-Source0:	securedrop-workstation-dom0-config-0.5.3.tar.gz
+Source0:	%{srcname}-%%{version}.tar.gz
 
 BuildArch:      noarch
-BuildRequires:	python3-setuptools
-BuildRequires:	python3-devel
+# Disable declaration of build dependencies, because
+# we build on Debian stable.
+#BuildRequires:	python3-setuptools
```
emkll, line 24: I am guessing this is why we add install-deps

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -84,13 +98,19 @@ install -m 644 launcher/sdw_notify/*.py %{buildroot}/opt/securedrop/launcher/sdw
 install -m 644 launcher/sdw_notify/*.py %{buildroot}/srv/salt/launcher/sdw_notify/
 install -m 644 launcher/sdw_util/*.py %{buildroot}/opt/securedrop/launcher/sdw_util/
 install -m 644 launcher/sdw_util/*.py %{buildroot}/srv/salt/launcher/sdw_util/
+install -m 644 config.json.example %{buildroot}/usr/share/%{srcname}/
+install -m 644 README.md LICENSE %{buildroot}/%{_custom_docdir}/
+find %{buildroot} -type d -iname '*.egg-info' -print0 | xargs -0 -r rm -rf
```
emkll, line 103: isn't this why we set `--skip-build` in setup.py for? can we replace this command by a `setup.py` parameter that would make it more maintainable?

scripts/build-dom0-rpm
```diff
@@ -1,67 +1,42 @@
...
-
-local_rpms="$(find rpm-build/ -type f -iname '*.rpm')"
-
-printf "\nRPM packages can be found at:\n\n%s\n" "$local_rpms"
+set -u
+set -o pipefail
+
+# Prepare tarball, rpmbuild will use it
+mkdir -p dist/
+rm -f dist/*.tar.gz
```
emkll, line 10: Are removing tarballs and historical RPMs strictly required here? This behavior may delete a developer's historical RPMs that they would not want to delete?

scripts/rpmbuilder.Dockerfile
```diff
@@ -1,35 +0,0 @@
-ARG CONTAINER_HASH
```
emkll, line 1: :wave:

scripts/build-dom0-rpm
```diff
@@ -1,67 +1,42 @@
...
+rm -f rpm-build/RPMS/noarch/*.rpm
+/usr/bin/python3 setup.py sdist
+
+# Use the epoch time of the highest semver tag available.
+# SOURCE_DATE_EPOCH="$(git tag | sort -V | tail -n 1 | xargs git log -1 --format=%at)"
+# Use the epoch time of the most recent commit. If works in dev,
+# as well as building from signed tags.
+SOURCE_DATE_EPOCH="$(git log -1 --format=%at HEAD)"
+export SOURCE_DATE_EPOCH
+
```
emkll, line 20: we can check for the presence of rpm build and other dependencies required here, and prompt the user to run `make install-deps` since the behavior is new here.

Makefile
```diff
@@ -21,14 +21,27 @@ dev staging: assert-dom0 ## Configures and builds a dev or staging environment
...
+
+.PHONY: reprotest-ci
+reprotest-ci: ## Runs reprotest for RPMs, with CI-compatible skips in variations
+	# Disable a few variations, to support CircleCI container environments.
+	# Requires a sed hack to reprotest, see .circle/config.yml
+	TERM=xterm-256color reprotest --variations "+all, +kernel, -domain_host, -fileordering" \
+		 -c "make dom0-rpm" . "rpm-build/RPMS/noarch/*.rpm"
+
+.PHONY: install-deps
+install-deps: ## Installs apt package dependencies, for building RPMs
```
emkll, line 41: This may make it more difficult for Fedora-based hosts to build the rpm with this change, since the logic here is Debian-only. Perhaps we can check the distro and conditionally apt or yum install the appropriate dependencies? This used to be enforced by `BuildRequires` in the rpm specfile below

setup.py
```diff
@@ -17,13 +17,12 @@
     license="GPLv3+",
     python_requires=">=3.5",
     url="https://github.com/freedomofpress/securdrop-workstation",
-    data_files=[("share/securedrop-workstation-dom0-config", ["config.json.example"])],
```
emkll, line 20: will the removal of the data_file here have any implication in the docs? Since the docs say we should copy the file, I don't think there's an impact: https://workstation.securedrop.org/en/stable/admin/install.html?highlight=config.json.example#configure-securedrop-workstation-estimated-wait-time-60-90-minutes

emkll (commented):
Thanks @conorsch for your great work here, this is very exciting. Preliminary testing suggests this works as expected in Debian Buster based dev vm, but I will do another round of testing (especially with a F32 dom0 to confirm no issues with python)

Very significant performance increases, and the removal of Docker is great:
- Docker build with no image in cache: 5 minutes
- Docker build with image in cache: 7 seconds
- Build using this logic: 2.5 seconds

Some comments inline for maintainability/usability purposes
Of note, README.md at Lines 144, 318 and 335 contain references to Docker that should probably be removed here as part of this PR.

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
conorsch: Good point, I'll amend the commit message with specifics. And the `buildflags` actually isn't required anymore, I'd documented that based on compared builds between F25 & F32, but rpms are reproducible without that particular override.

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
conorsch: Precisely. I believe we could technically pass in a macro here via the shell wrapper, based on distro, but that much indirection sounds confusing, at least at this stage in our pursuit of reproducible builds.

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
conorsch: I cannot find any python/setuptools invocation that will entirely skip creation of the "egg" directory (or zip archive). If you can, I would love to see it!

scripts/build-dom0-rpm
conorsch: Great point, will add!

Makefile
conorsch: According to https://github.com/freedomofpress/securedrop-workstation/issues/598, RPM builds on Fedora _already_ didn't work, so there's no loss of functionality here. If anything, these generalized changes make it a bit more straight forward to run under Fedora, by dynamically switching dependencies as you describe. Whether we even want to support Feodra as an additional build environment is an open question, in my mind, which is why I left #598 as "referenced" only in this PR.

setup.py
conorsch: No implication: the `config.json.example` file should be installed to the exact same location as already documented.

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
conorsch: Added comments inside the specfile. Ended up preserving optflags, too.

scripts/build-dom0-rpm
conorsch: Added.

scripts/build-dom0-rpm
conorsch: Removed the clean behavior.

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
conorsch: No changes made here—I'm still unable to skip the egg creation dir entirely, but open to suggestions. I suspect if it's possible it'd require a newer version of setuptools than what Debian Stable provides. Possible, but would prefer to defer for now.

conorsch:
All comments addressed, @emkll. Ready for re-review.

conorsch:
Pushed ( & squashed) a small change just now, touching up the docs and removing the now-unused dockerignore config:

```patch
diff --git a/.dockerignore b/.dockerignore
deleted file mode 100644
index fdc6683..0000000
--- a/.dockerignore
+++ /dev/null
@@ -1,4 +0,0 @@
-# ignore everything
-**
-# whitelist just the pubkey, used for verifying sigs
-!sd-workstation/apt-test-pubkey.asc
diff --git a/README.md b/README.md
index c40c3e8..332c9f8 100644
--- a/README.md
+++ b/README.md
@@ -332,10 +332,8 @@ https://github.com/freedomofpress/securedrop-debian-packaging/
 make dom0-rpm
\```

-This uses a base docker image as defined in https://github.com/freedomofpress/containers/.
-If you need to bump versions of the rpmbuild tooling, make an update to that
-repo's metadata, and increment the version as defined in the `Makefile`. See the
-`RPM_BUILD_VER` variable.
+The build assumes use of Debian Stable as the build environment. You can install
+the necessary dependencies from system packages via the `make install-deps` target.
```

emkll (approved):
Changes look great, thanks for updating the docs. I've tested the build again, reproducible across different machines. I noticed that it no longer generates a SourceRPM (we've never used it, so I don't think it's an issue here).

make clean/make clone/make dev/make test has all tests passing locally, with the sdw-admin scripts also working as expected. I think is is now ready for merge 



One last thing, after going through the build logic once again, I am no longer convinced this was a good suggestion https://github.com/freedomofpress/securedrop-workstation/pull/666#discussion_r596104617 . I'll let you make a final call before merging, but it's a minor detail.

scripts/build-dom0-rpm
```diff
@@ -1,67 +1,49 @@
 #!/bin/bash
-#
-#
-# Builds RPM for dom0 bootstrap logic
-# $1 - Digest Hash of base docker container to use
-
-set -u
+# Builds RPM for installation in dom0.
```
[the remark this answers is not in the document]
conorsch: Updated comment.

conorsch:
> I'll let you make a final call before merging, but it's a minor detail.

No sweat, I uncommented the clean action, so it's active again. Thanks for pointing out! Will merge after CI passes.

emkll (approved):
Thanks!

Issue #668, opened 2021-03-18
Reports of updater hanging without observable activity


eloquence:
We've received several reports of the preflight updater hanging and not recovering. I've personally seen it happen with the following STR:

1. Uninstall
2. Install 0.5.3 prod RPM and ensure config is set to `prod`
3. Run `sdw-admin --apply`
4. Run preflight updater (preferably in terminal so you see full output)

## Expected behavior

Updater will trigger full migration due to postinst logic added in https://github.com/freedomofpress/securedrop-workstation/pull/661/commits/6cf625ce3f02dbce0b7516e2059b6c345a82956c, but will eventually run to completion

## Actual behavior

Updater gets stuck at 35% with no observable activity. The issue is resolved on reboot, which is expected because `/tmp` will be empty and the migration will not be applied.

eloquence:
As part of investigating this issue, one immediate improvement we could make to the updater is to ensure that all output (not just the launcher's own log file) is written to disk for easier debugging.

emkll:
I have observed this issue locally while testing #666 , the updater will try to run a full apply run, but hangs at 35% after applying dom0 state. The reboot did resolve. I did observe that a task failed while initially applying dom0 state, prior to running the full provisioning run:
```
Mar 23 11:23:32 dom0 org.xfce.FileManager[5810]: ----------
Mar 23 11:23:32 dom0 org.xfce.FileManager[5810]:           ID: dom0-adjust-desktop-icon-size-xfce
Mar 23 11:23:32 dom0 org.xfce.FileManager[5810]:     Function: cmd.script
Mar 23 11:23:32 dom0 org.xfce.FileManager[5810]:         Name: salt://update-xfce-settings
Mar 23 11:23:32 dom0 org.xfce.FileManager[5810]:       Result: False
Mar 23 11:23:32 dom0 org.xfce.FileManager[5810]:      Comment: Command 'salt://update-xfce-settings' run
Mar 23 11:23:32 dom0 org.xfce.FileManager[5810]:      Started: 11:23:14.221777
Mar 23 11:23:32 dom0 org.xfce.FileManager[5810]:     Duration: 118.96 ms
Mar 23 11:23:32 dom0 org.xfce.FileManager[5810]:      Changes:
Mar 23 11:23:32 dom0 org.xfce.FileManager[5810]:               ----------
Mar 23 11:23:32 dom0 org.xfce.FileManager[5810]:               pid:
[...]

PR #684, opened 2021-04-02
Make DBUS session check more resilient

eloquence:
We should always ensure that the DBUS session is set to the logged in user, who should not be root.

Towards #668 

## Status

Ready for review

## Test plan

1. Configure your environment to `staging`
2. Check out this branch in `dom0` via `make clone`
3. Build a staging env via `make staging`
4. - [ ] Observe that `update-xfce-settings` runs successfully
5. Attempt to run the `update-xfce-settings` script as root
6. - [ ] Observe that you can't

### DIFF
--- a/dom0/update-xfce-settings
+++ b/dom0/update-xfce-settings
@@ -13,6 +13,11 @@ set -e
 set -u
 set -o pipefail
 
+if [[ $EUID -eq 0 ]]; then
+   echo "This script should not be run as root; it must be run as a user with an active login session."
+   exit 1
+fi
+
 TASK=${1:-none}
 ICONSIZE=64
 
@@ -23,10 +28,7 @@ fi
 
 # This script requires a valid DBUS session to work. When run non-interactively,
 # we assume that a sesssion is running for the current user.
-if [ -z "${DBUS_SESSION_BUS_ADDRESS:-}" ]; then
-   export DBUS_SESSION_BUS_ADDRESS="unix:path=/run/user/$(id -u)/bus"
-fi
-
+export DBUS_SESSION_BUS_ADDRESS="unix:path=/run/user/$(id -u $USER)/bus"
 
 if [[ $TASK == "disable-unsafe-power-management" ]]; then
   echo "update-xfce-settings: Disabling unsafe power management options for user $USER"


conorsch (approved):
Works as advertised. Still can't explain how uid 0 was running those tasks, but the error report was clear that it was happening.

Spent some time reading up on how dbus works, which has changed a bit in versions of systemd, so we may want to revisit this code during the Qubes 4.1 transition (when systemd should go from v231 to v245). I'll post a patch that I tested locally, but am not appending to the PR as part of the merge, since the changes presented work well as they are.

eloquence:
> but the error report was clear that it was happening.

Just noting that, if you look at @emkll's reported output,  you'll notice that $USER was correctly set to `m` (his Qubes username). My hypothesis is that `runas` was working just fine, but for some reason (use of `sudo` or something else) there was also a DBUS session for root, which the script as previously written did not override.

conorsch:
Tried this locally: 

```patch
diff --git a/dom0/update-xfce-settings b/dom0/update-xfce-settings
index f1adecf..4d64730 100755
--- a/dom0/update-xfce-settings
+++ b/dom0/update-xfce-settings
@@ -23,12 +23,15 @@ ICONSIZE=64
 
 if ! [ -x "$(command -v xfconf-query)" ]; then
   echo "Error: xfconf-query is not installed." >&2
-  exit 1
+  exit 2
 fi
 
-# This script requires a valid DBUS session to work. When run non-interactively,
-# we assume that a sesssion is running for the current user.
-export DBUS_SESSION_BUS_ADDRESS="unix:path=/run/user/$(id -u $USER)/bus"
+# This script requires a valid DBUS session to work. When run interactively,
+# DBUS_SESSION_BUS_ADDRESS should already be populated. Otherwise, create
+# a session so that the xfconf-query commands succeed.
+if [[ -z "${DBUS_SESSION_BUS_ADDRESS:-}" ]]; then
+    eval "$(dbus-launch)"
+fi
``` 

and it handled the edge case of DBUS_SESSION_BUS_ADDRESS not being defined. Which notably it isn't during a `make staging` run: if you `test -n $DBUS_SESSION_BUS_ADDRESS` in that script, it will fail. Whether we hardcode the path or politely request a session doesn't seem to make much of a difference. Once downside of politely requesting a fresh session via dbus-launch is that the session appears to hang around forever. In the case of the Workstation, we expect shutdowns almost daily, but I'm still not comfortable forking the dbus process on every update without identifying a cleanup mechanism.

PR #690, opened 2021-04-28
Add detailed logging to separate logfile

### LINKED ISSUE #688: Increase level of updater logging detail
The SecureDrop Workstation updater currently provides a basic level of logging: did updates/migrations succeed or fail? If they failed, finding the cause can be a bit tricky.

1. If the updater is launched from the desktop, you can find detailed output from a full migration (`sdw-admin --apply`) and from applying the `dom0` Salt state in the `journalctl` output for `org.xfce.FileManager`.
2. For updates of individual VMs, the `/var/log/qubes/mgmt-*` log files for the specific VM that is updated may include details about errors or packages that have been updated.

We should consider logging additional detail. A first proposal:
- Log all `apply_dom0` and `run_full_install` output in separate log files in `~/.securedrop_launcher/logs`.
- Log the full path to the management VM update log in `launcher.log` for individual VMs, to make it easier to locate.

eloquence:
## Status

Ready for review

## Description of Changes

Fixes #688

Makes our logging logic more flexible to configure multiple loggers that write to different logfiles, so `launcher.log` continues to be useful as a high level overview of all update events. Detailed information is written to `launcher-detail.log`, which is also rotated out after 10 copies.

To enable this additional logging, this PR switches the relevant subprocess calls from `check_call` to `check_output`. It distinguishes between success & error state, but logs in both cases.

## Testing

(Estimated wall time ~30 minutes, but mostly waiting.)

1. `make clone` this branch into `dom0` and `sudo dnf reinstall` (or `make staging` if you don't have a previously provisioned environment) the freshly built RPM from `rpm-build/RPMS/noarch`.
2. Remove or relocate the contents of `~/.securedrop_launcher/logs` so you can see the behavior with newly created log files.
3. In `dom0`, `mkdir /tmp/sdw-migrations && touch /tmp/sdw-migrations/potato`. This will force a full `sdw-admin --apply` run on the next update.
4. In `dom0`, run `/opt/securedrop/launcher/sdw-launcher.py --skip-delta 0`. This will force an updater run.
5. Look at the logs in `~/.securedrop_launcher/logs`
6. - [ ] Observe that `launcher-detail.log` contains output from `sudo qubesctl --show-output state.highstate` and from `sdw-admin --apply`
7. Modify the file `/srv/salt/update-xfce-settings` and insert an `echo XYZZY && exit 1` statement after the `set` statements. This is how we'll cause all Salt runs to error out.
8. Repeat steps 3 to 5.
9. - [ ] Observe that `launcher.log` contains error log entries pointing to `launcher-detail.log`
10. - [ ] Observe that `launcher-detail.log` contains error logs containing the full Salt output and the XYZZY magic word we had the script insert
11. Undo your modification from step 7 to restore the environment to a working state
  
## Checklist

### DIFF
--- a/launcher/sdw_util/Util.py
+++ b/launcher/sdw_util/Util.py
@@ -5,6 +5,7 @@ Utility functions used by both the launcher and notifier scripts
 import fcntl
 import os
 import logging
+import re
 import subprocess
 
 from logging.handlers import TimedRotatingFileHandler
@@ -27,12 +28,19 @@ LOCK_ERROR = "Error obtaining lock on '{}'. Process may already be running."
 # Format for those logs
 LOG_FORMAT = "%(asctime)s - %(name)s:%(lineno)d(%(funcName)s) " "%(levelname)s: %(message)s"
 
-sdlog = logging.getLogger(__name__)
+# Namespace for primary logger, additional namespaces should be defined by module user
+SD_LOGGER_PREFIX = "sd"
+
+sdlog = logging.getLogger(SD_LOGGER_PREFIX + "." + __name__)
 
 
-def configure_logging(log_file):
+def configure_logging(log_file, logger_namespace=SD_LOGGER_PREFIX, backup_count=0):
     """
     All logging related settings are set up by this function.
+
+    `log_file` - the filename
+    `logger_namespace - the prefix used for all log events by this logger
+    `backup_count` - if nonzero, at most backup_count files will be kept
     """
 
     if not os.path.exists(LOG_DIRECTORY):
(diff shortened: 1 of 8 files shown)


launcher/sdw_util/Util.py
```diff
@@ -163,3 +173,17 @@ def get_qt_version():
...
+        return logging.getLogger(prefix)
+    else:
+        return logging.getLogger(prefix + "." + module)
+
+
+def strip_ansi_colors(str):
+    """
+    Strip ANSI colors from command output
+    """
+    return re.sub(r"\u001b\[.*?[@-~]", "", str)
```
eloquence, line 189: Open to a more robust solution here, of course, just don't want to write color codes to the log file.

conorsch (approved):
This worked quite well! Big improvement to have these details at hand, particularly after a failed run.

PR #704, opened 2021-06-07
Update PULL_REQUEST_TEMPLATE.md

creviera:
Signed-off-by: Allie Crevier <allie@freedom.press>

## Status

Ready for review

## Description of Changes

Updated our PR template to include a deployment section so that we consider how changes might affect both existing and fresh installs when applied. Also a reorg and some new bullets. I also moved the part about adding a test plan to the testing section.


conorsch (approved):
Very nice, thanks for taking the time, @creviera

Issue #705, opened 2021-06-08
DispVM functionality is broken


conorsch:
When opening files in the SecureDrop Client recently, I observed DispVMs failing to display documents.

### Steps to reproduce
1. Run a recent version of the SecureDrop Workstation. Observed on both staging and prod 0.5.4.
2. Open SecureDrop Client, download a submission.
3. Click "open" to view the document.
4. Observe "dispXXX is starting..." notification
5. Observe "dispXXX has halted" notification, with no document ever having been displayed

### Expected results
DispVM displays the document requested, and remains open until I close the window

### Actual results
DispVM never displays the document, and closes immediately.

### More info
Reported earlier in https://github.com/freedomofpress/qubes-template-securedrop-workstation/issues/20#issuecomment-857159277, but warrants its own issue. I suspect that mimetype associations are to blame, will start investigating logs.

eloquence:
I can repro on prod. If I attempt to open a file like so in `sd-app`:
`qvm-open-in-vm --view-only @dispvm:sd-viewer somefile.png`

I observe the following `qrexec` lines in `journalctl`:

```
Jun 08 15:01:45 dom0 qrexec[5963]: qubes.OpenInVM: sd-app -> @dispvm:sd-viewer: allowed to @dispvm:sd-viewer
Jun 08 15:02:13 dom0 qrexec[6471]: qubes.GetDate: disp7607 -> @default: allowed to dom0
Jun 08 15:02:14 dom0 qrexec[6491]: securedrop.Log: disp7607 -> sd-log: allowed to sd-log
Jun 08 15:02:15 dom0 qrexec[6493]: qubes.WindowIconUpdater: disp7607 -> dom0: allowed to dom0
Jun 08 15:02:16 dom0 qrexec[6503]: qubes.OpenInVM: disp7607 -> @dispvm:sd-viewer: denied: denied by policy /etc/qubes-rpc/policy/qubes.OpenInVM:6
Jun 08 15:02:25 dom0 qrexec[6493]: qubes.WindowIconUpdater: disp7607 -> dom0: denied: qrexec-client failed
```

creviera:
also dispvm logs (not super useful but including to help us get a complete picture):
```
user@sd-log:~/QubesIncomingLogs$ tail disp8832/syslog.log 
Jun  8 17:59:54 localhost rtkit-daemon[883]: Failed to look up client: No such file or directory
[...]

PR #707, opened 2021-06-09
Use correct VM name for mimetype handling

conorsch:
## Status

Ready for review 

## Description of Changes

Intended to address #705. 

We've observed inconsistent symlinking based on "grains.id", which resolve to the VM hostname in all cases. Let's instead call out to `qubesdb-read /name`, which sidesteps all Salt machinery, and polls for the VM name via the configuration API.

## Testing

1. Set up a _prod_ install, fully updated
2. Make sure you can reproduce a failing DispVM setup, as described in #705. For example, run `ln -sf /opt/sdw/mimeapps.list.sd-default ~/.local/share/applications/mimeapps.list` inside `sd-viewer` to break it. You should confirm inside SecureDrop Client that DispVMs fail to open submissions.
3. Run `make clone` to fetch these changes into dom0
4. Manually install the freshly fetched rpm package with `sudo dnf reinstall <rpm>`
5. Run `sdw-admin --apply` to do a full state run. 
6. Re-test the DispVM setup. Is it fixed?

### DIFF
--- a/dom0/sd-mime-handling.sls
+++ b/dom0/sd-mime-handling.sls
@@ -20,7 +20,8 @@ sd-private-volume-mimeapps-config-dir:
     - makedirs: True
     - mode: "0755"
 
-{% if grains['id'] in ["sd-viewer", "sd-app", "sd-devices-dvm"] %}
+{% set vm_name = salt['cmd.shell']('qubesdb-read /name') %}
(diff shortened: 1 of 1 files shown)


creviera (approved):
this fixes the DispVM issue (just needed to reboot sd-viewer after doing a full state run)

PR #708, opened 2021-06-09
Bump version to 0.5.5

conorsch:
## Status

Ready for review

## Description of Changes
Releasing the changes merged in [0]. Intentionally
preserving the full-migration flag in postinst to trigger another
provisioning run, in order to apply this fix for those who need it.

[0] https://github.com/freedomofpress/securedrop-workstation/pull/707/files

## Testing
Visual review is sufficient.


conorsch:
Not just a release bump: I tacked on a clarifying commit as a follow-up to #707, as discussed in standup today.

eloquence:
I've run `sdw-admin --apply` with these changes and the symlink was set to the correct target. :tada:

eloquence:
I've also tested without the fixup commit 4027ae48357f0b6d73773dab01f7b741e457a19b, and that resulted in a symlink to `/opt/sdw/mimeapps.list.sd-large-buster-template`, which does not exist. So now we know what grains["id"] actually resolves to in this situation: `sd-large-buster-template`.

creviera (approved):
https://github.com/freedomofpress/securedrop-workstation/commit/4027ae48357f0b6d73773dab01f7b741e457a19b looks good and erik has confirmed the fix

Issue #653, opened 2021-02-04
Automate diskspace maintenance tasks


eloquence:
SecureDrop Workstation templates can accumulate a large amount of used diskspace. We should consider if we can run certain maintenance tasks automatically as part of update runs. For example, one one system, just running `sudo apt autoremove` freed up ~3.5 GB of system space (in 3 different templates, each). With a default size of 10GB, that can be the difference between a working template or a broken one.

eloquence:
@marmarek Disk space inflation inside templates is biting both developers and end users on a regular basis; as noted above, `autoremove` can often free up several GBs of space. While we can of course add our own `autoremove` runs to the SecureDrop Workstation updater, it seems better resolved at the OS level. Are there flags we should set during updates using `update.qubes-vm`  with `qubesctl`, or improvements coming in Qubes 4.1? If not, perhaps this is something that we could usefully contribute to upstream?

marmarek:
This is a very good point indeed. I think it should be easy enough to add autoremove call to `update.qubes-vm` salt file. I don't see proper abstraction for it in the `pkg` state, so it will  likely need `cmd.run` state.

conorsch:
Observed this in the wild today, recommended as a workaround: `qvm-volume resize sd-large-buster-template:root 15G`.

conorsch:
And another one! Same resolution. We should definitely set 15G by default as part of the dom0 state—similar to what we already do for the sd-app private volume.

Anecdotally, @creviera ran `apt autoremove` tasks in her long-running large template, and it cleaned up >3GB (!) of space.

eloquence:
@conorsch @creviera I'm wondering if we can do more in terms of how we ship the grsec package(s) and declare their dependencies to clean up old kernels after each update. At least in anecdotal observations, most of the `autoremove` diskspace cleanup has been due to old kernel packages that are no longer needed still hanging around.

conorsch:
> At least in anecdotal observations, most of the autoremove diskspace cleanup has been due to old kernel packages that are no longer needed still hanging around.

See also #442

zenmonkeykstop:
Flagging for @deeplow - if this functionality is available in the new Qubes updater we can close this, if not it might make sense to raise it upstream again. Fallback would be to do it ourselves after the updater run finishes.

marmarek:
`apt-get dist-upgrade`

[...]

Issue #442, opened 2020-02-06
Remove old kernels from long-running templates


emkll:
Because `linux-headers` and `linux-image` packages are pulled in via apt dependency (`securedrop-workstation-grsec` metapackage), when the metapackage, headers and image are updated, the previous versions will remain installed. we should run autoremove as part of the update process (maybe the launcher?) to ensure we don't exhaust system storage. 

A single kernel/header version takes ~650MB on disk (system storage default is 10GB, with ~50% used)

conorsch:
Should be resolved by https://github.com/freedomofpress/securedrop-workstation/pull/722, specifically https://github.com/freedomofpress/securedrop-workstation/pull/722/files#diff-c53f24489d05b8fbea9800086d050eb3ae5afa8742c864c0f655f8a78870a017R24-R28  Not marking as closed until we observe that change cleaning up historical kernel packages on a staging instance.

legoktm:
Indeed, this was fixed back in 2021 in https://github.com/freedomofpress/securedrop-workstation/commit/46abb0b68136b0f0ba9c953e91360085f39be428 which added `apt-get autoremove -y` to the update process.

PR #722, opened 2021-08-24
Honor oldstable repo config for Debian 10 Buster

### LINKED ISSUE #572: fpf-apt-test-repo.sls state name is misleading
The `fpf-apt-test-repo.sls` Salt state sets up the test _or_ the prod repo, depending on the configuration. The name is misleading, which can be confusing when inspecting logs. We should pick a name that accurately reflects the purpose of the state.

### LINKED ISSUE #651: Configure FPF apt repo early in provisioning
Right now, the FPF apt repo—whether `apt-test` or `apt`—is configured _after_ we run the upstream Qubes salt state for upgrading all packages within a VM: 

https://github.com/freedomofpress/securedrop-workstation/blob/9cb41a836e25112f9b3199d70c38d0971da779fc/dom0/fpf-apt-test-repo.sls#L10-L11

Let's switch that logic, so the apt repo is configured first, and the full package upgrade at install time includes the FPF apt repo. A bit more indirection is likely the solution, so that a new state file includes both the upstream logic and the FPF apt repo logic. Fine opportunity to resolve #572, as well.

### LINKED ISSUE #721: Templates fail to update due to buster "oldstable" status
During a clean install in support of #717, I encountered this failure:

```
  ----------
            ID: update
      Function: pkg.uptodate
        Result: False
       Comment: E: Repository 'https://deb.debian.org/debian buster InRelease' changed its 'Suite' value from 'stable' to 'oldstable'
                E: Repository 'https://deb.debian.org/debian-security buster/updates InRelease' changed its 'Suite' value from 'stable' to 'oldstable'
       Started: 09:47:31.090662
      Duration: 5025.744 ms
       Changes:   
  ----------
```

That same error will bite us on the updater runs, blocking upgrades until resolved. 

There's an option we can use to resolve: `apt-get --allow-releaseinfo-change update` but we'll need to make sure we can run that prior to any other apt commands.

conorsch:
## Status

Ready for review

## Description of Changes

Unbreaks clean installs by respecting the renaming of Debian Buster apt sources to "oldstable" by upstream Debian. Additionally, cleans up some of the apt repo config to resolve some long-standing issues. 

* Closes #721
* Closes #651
* Closes #572 

## Testing

It's critical to remove the TemplateVM RPM prior to confirming resolution. That's the "dnf remove" action below. Because the template RPM will need to be re-downloaded, expect ~600MB download to dom0. 

```
make clone && make clean
sudo dnf remove qubes-template-securedrop-workstation-buster # necessary because make clean doesn't remove TemplateVM
sudo dnf clean all
make dev
```
If you see no errors and the provisiong completes successfully, then the issue is resolved. 

## Deployment

So far, testing shows that the updater runs are not affected, so existing installs do not appear to be at risk. I don't quite understand why that is, unfortunately, other than the age of the apt lists in use.

### DIFF
--- a/dom0/fpf-apt-repo.sls
+++ b/dom0/fpf-apt-repo.sls
@@ -2,16 +2,30 @@
 # vim: set syntax=yaml ts=2 sw=2 sts=2 et :
 #
 
-# Import the Qubes-maintained Salt logic for upgrading VM packages.
-# Intelligently handles both Debian & Fedora VMs. For reference, see:
+# Don't start with the Qubes-maintained Salt logic for upgrading VM packages:
 #
 #   dom0:/srv/formulas/base/update-formula/update/qubes-vm.sls
 #
-include:
-  - update.qubes-vm
+# We want to make sure that certain maintenance tasks like cleaning out
+# old packages and updating apt lists are handled first, otherwise
+# the subsequent tasks will fail. For reference
+# include:
+#  - update.qubes-vm
 #  - sd-default-config
 
 {% from 'sd-default-config.sls' import sdvars with context %}
+
+# Debian Buster was changed from 'stable' to 'oldstable' on 2021-08,
+# so honor that change in the apt sources.
+update-apt-cache-with-stable-change:
+  cmd.run:
+    - name: apt-get update --allow-releaseinfo-change
+
+autoremove-old-packages:
+  cmd.run:
+    - name: apt-get autoremove -y
+    - require:
+      - cmd: update-apt-cache-with-stable-change
 
 # That's right, we need to install a package in order to
 # configure a repo to install another package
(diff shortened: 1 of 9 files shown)


creviera:
Sorry, was out sick and just got back to this. I lost the

> Completed another run without errors (I did hit an issue with `make clean` due to `rsyslog` being masked - looks like this is both interfering with logging, which you added a card for @conorsch, and at least sometimes with cleanup as well).

I created an issue with a workaround to capture this: https://github.com/freedomofpress/securedrop-workstation/issues/723

Second time around this worked as expected.

creviera (approved):

conorsch:
> The change to running the fpf-apt-repo state in the updater is significant; it looks like this will also help towards #653 since it will run autoremove on each updater run, correct?

Yes, that's the intention! There's also #442, which I haven't closed, pending confirmation of the new behavior on a staging setup.

PR #734, opened 2021-10-28
Remove misleading Docker requirement

eloquence:
## Status

Ready for review 
## Description of Changes

Docker _was_ required to build RPMs until #666. It no longer is. While it can be useful if folks want to use the Docker-based _server_ environment, said environment is not actually helpful in getting a SecureDrop Workstation up and running (you need an `.onion` for that), which is what this README is focused on. So, on balance, it seems most prudent to simply remove this sentence to avoid confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ The development environment is just like the [Staging Environment](#staging-envi
 
 #### Download, Configure, Copy to `dom0`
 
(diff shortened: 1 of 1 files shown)


conorsch (approved):
Thanks! And hat tip to @gonzalo-bulnes, too, who's been very helpful in bringing fresh eyes to bear on the dev docs throughout SD.

PR #739, opened 2021-11-18
update deps and add virtual env makefile targets

creviera:
# Description

Support more developers who want to contribute to the Updater codebase by adding virtual environment helpers, a test helper, updating our dev dependencies to support more environments, and updating the readme documentation.

# Test  

### Running the updater
Follow the "Running the Updater" instructions added to the readme here: https://github.com/freedomofpress/securedrop-workstation/blob/40c6141b13351ff5edd414207d35f1c0a9f4f1fd/launcher/README.md#running-the-updater
- [ ] Confirm you can launch the preflight updater following the instructions
- [ ] Confirm you can open the notifier dialog following the instructions
   * make sure to update `~/.securedrop_launcher/sdw-last-updated` so that it's been at least 5 days since  the last updater run and that you didn't just reboot your system -- this test needs to run at least 30 minutes after reboot)

### Developer environment
Follow the "Developer environment" instructions added to the readme here: https://github.com/freedomofpress/securedrop-workstation/blob/40c6141b13351ff5edd414207d35f1c0a9f4f1fd/launcher/README.md#developer-environment
- [ ] Confirm you can run both the updater and launcher in the pyqt5 virtual environment
- [ ] Confirm you can run both the updater and launcher in the pyqt4 virtual environment
- [ ] Confirm you can run `make check` in the pyqt5 virtual environment
- [ ] Confirm you can run `make check` in the pyqt4 virtual environment


creviera:
Mostly want eyes on: https://github.com/freedomofpress/securedrop-workstation/blob/be834403b45ba996ee3c250d360ec2c2709a15a0/launcher/README.md 

It seems like we could do away with the env var to set which version of pyqt we're using if we use `--site-system-packages` (see Makefile changes).

launcher/README.md
```diff
@@ -1,24 +1,40 @@
...
-fedora-25 only includes PyQt4, which is why we have to support it for now.
-The next version of Qubes, Qubes 4.1, will include PyQt5 in dom0.
+To run the preflight updater:
+1. Open a `dom0` terminal
+2. Run `/opt/securedrop/launcher/sdw-launcher.py --skip-delta 0`
 
-## Installing PyQt4
+To run the notifier that pops up if `/proc/uptime` (how long the system has been on since its last restart) is greater than 30 seconds and `~/.securedrop_launcher/sdw-last-updated` (how long it's been since the Updater last ran) is greater than 5 days:
+1. Open a `dom0` terminal
+2. Run `/opt/securedrop/launcher/sdw-notify.py --skip-delta 0`
```
eloquence, line 15: `sdw-notify.py` does not currently support the `--skip-delta` argument. We could add that, or perhaps a more straightforward `--force` or similar.

launcher/README.md
```diff
@@ -1,24 +1,40 @@
...
-Qubes 4.0.3 uses an end-of-life Fedora template in dom0 (fedora-25). See
-rationale here:
+## Running the Updater
 
-https://www.qubes-os.org/doc/supported-versions/#note-on-dom0-and-eol
+Qubes 4.0.3 uses an end-of-life Fedora template in dom0 (fedora-25). See rationale here: https://www.qubes-os.org/doc/supported-releases/#note-on-dom0-and-eol.
 
-fedora-25 only includes PyQt4, which is why we have to support it for now.
-The next version of Qubes, Qubes 4.1, will include PyQt5 in dom0.
+To run the preflight updater:
```
eloquence, line 9: It may be useful to explicitly say "on a Debian Buster system" since those instructions won't work e.g. on Ubuntu 20.04 (no more PyQt4 packages :/).

launcher/README.md
```diff
@@ -1,24 +1,40 @@
...
+The next version of Qubes will include PyQt5 in dom0, which is why we also support PyQt5.
 
-The recommended version of PyQt5 is included in the developer requirements
-for this project, which you can install via:
+### PyQt4 instructions
 
-pip install --require-hashes -r dev-requirements.txt
+To run the preflight updater outside of `dom0`:
+1. `cd securedrop-workstation/launcher`
+2. Make the launcher script executable: `chmod +x sdw-launcher.py`
```
eloquence, line 25: `git` lets us version that executable bit, which was already the case for `sdw-notify.py`. I can't think of a reason not to also set it for `sdw-launcher.py`, so I pushed up a small commit to do so. That should make this step unnecessary.

launcher/README.md
```diff
@@ -1,24 +1,40 @@
...
+4. Set up your virtual environment by running `make venv-pyqt4 && source .venv-pyqt4/bin/activate`
+4. `export SDW_UPDATER_QT=4` (in case it was set to `5` when testing against PyQt5)
+5. Now you can run the updater: `./sdw-launcher.py` (it won't actually update VMs unless you are in `dom0`)
+6. You can also run the notifier: `./sdw-notify.py`
+
+### PyQt5 instructions
+
+To run the preflight updater outside of `dom0`:
+1. `cd securedrop-workstation/launcher`
+2. Make the launcher script executable: `chmod +x sdw-launcher.py`
```
eloquence, line 36: (See above.)

eloquence:
This looks great @creviera; I've tested the new venv in a Debian Buster VM and it works as expected. :)

I pushed up a couple of small tweaks. Please note the comment re: `sdw-notify.py`; we'll either have to omit that example from the README, or add support for the argument.

creviera:
I still need to respond to review feedback- just taking this out of draft mode.

launcher/README.md
creviera: Since the step below says to open a `dom0` terminal, I could say instead, open a `Qubes 4.0.4 dom0` terminal. That should cover it.

launcher/README.md
creviera: yeah, not sure I want to propagate the term `skip-delta`. switching to `force` would be less confusing to developers. i'll remove it from the docs here to address this for now.

launcher/README.md
eloquence: Sorry, I commented on the wrong line - I meant to respond to the instructions "To run the preflighter outside dom0" later in the doc, specifically re: PyQt4 - in that context I think it might be useful to specify that a Debian Buster system is required.

creviera: Actually, we already mention how Qubes 4.0.3 (updating it now to 4.0.4) and that it uses a fedora-25 template, so i think we're good if i add back the bit about pyqt4 being the latest version of pyqt that we can install on fedora-25.

creviera:
PR review comments addressed, this is ready for a final review.

conorsch:
Whoops, missed this one. @rocodes will review; ping with any questions, @rocodes, and I can merge after you've taken a look!

rocodes:
Instructions work well, both venvs build as expected, and I can launch the preflight updater and the notify dialog in both PyQt4 and PyQt5, however `make check` has failing tests in each environment:

PyQT4:

```
=================================================== short test summary info ====================================================
FAILED tests/test_util.py::test_pick_qt[os-release-qubes-4.1-5-None-None] - assert 4 == 5
================================================ 1 failed, 144 passed in 0.49s =================================================
make: *** [Makefile:33: test] Error 1
(.venv-pyqt4) user@sd-dev:~/projects/securedrop-workstation/launcher$ echo $SDW_UPDATER_QT
4
``` 


PyQT5:

```
=================================================== short test summary info ====================================================
FAILED tests/test_util.py::test_pick_qt[os-release-qubes-4.0-4-None-None] - assert 5 == 4
FAILED tests/test_util.py::test_pick_qt[os-release-ubuntu-4-None-None] - assert 5 == 4
FAILED tests/test_util.py::test_pick_qt[no-such-file-4-None-None] - assert 5 == 4
FAILED tests/test_util.py::test_pick_qt[bad-os-release-file-4-None-None] - assert 5 == 4
================================================ 4 failed, 141 passed in 0.49s =================================================
make: *** [Makefile:33: test] Error 1
(.venv) user@sd-dev:~/projects/securedrop-workstation/launcher$ echo $SDW_UPDATER_QT
5
```

creviera:
Good find @rocodes. Looks like an issue with the test logic that will need some digging into. If you run `unset SDW_UPDATER_QT` before running `make test` all tests pass. But when the environment variable is set, whether or not it's the correct value, I see what you see above, e.g.:

```
export SDW_UPDATER_QT=5
(.venv) creviera@debian:~/workspace/freedomofpress/securedrop-workstation/launcher(add-venv-target○) » ./sdw-launcher.py      
Using Qt5 (experimental)
(.venv) creviera@debian:~/workspace/freedomofpress/securedrop-workstation/launcher(add-venv-target○) » make check  
...
======================================================= short test summary info ========================================================
FAILED tests/test_util.py::test_pick_qt[os-release-qubes-4.0-4-None-None] - assert 5 == 4
FAILED tests/test_util.py::test_pick_qt[os-release-ubuntu-4-None-None] - assert 5 == 4
FAILED tests/test_util.py::test_pick_qt[no-such-file-4-None-None] - assert 5 == 4
FAILED tests/test_util.py::test_pick_qt[bad-os-release-file-4-None-None] - assert 5 == 4
==================================================== 4 failed, 141 passed in 0.65s =====================================================
make: *** [Makefile:33: test] Error 1
```

It seems like the intention is to ensure that, when the environment variable is set, that the correct version of Qt is used. This doesn't work on `main` either, right? @eloquence what that the intention?

eloquence:
The failing tests don't mock the environment variable (they should); this hasn't become apparent previously because we've relied on no environment variable being set when running these tests. @rocodes Would you like to take a stab at fixing up the tests? There's existing mocking in place in other tests (`os.environ`) that you can use as an example.

rocodes:
After unsetting the environment vars and running `make check` from the `launcher` directory, it looks like black wants to check the .venv and .venv-pyqt4 directories as well. Should we use the `--exclude` directive to avoid this? (Black will default exclude what's in `.gitignore`, but since this is a subfolder and not the root project directory, we don't benefit from that).

Another small thing is that I noticed that black in the root project Makefile does not include the `--line-length=100` parameter, but in the launcher directory, it does. I think @creviera you mentioned something about this, I'm just wondering if we want these two to be the same.

creviera:
> it looks like black wants to check the .venv and .venv-pyqt4 directories as well. Should we use the --exclude directive to avoid this? (Black will default exclude what's in .gitignore, but since this is a subfolder and not the root project directory, we don't benefit from that).

looks like i have this change locally that i just forgot to push, and i totally agree about not using .gitignore here... pushing the commit now

creviera:
> Another small thing is that I noticed that black in the root project Makefile does not include the --line-length=100 parameter, but in the launcher directory, it does. I think @creviera you mentioned something about this, I'm just wondering if we want these two to be the same.

good point, i'll do that. also, i can do what erik suggests [here](https://github.com/freedomofpress/securedrop-workstation/pull/739#issuecomment-988161381) to fix up that test.

creviera:
Thanks, Erik, for fixing up that test. Looks like the parent project already has a pyproject.toml that sets the black tool line length to 100, so all PR comments have been addressed at this point. @rocodes ready for another round

rocodes (approved):
Stepped though again and everything LGVTM! (Note that I don't have merge rights so we'll still need someone to do the honours here)

conorsch:
Thanks for testing, @rocodes! Approving and merging based on your positive review.

conorsch (approved):

PR #740, opened 2021-11-29
Add option to launch updater from sdw-notify script

### LINKED ISSUE #595: Add "Remind me later", "Update now" options to sdw-notify.py
The notifier script currently just shows a warning/reminder that it's time to update, and will come back on the dot every hour if you defer the update. It would be good to:

1. Give the user the option to launch the updater from the notification
2. Give the user the option to silence the notification for a limited time (e.g., 8 hours).

The "Update now" option should detect if `sd-app` is currently running and, if so, show a warning that it will be shut down as part of the update.

We should decide on the full UX before implementation. Many UIs do this with "Remind me later", "Update now" type actions (e.g., the [Ubuntu updater](https://149366088.v2.pressablecdn.com/wp-content/uploads/2013/04/software-updater.jpg)).

rocodes:
## Status

Ready for review

## Description of Changes

Fixes #595 

Move Notify dialog into separate file and allow user to launch Preflight updater directly from the Notify dialog. 

Also separates out constants from Notify.py and Updater.py into a constants file (`sdw_util/Const.py`). Towards goals of a) reducing mocking of class or module internals during testing, b) reducing dependency surface/moving towards separation of UI and logic components. I've refactored Notify.py but not Updater.py to use the constants file. If this approach works for people then in a subsequent PR I'll also make some similar changes to Updater.py. 

Misc: 
- I've stuck to the [edit for clarity] camel case naming convention going on here (NotifyApp.py), but I'd be happy to switch to more typical Python naming conventions, possibly in another PR so as to keep things clean. 

- As discussed in sdw developer hangout, a separate PR will be made to begin adding UI tests for all the launcher/notifier components.

### Changes proposed in this pull request:

- Replace simple notification ("ok") with option to launch updater from sdw-notify script, or to defer reminder (deferral period is until the next time the cron job runs)
- Change cron job that runs notifier to run every 2 hours instead of hourly  

## Testing

Check out this branch in your sd dev vm.

**Manual testing**
Copy this branch to dom0 via `make clone`, then `make dev`. 

Ensure `sd-app` VM is shut down. 
Edit the file in `dom0` called `~/.securedrop_launcher/sdw-last-updated` to change the date to a date >= 5 days ago.
Run `python3 /opt/securedrop/launcher/sdw-notify.py`:
- [ ] Notification appears with prompt to check for updates. 
- [ ] No "**Warning:**" text is present in the notification
- [ ] The default selected button option is "Check for updates".
- [ ] Clicking "Remind me later" (or pressing the Esc key) closes the dialog window. Note the time.

Start `sd-app` vm via `qvm-start sd-app` in `dom0`. Wait for the notify prompt to reappear.  
- [ ] The notify prompt appears again after a delay of t < 2 hours (The reminder runs every 2 hours on the even hours. The delay will depend on what time the user dismissed the dialog). 
- [ ] This time, the "**Warning:**" text is present in the notification
- [ ] Clicking "Check for updates" launches the Preflight Updater.  
- [ ] The Preflight Updater can run updates successfully. 
- [ ] No GTK warnings or other errors are generated or printed to the `dom0` console when the Preflight Updater is launched.

**Automated testing**
- [ ] All tests pass

## Deployment

2. New installs

### DIFF
--- a/launcher/sdw-notify.py
+++ b/launcher/sdw-notify.py
@@ -4,18 +4,19 @@ Displays a warning to the user if the workstation has been running continuously
 for too long without checking for security updates. Writes output to a logfile,
 not stdout. All settings are in Notify utility module.
 """
-
 import sys
-
-from sdw_notify import Notify
-from sdw_updater_gui import Updater
+from sdw_notify import Notify, NotifyApp
+from sdw_updater_gui import Updater, UpdaterApp
 from sdw_util import Util
 
 if Util.get_qt_version() == 5:
     print("Using Qt5 (experimental)")
-    from PyQt5.QtWidgets import QApplication, QMessageBox
+    from PyQt5.QtWidgets import QApplication
 else:
-    from PyQt4.QtGui import QApplication, QMessageBox
+    from PyQt4.QtGui import QApplication
+
+Util.configure_logging(Notify.LOG_FILE)
+log = Util.get_logger(Notify.LOG_FILE)
 
 
 def main():
@@ -24,8 +25,6 @@ def main():
     the preflight updater is running, and certain checks suggest that the
     system has not been updated for a specified period
     """
-
-    Util.configure_logging(Notify.LOG_FILE)
 
     if Util.is_conflicting_process_running(Notify.CONFLICTING_PROCESSES):
         # Conflicting system process may be running in dom0. Logged.
@@ -52,19 +51,41 @@ def main():
 def show_update_warning():
     """
     Show a graphical warning reminding the user to check for security updates
-    using the preflight updater.
-    """
-    app = QApplication([])  # noqa: F841
+    using the Preflight Updater.
 
-    QMessageBox.warning(
-        None,
-        "Security check recommended",
-        "This computer has not been checked for security updates recently. "
-        "We recommend that you launch or restart the SecureDrop app to "
-        "check for security updates.",
-        QMessageBox.Ok,
-        QMessageBox.Ok,
-    )
+    If the user opts to check for updates, launch the Preflight Updater.
+    If the user opts to defer, they will be reminded again the next time the
+    notify script runs.
+    """
+
+    app = QApplication([])  # noqa: F841
+    dialog = NotifyApp.NotifyDialog(Util.is_sdapp_halted())
+    result = dialog.run()
+
+    # Check results of Notify Dialog and launch the Preflight Updater if user
+    # has opted to check for updates.
+    if result == NotifyApp.NotifyStatus.CHECK_UPDATES:
+        log.info("Launching Preflight Updater")
+        updater = UpdaterApp.UpdaterApp()
+        updater.show()
+        sys.exit(app.exec_())
+    elif result == NotifyApp.NotifyStatus.DEFER_UPDATES:
+        # Currently, `DEFER_UPDATES` is a no-op, because the deferral period is
+        # simply the period before the next run of the notify script (defined in
+        # `securedrop-workstation/dom0/sd-dom0-crontab.sls`).
+        log.info(
+            "User has deferred update check. sdw-notify will run "
+            "again at the next scheduled interval."
+        )
+        sys.exit(0)
+    else:
+        # NotifyApp.NotifyStatus.ERROR_UNKNOWN, meaning the dialog returned an
+        # unexpected state. The error is logged in NotifyDialog.run().
+        log.info(
+            "Unexpected result from NotifyDialog. sdw-notify will run "
+            "again at the next scheduled interval."
+        )
+        sys.exit(result)
 
 
 if __name__ == "__main__":
--- a/launcher/sdw_notify/NotifyApp.py
+++ b/launcher/sdw_notify/NotifyApp.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+"""
+Notification dialog that appears when user has not applied security updates
+recently. Prompts user to check for updates or defer reminder.
+"""
+from enum import Enum
+from sdw_util import Util
+from sdw_notify import Notify, strings
+
+if Util.get_qt_version() == 5:
+    print("Using Qt5 (experimental)")
+    from PyQt5.QtWidgets import QMessageBox
+else:
+    from PyQt4.QtGui import QMessageBox
+
+sdlog = Util.get_logger(Notify.LOG_FILE)
+
+
+class NotifyStatus(Enum):
+    """
+    Supported exit states from Notify dialog.
+    """
+
+    CHECK_UPDATES = "0"
+    DEFER_UPDATES = "1"
+    ERROR_UNKNOWN = "2"
+
+
+class NotifyDialog(QMessageBox):
+    """
+    Shows notification advising user that they have not checked for updates
+    recently, and offering option to check now or defer the check.
+
+    Constructor takes a boolean parameter, `is_sdapp_stopped`, which determines
+    whether a longer error message indicating the updater's impact on a
+    currently-running client session will be shown.
+    """
+
+    def __init__(self, is_sdapp_stopped: bool):
+        super().__init__()
+        self._is_sdapp_stopped = is_sdapp_stopped
+        self._ui()
+
+    def _ui(self):
+        self.setWindowTitle(strings.headline_notify_updates)
+        self.setIcon(QMessageBox.Warning)
+        self.setStandardButtons(QMessageBox.No | QMessageBox.Ok)
+        self.setDefaultButton(QMessageBox.Ok)
+        self.setEscapeButton(QMessageBox.No)
+        button_check_now = self.button(QMessageBox.Ok)
+        button_check_now.setText(strings.button_check_for_updates)
+        button_defer = self.button(QMessageBox.No)
+        button_defer.setText(strings.button_defer_check)
+
+        if self._is_sdapp_stopped:
+            self.setText(strings.description_notify_updates)
+        else:
+            self.setText(strings.description_notify_updates_sdapp_running)
+
+    def run(self) -> NotifyStatus:
+        """
+        Launch dialog and return user selection.
+        """
+        result = self.exec_()
+
+        if result == QMessageBox.Ok:
+            sdlog.info(
+                "NotfyDialog returned {}, user has opted to check for updates".format(result)
+            )
+            return NotifyStatus.CHECK_UPDATES
+        elif result == QMessageBox.No:
+            sdlog.info("NotfyDialog returned {}, user has opted to defer updates".format(result))
+            return NotifyStatus.DEFER_UPDATES
+        else:
+            # Should not occur, as this dialog which can only return
+            # one of two states.
+            sdlog.error(
+                "NotfyDialog returned unexpected value {}; consult "
+                "QMessageBox API for more information".format(result)
+            )
+            return NotifyStatus.ERROR_UNKNOWN
--- a/launcher/tests/test_util.py
+++ b/launcher/tests/test_util.py
@@ -298,3 +298,76 @@ def test_get_logger():
     assert logger.name == test_prefix
     logger = util.get_logger(prefix=test_prefix, module=test_module)
     assert logger.name == "{}.{}".format(test_prefix, test_module)
+
+
+@pytest.mark.parametrize(
+    "os_release_fixture,version_contains",
+    [
+        ("os-release-qubes-4.0", "4.0"),
+        ("os-release-qubes-4.1", "4.1"),
+        ("os-release-ubuntu", None),
+        ("no-such-file", None),
+    ],
+)
+@mock.patch("Util.OS_RELEASE_FILE", os.path.join(FIXTURES_PATH, "os-release-qubes-4.0"))
+def test_is_sdapp_halted_yes(os_release_fixture, version_contains):
+    """
+    When sd-app state is 'Halted'
+    Then `Util.is_sdapp_halted()` should return True
+    """
+    output = bytes(
+        "NAME     STATE     CLASS     LABEL     TEMPLATE\nsd-app"
+        "    Halted    AppVM   yellow     sd-small-buster-template\n",
+        "utf-8",
+    )
+
+    with mock.patch("subprocess.check_output") as patched_subprocess_check:
+        patched_subprocess_check.return_value = output
+        assert util.is_sdapp_halted()
+
+
+@pytest.mark.parametrize(
+    "os_release_fixture,version_contains",
+    [
+        ("os-release-qubes-4.0", "4.0"),
+        ("os-release-qubes-4.1", "4.1"),
+        ("os-release-ubuntu", None),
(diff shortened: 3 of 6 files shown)


launcher/sdw-notify.py
```diff
@@ -54,17 +54,44 @@ def show_update_warning():
...
+    button_check_now.setText(strings.button_check_for_updates)
+    button_defer = dialog.button(QMessageBox.No)
+    button_defer.setText(strings.button_defer_check)
+
+    if Util.is_sdapp_halted():
+        dialog.setText(strings.description_notify_updates)
+    else:
+        # If sd-app is running or state is unknown, use longer warning message
+        dialog.setText(strings.description_notify_updates_sdapp_running)
+
```
gonzalo-bulnes, line 75: I think this dialog could be extracted as a standalone component (widget).
The way I see it, it is a static dialog (the contents wouldn't change while the dialog is open) so the condition `Util.is_sdapp_halted()` could be a parameter to the constructor.

The API of the new class would be the same as the `QMessageBox` API, so none of the code below should need any change, that's what makes me think the refactoring could be a good idea.

gonzalo-bulnes: Naming-wise, my first thoughts would go towards "update prompt", "security check prompt"... what do you think @rocodes?

launcher/sdw-notify.py
```diff
@@ -4,11 +4,10 @@
 for too long without checking for security updates. Writes output to a logfile,
 not stdout. All settings are in Notify utility module.
 """
-
 import sys
 
-from sdw_notify import Notify
-from sdw_updater_gui import Updater
+from sdw_notify import Notify, strings
```
gonzalo-bulnes, line 9: I really like the separate module of strings.

If we were to extract the dialog below, then those strings could be encapsulated within that namespace :slightly_smiling_face: 

Not sure if that's the most idiomatic approach, but when thinking of reducing the API surface, I think:

```sh
tree .
 .
└── security_check         # or whatever name makes sense, it's the one that will show up in imports
           ├── __init__    # exposes the dialog class
           ├── prompt.py   # or whatever is a good name for the dialog class
           └── strings.py  # imported by the file above but not exposed by __init__
``` 

Usage could look like (just examples, I picked names to illustrate what I'd try to expose from the module):

```py
import security_check
user_choice = security_check.Prompt(sdapp_halted=True).exec()
# I personally like this best, because there is no namespace collapse, "Prompt" is always namspaced.
```

or

```py
from security_check import SecurityCheckPrompt
user_choice = SecurityCheckPrompt(sdapp_halted=True).exec()
# Might be more idiomatic? I don't like too much that "SecurityCheckPrompt" suddenly is exported by this module too - :grimacing: 
```

gonzalo-bulnes (commented):
@rocodes As discussed earlier, here are a few thoughts re: **opportunity to extract the UI / dialog** out of the notifier. I've given temporary names to things, but it's really jsut to illustrate the general idea, see if you think it's something you'd want to do.


About **testing the UI**: I too have been thinking about that (mostly with https://github.com/freedomofpress/securedrop-client/issues/1345 in mind). This is what I've got to so far: [example](https://github.com/freedomofpress/securedrop-client/pull/1341/files#diff-b4d9842eb21d46ff58d89e9b4453eac3c37b21f2b9b586d36aeb31cb6e8b1aa2), [another example](https://github.com/freedomofpress/securedrop-client/pull/1346/files#diff-30ac31cdba573941cb1b5b9c12f5414f995538f23391b300d20aeb84bdb52f11) with [comments](https://github.com/freedomofpress/securedrop-client/pull/1346#discussion_r754943522).

rocodes:
`make black` passes on my local branch (for sanity, `make black --check --line-length=100` reports no changes would be made). Just trying to see what the difference is.

launcher/sdw-notify.py
rocodes: @gonzalo-bulnes see if the changes in ed8d97f work for you in terms of separation of concerns :)

I _was_ hoping that I could completely eliminate the dependency on Qt components in `sdw-notify.py`, which I think would be very clean, but that requires just a little more time and thought to avoid issues with creating/handling `QApplication` (managing event loops). My goal would be to start here, perhaps make some similar improvements to the `sdw-launcher` in terms of improving the way it does its imports, and eventually move away from dependencies on Qt components in all but the view components. But I'd like to hear your thoughts.

launcher/tests/test_util.py
```diff
@@ -302,3 +302,56 @@ def test_get_logger():
...
+def test_is_sdapp_halted_runs_once():
+    """
+    When check for sd-app state is run
+     Then exactly one subprocess call should be run
+    """
+    with mock.patch("subprocess.check_output") as patched_subprocess_check:
+        util.is_sdapp_halted()
+        patched_subprocess_check.assert_called_once()
+
+
```
gonzalo-bulnes, line 316: This sub-process call doesn't have any side-effects, it is used _for the value it returns_.

From that point of view, I don't think it is needed or useful to assert that it is called, as long as the behavior of the function (`util.is_sdapp_halted`) is tested for all the possible return values of the sub-process call. And that's what you did with the two tests below. :+1: 

On the other size, these call assertions are a bit brittle: they tend to break if the number of calls or the parameters change, in ways that often don't reflect an actual change of behavior of the function under test ("false positives/negatives"), and get in the way of refactoring.

Because of that I think it's safe, and I'd suggest removing this test.

gonzalo-bulnes (dismissed):
I think this does a great job at separating concerns! :heavy_check_mark: The `QApplication` event loop is indeed a bit tricky to extract, we can think of ways of doing that (I'll have the same question for the Client).

I'd remove one of the tests (specially because I've encountered a bunch of failing-for-no-good-reason tests yesterday so the feeling is fresh :grimacing:), and I left you a couple of naming thoughts, but none of these are blockers, so fell free to proceed! :slightly_smiling_face:

launcher/sdw_notify/Notify.py
```diff
@@ -16,19 +16,6 @@
 LAST_UPDATED_FILE = os.path.join(BASE_DIRECTORY, "sdw-last-updated")
 LAST_UPDATED_FORMAT = "%Y-%m-%d %H:%M:%S"
 
-# The lockfile basename used to ensure this script can only be executed once.
-# Default path for lockfiles is specified in sdw_util
-LOCK_FILE = "sdw-notify.lock"
```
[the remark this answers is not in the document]
eloquence: In truth, package users should disregard these variables - they're intended for developer use, defined as shared pseudo-constants to make them easier to modify in one place. If we ever expose any aspect of the updater to user configuration, it would likely be via the configuration file `config.json` in `/usr/share/securedrop-workstation-dom0-config` in `dom0`.

eloquence:
Re: caps for constants
> I'd be happy to switch to more typical Python naming conventions, 

AFAIK the all-caps style is idiomatic, see https://www.python.org/dev/peps/pep-0008/#constants

launcher/sdw_notify/Notify.py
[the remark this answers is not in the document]
gonzalo-bulnes: By "package users" I meant devs, not end-users! Sorry my wording was unclear. In my mind the "package" being a Python concept is used by devs. Anyhow, I agree in not thinking there is much/anything we'd want end-users (SecureDrop admins?) to configure in the foreseeable future. (The workstation is probably complex enough with reasonably set defaults!)

launcher/sdw_notify/NotifyApp.py
```diff
@@ -0,0 +1,69 @@
...
+    from PyQt4.QtGui import QMessageBox
+
+
+class NotifyStatus(Enum):
+    """
+    Supported exit states from Notify dialog.
+    """
+
+    DEFER_UPDATES = "0"
+    CHECK_UPDATES = "1"
```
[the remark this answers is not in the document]
rocodes: @gonzalo-bulnes I thought about the random strings/longer ints, but because I am handling QMessageBox's own return values (which can only be `QMessageBox.Ok` or `QMessageBox.No` and returning basically a flag, I think the risk of the wrong thing being returned is minimal.

As for why they are strings, it was for stylistic consistency with the existing enum structure in `Updater.py`, but I am open to changing it and/or changing both at the same time (again to keep consistency).

launcher/sdw_notify/Notify.py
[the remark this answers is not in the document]
rocodes: This is not an essential part of the changes, and is definitely the part I thought might be contentious ;) I'm open to reverting it, and/or changing the name of the file if that's the sticking point. (Condensed rationale for splitting out the variables into a separate file is in the PR description)

rocodes:
> Re: caps for constants
> 
> > I'd be happy to switch to more typical Python naming conventions,
> 
> AFAIK the all-caps style is idiomatic, see https://www.python.org/dev/peps/pep-0008/#constants

Yes - I was referring to the filenames themselves (`NotifyApp`, `UpdaterApp` instead of `notify_app`), I should have written camel vs snake case.

launcher/sdw_notify/Notify.py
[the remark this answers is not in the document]
gonzalo-bulnes: Certainly **not** a sticking point!

Clarification: I was trying to find out what package name would best describe what constants are expected in it, because I think the rationale for splitting them apart makes sense and I anticipate we're likely to want to repeat that in the future. At that point, the question will likely be: do this constant and that one belong together or should they be in distinct files? At that point I feel that `Const` as a name doesn't help much making distinctions.

Let's cross that naming bridge when we get there!

launcher/sdw_notify/NotifyApp.py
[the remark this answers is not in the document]
gonzalo-bulnes: No strong feelings, as you say, they're flags and I too think we'd notice any (unlikely) misuse during review. The type is easy to change anytime if we want to, so I wouldn't overthink it. If we find a strong argument in the future, it will be easy to do whatever makes sense then 👍

creviera (dismissed):
I can run through the test plan tomorrow. Your code comments are helpful and the new `NotifyDialog` class is a nice abstraction.

launcher/sdw-notify.py
```diff
@@ -52,19 +50,34 @@ def main():
...
+        # because the deferral period is simply the period before the next run of
+        # the notify script (defined in `securedrop-workstation/dom0/sd-dom0-crontab.sls`).
+        # Custom deferral logic could be defined by checking for a return value of
+        # NotifyStatus.DEFER_UPDATES.
+        if result == NotifyApp.NotifyStatus.CHECK_UPDATES:
+            log.info("Launching Preflight Updater")
+            updater = UpdaterApp.UpdaterApp()
+            updater.show()
+            sys.exit(app.exec_())
+    except ValueError:
```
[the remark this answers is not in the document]
rocodes: Yeah, maybe it's a bit clumsy--that was me trying to responsibly handle the scenario where `QMessageDialog.exec_()` returns something we aren't expecting. (Currently the dialog will only return one of `QMessageBox.Yes` or `QMessageBox.No` for all user behaviour, including closing the dialog or using the keyboard keys to manipulate the dialog. But if for example some future changes were made such that the dialog could enter a state where it returns a third option, like a `QMessageBox.Apply, I wanted to be clear that we aren't expecting that kind of behaviour.)
I'll add  comment but I'm also open to rewriting this if you think there's a nicer way to handle it.

rocodes:
@creviera Thank you for your review! I've addressed most of your review comments in fca95fb, and have rebased/squashed commits. The only thing I haven't addressed is the caps case in `strings.py`, since `launcher/sdw_update_gui/strings.py` uses lowercase too--I can put in a small formatting PR that addresses both if we decide to standardize on caps for the strings files.

launcher/sdw_notify/NotifyApp.py
```diff
@@ -0,0 +1,82 @@
...
+
+class NotifyStatus(Enum):
+    """
+    Supported exit states from Notify dialog.
+    """
+
+    CHECK_UPDATES = "0"
+    DEFER_UPDATES = "1"
+    ERROR_UNKNOWN = "2"
+
```
rocodes, line 27: @creviera per your feedback, reordered these and added the error state

launcher/sdw_notify/NotifyApp.py
```diff
@@ -0,0 +1,82 @@
...
+        if self._is_sdapp_stopped:
+            self.setText(strings.description_notify_updates)
+        else:
+            self.setText(strings.description_notify_updates_sdapp_running)
+
+    def run(self) -> NotifyStatus:
+        """
+        Launch dialog and return user selection.
+        Raises ValueError if dialog encounters unhandled state.
```
creviera, line 63: this can be removed now

rocodes: oops. removed in 2c70adf

launcher/sdw_notify/NotifyApp.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def run(self) -> NotifyStatus:
+        """
+        Launch dialog and return user selection.
+        """
+        result = self.exec_()
+
+        if result == QMessageBox.Ok:
+            sdlog.info(
+                "NotfyDialog returned {}, user has opted to check for " "updates".format(result)
```
creviera, line 68: [nit] remove `" "` appended to `updates`

creviera (approved):
## Testing

Check out this branch in your sd dev vm.

**Manual testing**
Copy this branch to dom0 via `make clone`, then `make dev`. 

Ensure `sd-app` VM is shut down. 
Edit the file in `dom0` called `~/.securedrop_launcher/sdw-last-updated` to change the date to a date >= 5 days ago.
Run `python3 /opt/securedrop/launcher/sdw-notify.py`:
- [x] Notification appears with prompt to check for updates. 
- [x] No "**Warning:**" text is present in the notification
- [x] The default selected button option is "Check for updates".
- [x] Clicking "Remind me later" (or pressing the Esc key) closes the dialog window. Note the time.

Start `sd-app` vm via `qvm-start sd-app` in `dom0`. Wait for the notify prompt to reappear.  
- [ ] The notify prompt appears again after a delay of t < 2 hours (The reminder runs every 2 hours on the even hours. The delay will depend on what time the user dismissed the dialog). 
  * Skipping after confirming that you tested this already
- [x] This time, the "**Warning:**" text is present in the notification
- [x] Clicking "Check for updates" launches the Preflight Updater.  
- [x] The Preflight Updater can run updates successfully. 
- [x] No GTK warnings or other errors are generated or printed to the `dom0` console when the Preflight Updater is launched.

**Automated testing**
- [x] All tests pass

creviera:
waiting for updater to finish, but all looks good! skipping the test for the `0 */2 * * *` change as part of the test plan since i confirmed with @rocodes that this was already tested and i don't think we need to wait until noon to confirm it again.

creviera:
and updater finished, client is launched, looks lovely :rocket:

Issue #649, opened 2020-11-20
Add tests for launcher/UpdaterApp.py


emkll:
Related to #646 

We currently, by policy, targeting 100% test coverage for `Updater.py`, which contains most of updater logic.

While `UpdaterApp.py` primarily handles the UI, we should still consider adding test coverage there to prevent regressions, given that we are introducing new functionality and further complexity (e.g. handling migrations)

zenmonkeykstop:
Test coverage has been increased from 2022 on. Closing, but test coverage for new launcher should also be at acceptable levels.

PR #610, opened 2020-09-04
Add support for PyQt5

eloquence:
## Status

Ready for review

## Description of Changes

Fixes #607.

Uses the appropriate Qt version:
- Qt4 on Qubes 4.0
- Qt5 on Qubes 4.1, or if `SDW_UPDATER_QT=5` is set
- Qt4 on other operating systems.

Adds PyQt5 to `dev-requirements.txt` (renamed from `test-requirements.txt` for consistency with https://github.com/freedomofpress/securedrop-workstation/blob/main/dev-requirements.txt).

**NOTE:** Once this PR is merged, the Qt5 version of the UI will need to be regenerated along with the Qt4 version every time `sdw_updater.ui` is modified, using `pyuic5`.

### Qubes 4.0
1. Copy this branch into `dom0` in a provisioned Qubes 4.0 SDW development or staging environment
2. Run the updater via `python3 launcher/sdw-launcher.py --skip-delta 0`
- [ ] Observe that the updater runs as before.
3. Run the updater via `SDW_UPDATER_QT=5 python3 launcher/sdw-launcher.py --skip-delta 0`
- [ ] Observe that the updater fails to run because Qt 5 is not installed.
4. Remove the file `~/.securedrop_launcher/sdw-last-updated`
5. Ensure that your system uptime is at least 30 minutes.
6. Run the notify script via `python3 launcher/sdw-notify.py`
- [ ] Observe that the update alert is shown as before.
### Qubes 4.1 (nightly)
1. Copy this branch into `dom0` in a provisioned Qubes 4.1-nightly SDW development or staging environment
2. Run the updater via `python3 launcher/sdw-launcher.py --skip-delta 0`
- [ ] Observe that the updater runs to completion with no graphical errors.
3. Remove the file `~/.securedrop_launcher/sdw-last-updated`
4. Ensure that your system uptime is at least 30 minutes.
5. Run the notify script via `python3 launcher/sdw-notify.py`
- [ ] Observe that the update alert is shown.
### Development environment
1. Activate a virtualenv and install the Qt5 requirements via `pip install --require-hashes -r launcher/qt5-requirements.txt`
2. Run the updater via `SDW_UPDATER_QT=5 python3 launcher/sdw-launcher.py --skip-delta 0`
- [ ] Observe that the updater starts (you won't be able to run it), and that your terminal indicates that Qt5 is used.
3. Remove the file `~/.securedrop_launcher/sdw-last-updated`
4. Ensure that your system uptime is at least 30 minutes.
[...]

PR #743, opened 2021-12-09
Check for network connection before running preflight updater

### LINKED ISSUE #633: Updater should exit early if no connection
The preflight updater currently happily goes through the motions of attempting an update, even if the network is down for the duration. We should instead attempt to detect network connectivity early, and error out if the network is down.

# Steps to reproduce

1. Disconnect from the network
2. Force a preflight updater run

# Expected result

An immediate error message that says "I cannot change the laws of physics, Captain! A've got to have network!"

# Actual result

Updater slowly progresses all the way to 100% before it reports an error.

rocodes:
## Status

Ready for review

## Description of Changes

Fixes #633 
Towards #649 

- Add network connectivity check before running Preflight Updater, Only launch UpgradeThread if network check is successful. 
- If network check is unsuccessful, show an error message. 
-  Add override command-line option, `--skip-netcheck`, to bypass the network check and allow the Prelight Updater to run as usual (this is for development/testing purposes).
- Add tests for `UpdaterApp.py` to test the above functionality.

## Testing

**Manual testing in Qubes/dom0**

Check out this branch and `make clone` it into dom0. Disconnect your computer from the internet. 
- [ ] run `python3 /opt/securedrop/launcher/sdw-launcher.py` and click "Apply Updates." Observe an immediate network failure error message.
- [ ] Reconnect to the internet and rerun the launcher script. Ensure you can successfully run the preflight updater.
- [ ] Ensure the above steps work with both ethernet and wireless connections.   

**Manual testing outside of dom0**

- [ ] check out this branch, activate your virtual environment, and run `python3 /opt/securedrop/launcher/sdw-launcher.py`. When you attempt to apply updates you should see the "network failed" message (regardless of your network status).
- [ ] Now run `python3 /opt/securedrop/launcher/sdw-launcher.py --skip-netcheck` and you should be able to continue to the updater "run" (which will not actually run or apply updates since you're outside of dom0). The network error dialog view should not appear. The updater "run" will finish with a failure message. (Note: you will encounter error messages printed to the console relating to the failure to run some updater functionality outside of dom0: 
`sudo: qubes-dom0-update: command not found`, 
`sudo: qubesctl: command not found`, `Running 'echo '3' > /home/user/.securedrop_client/sdw-update-status' on sd-app \ sd-app: Failed to access 'default_user' property`)  
- [ ] Ensure the above steps work with both the PyQt4 and the PyQt5 environments. You will need to `export SDW_UPDATER_QT=4` for PyQt4 (and 5 for PyQt5) as per [these instructions](https://github.com/freedomofpress/securedrop-workstation/tree/d1e8c98cc099152c7e8957c97248258a5dec6df7/launcher#developer-environment).

### DIFF
--- a/launcher/sdw_updater_gui/UpdaterApp.py
+++ b/launcher/sdw_updater_gui/UpdaterApp.py
@@ -32,10 +32,11 @@ def launch_securedrop_client():
 
 
 class UpdaterApp(QDialog, Ui_UpdaterDialog):
-    def __init__(self, parent=None):
+    def __init__(self, should_skip_netcheck: bool = False, parent=None):
         super(UpdaterApp, self).__init__(parent)
 
         self.progress = 0
+        self._skip_netcheck = should_skip_netcheck
         self.setupUi(self)
 
         # We use a single dialog with button visibility toggled at different
@@ -127,9 +128,11 @@ class UpdaterApp(QDialog, Ui_UpdaterDialog):
         simply stops the update attempt and does not affect the last
         UpdateStatus or affect the update timestamp.
         """
-        logger.info("Checking network connectivity")
-        updatevm = _get_update_vm()
-        if updatevm and _is_updatevm_connected(updatevm):
+
+        if self._skip_netcheck:
+            logger.info("Network check skipped; launching updater")
+            self.apply_all_updates()
+        elif _is_netcheck_successful():
             logger.info("Network check successful; checking for updates.")
             self.apply_all_updates()
         else:
@@ -186,46 +189,24 @@ class UpdaterApp(QDialog, Ui_UpdaterDialog):
         sys.exit()
 
 
-def _get_update_vm() -> str:
-    """
-    Helper function to query qubes-prefs for the designated UpdateVM.
-    """
-    updatevm = None
-
-    try:
-        updatevm_bytes = subprocess.check_output(["qubes-prefs", "updatevm"])
-        if updatevm_bytes:
-            updatevm = updatevm_bytes.decode("utf-8").strip()
-        else:
-            logger.error("qubes-prefs returned null value for UpdateVM")
-    except subprocess.CalledProcessError as e:
-        logger.error("Failed to complete UpdateVM query")
-        logger.error(str(e))
-
-    return updatevm
-
-
-def _is_updatevm_connected(updatevm: str) -> bool:
+def _is_netcheck_successful() -> bool:
     """
     Helper function to assess network connectivity before launching updater.
 
-    Assess network connectivity by attempting to reach one of sites used for
-    package updates.
+    Assess network connectivity by checking connection status (via nmcli) in
+    sys-net.
     """
-    curl_commands = [
-        b"curl -fsI https://mirrors.fedoraproject.org --connect-time 2 --max-time 5",
-        b"curl -fsI https://apt.freedom.press --connect-time 2 --max-time 6",
-        b"curl -fsI https://ftp.qubes-os.org --connect-time 3 --max-time 8",
-    ]
-    for command in curl_commands:
-        try:
-            result = subprocess.check_call(["qvm-run", "--quiet", updatevm, command])
-            if result == 0:  # If a network call completes successfully, stop checking
-                return True
-        except subprocess.CalledProcessError:
-            logger.error("{} (connectivity check) failed".format(command.decode("utf-8")))
+    command = b"nmcli networking connectivity check"
 
-    return False
+    if not Util.get_qubes_version():
+        logger.error("QubesOS not detected, cannot check network.")
+        return False
+    try:
+        result = subprocess.check_output(["qvm-run", "sys-net", command])
+        return result is not None and result.decode("utf-8") == "full"
+    except subprocess.CalledProcessError:
+        logger.error("{} (connectivity check) failed".format(command.decode("utf-8")))
+        return False
 
 
 class UpgradeThread(QThread):
--- a/launcher/sdw_updater_gui/UpdaterApp.py
+++ b/launcher/sdw_updater_gui/UpdaterApp.py
@@ -44,7 +44,7 @@ class UpdaterApp(QDialog, Ui_UpdaterDialog):
 
         self.applyUpdatesButton.setEnabled(True)
         self.applyUpdatesButton.show()
-        self.applyUpdatesButton.clicked.connect(self.apply_all_updates)
+        self.applyUpdatesButton.clicked.connect(self._check_network_and_update)
 
         self.cancelButton.setEnabled(True)
         self.cancelButton.show()
@@ -117,6 +117,34 @@ class UpdaterApp(QDialog, Ui_UpdaterDialog):
         self.progress = current_progress
         self.progressBar.setProperty("value", self.progress)
 
+    def _check_network_and_update(self):
+        """
+        Wrapper for `apply_all_updates` that ensures network connectivity
+        before updating, else stops the update and shows a connectivity error
+        message to the user.
+
+        Because this check happens before updates begin, an error at this stage
+        simply stops the update attempt and does not affect the last
+        UpdateStatus or affect the update timestamp.
+        """
+        logger.info("Checking network connectivity")
+        updatevm = _get_update_vm()
+        if updatevm and _is_updatevm_connected(updatevm):
+            logger.info("Network check successful; checking for updates.")
+            self.apply_all_updates()
+        else:
+            logger.error("Network connectivity check failed; cannot check for updates.")
+            self._show_network_error()
+
+    def _show_network_error(self):
+        """
+        Show the network error dialog state.
+        """
+        self.headline.setText(strings.headline_error_network)
+        self.proposedActionDescription.setText(strings.description_error_network)
+        self.cancelButton.setEnabled(True)
+        self.applyUpdatesButton.hide()
+
     def apply_all_updates(self):
         """
         Method used by the applyUpdatesButton that will create and start an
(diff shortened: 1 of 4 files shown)


launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -158,6 +186,48 @@ def exit_launcher(self):
...
+    Helper function to assess network connectivity before launching updater.
+
+    Assess network connectivity by attempting to reach one of sites used for
+    package updates.
+    """
+    curl_commands = [
+        b"curl -fsI https://mirrors.fedoraproject.org --connect-time 2 --max-time 5",
+        b"curl -fsI https://apt.freedom.press --connect-time 2 --max-time 6",
+        b"curl -fsI https://ftp.qubes-os.org --connect-time 3 --max-time 8",
+    ]
```
rocodes, line 219: This part is open to discussion. I think trying all 3 domains is overkill, 2 is probably sufficient to say we have connection problems (noting that more than one request will only be made if the first one fails). Also welcome feedback on timeout lengths.

rocodes:
A couple notes on this issue:

- Qubes doesn't bail early on updates if no network is detected  in either the update GUI or dom0 cli. It takes 2 minutes for an update (such as `sudo qubes-dom0-update` to error out when no network is present. The GUI updater will try to update each template/VM even if no network is present and even if the previous VMs have all error'd out due to lack of network.   
- Initially it seemed like there were 2 possible places to put this connectivity check logic; what I've implemented here (right before launching the UpgradeThread), or in `sdw-launcher.py.`  That one would actually be better (cleaner-looking and more separation of concerns), except that it would miss the cases where the notify script spawns the preflight updater.
- In the best case (users have a successful network connection), they see a ~1/4-1/2 second delay (whatever it takes the `curl` command to complete) and then the updates start. 
- However, in the worst case (no network connection), currently the users have a ~10-second delay while the connection attempts are made before the "Network connection failed" message appears, and in that time the UI is not responsive. I think we need to show _something_ to the users (for example: I could start the progress bar), but the reason I didn't implement any UI tied to the network check yet is that I didn't want to 'flash' a message that would disappear quickly for most users and look glitchy. (So it would need to be something like "if the check is taking more than a second or two, show the "Checking network..." UI).

Hope these comments give some context :)

launcher/sdw_updater_gui/UpdaterApp.py
conorsch: @rocodes How about:

```
qvm-run -p sys-net 'nmcli --terse -f state general'
```

That will output a string "connected" or "disconnected", depending on state. I've only tested it with wifi, but it should work equally well for ethernet, too.

rocodes: much improved, thank you :)  changed in  9c5b5e5

rocodes:
This is failing on CI due to #745, which I've proposed a fix for, but `make check` passes locally in both PyQt4-based and PyQt5-based venvs

rocodes:
@conorsch I've marked this as "Ready for review" because I'm ready for feedback, but #746 needs to land first before this will pass on CI.

launcher/sdw_updater_gui/UpdaterApp.py
rocodes: I did some digging and I'm not sure that `nmcli -f state general` differentiates between connected locally (but no internet access) and connected (with internet access), whereas `nmcli networking connectivity check` will check immediately, and report back `full`, `limited`, `none`, [etc](https://developer-old.gnome.org/NetworkManager/stable/nmcli.html) (so if a user is connected to a network with a captive portal, the status will be `portal`).

launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -158,6 +189,26 @@ def exit_launcher(self):
...
 
 
+def _is_netcheck_successful() -> bool:
+    """
+    Helper function to assess network connectivity before launching updater.
+
+    Assess network connectivity by checking connection status (via nmcli) in
+    sys-net.
+    """
+    command = b"nmcli networking connectivity check"
```
rocodes, line 199: see [this comment](https://github.com/freedomofpress/securedrop-workstation/pull/743#discussion_r770889988)

launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -32,10 +32,11 @@ def launch_securedrop_client():
 
 
 class UpdaterApp(QDialog, Ui_UpdaterDialog):
-    def __init__(self, parent=None):
+    def __init__(self, should_skip_netcheck: bool = False, parent=None):
         super(UpdaterApp, self).__init__(parent)
```
rocodes, line 36: The default value here isn't _strictly_ necessary, but it preserves the ability to create the Prelight Updater without any parameters, which looks cleaner and is nice for e.g. the `notify` script.

launcher/tests/test_updaterapp.py
```diff
@@ -0,0 +1,145 @@
...
+from unittest import TestCase
+from unittest import mock
+from unittest.mock import call
+from importlib.machinery import SourceFileLoader
+
+
+# If we're in a PYyQt4 virtual environment, it will be called `.venv-pyqt4` when set up
+# according to our instructions with the pyqt4 venv Makefile target, and we should use
+# the PyQt4 dependencies. Otherwise, use PyQt5
+if sys.prefix.endswith("pyqt4"):
```
rocodes, line 14: This is a little bit brittle because it relies on the name of the virtual environment to decide whether we should import PyQt4 or PyQt5. If a user didn't follow the developer docs (which include Makefile targets for PyQt4-based venv and PyQt5 based venv), or if we ever decided to drop those Makefile targets, this check wouldn't work. However, it would default to PyQt5, which is the newer/more common standard. 

I'm open to feedback if there's a better way of managing this for testing purposes though!

conorsch: All prod installs on Qubes 4.0 are still using PyQt4, via F25 in dom0, so that should be the default. The logic in the `if` statement on L14 seems to invert the "use 4 unless 5 is requested" pragma introduced back in #610. Taking a closer look, we're using Python3.8 in CI to test the launcher code, but dom0 is still on 3.5. This isn't a regression introduced by your changes, @rocodes, but your changes are surfacing them. =) I'll take a look at what switching CI over to a more prod-like testing environment would take, and report back.

conorsch:
@rocodes I took a look at the CI and unbroke it as best I could. Also sprinkled in some container updates as described in https://github.com/freedomofpress/securedrop-client/issues/1384, because it was convenient.

> Work in progress

The tests are passing, so to my eye, this PR is ready for review! Happy to run through a deep review if you feel you're finished with logic changes for now.

launcher/tests/test_updaterapp.py
rocodes: Thanks for fixing CI, lifesaver. :) 

My comment here (around PyQt4 vs PyQt5 being the default) was only relevant to machines that are performing unit testing (for machines that run `test_updaterapp.py` specifically).  But in any case, I've reworked this section to flip the logic to be more consistent with the rest of the codebase, and to check the `SDW_UPDATER_QT` environment variable to decide which version of PyQt to import (note: this means test environments must follow the dev docs and always set the environment variable before running tests).

rocodes:
@conorsch yup, ready for you, and thanks for your work on CI :)

.circleci/config.yml
```diff
@@ -3,33 +3,33 @@ version: 2
 jobs:
   lint:
     docker:
-      - image: circleci/python:3.7-buster
+      - image: debian:buster
```
conorsch, line 6: All these CI config and dependency updates can be pulled out into a separate PR, if you'd prefer—just say the word, @rocodes. Otherwise I'll leave them in as-is.

launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -158,6 +189,26 @@ def exit_launcher(self):
...
+    sys-net.
+    """
+    command = b"nmcli networking connectivity check"
+
+    if not Util.get_qubes_version():
+        logger.error("QubesOS not detected, cannot check network.")
+        return False
+    try:
+        result = subprocess.check_output(["qvm-run", "sys-net", command])
+        return result is not None and result.decode("utf-8") == "full"
```
conorsch, line 206: The subprocess call on L205 will _always_ return an empty string (`b''`), because `qvm-run` doesn't return stdout/stderr from the domU unless `--pass-io` is given—so the function containing this check will always return `False`, as written. After ensuring stdio is properly handled, make sure to account for any whitespace; I expect `b'full\n'` will be the raw return value. You can match against that, or strip whitespace and compare the string.

launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -158,6 +189,26 @@ def exit_launcher(self):
...
+    except subprocess.CalledProcessError:
+        logger.error("{} (connectivity check) failed".format(command.decode("utf-8")))
```
conorsch, line 208: Logging is good! More logging! However, I wasn't able to find this message in the logs during my testing. Logging the actual return value would be helpful during debugging sessions. From [the docs](https://docs.python.org/3/library/subprocess.html#subprocess.check_output): 

> If the return code was non-zero it raises a CalledProcessError. The CalledProcessError object will have the return code in the returncode attribute and any output in the output attribute.

So I suggest something like:

```
except subprocess.CalledProcessError as e:
    logger.error("{} (connectivity check) failed; state reported as '{}'".format(command.decode("utf-8"), e.output))
```

Make sure you can find that string in the on-disk log files when a failure event is hit.

launcher/sdw_updater_gui/UpdaterApp.py
```diff
@@ -158,6 +189,26 @@ def exit_launcher(self):
         sys.exit()
 
 
+def _is_netcheck_successful() -> bool:
```
conorsch, line 192: Logically the `_is_netcheck_successful` seems like a utility to me, and therefore best resides in `sdw_util/`. That's a nit, so don't change it just for cleanliness's sake, but given the logging problems mentioned in this review on L206, maybe moving into utils will make the logging config easier to bootstrap.

conorsch (changes requested):
Encountered two (2) problems during local testing:

1. network check always returns False, so updater doesn't run; see comments about `--pass-io`
2. some of the helpful logging messages never make it to disk; the problem seems only to affect logging calls like L206; I _did_ see log messages like those on L139.

Magnificent test plan, thanks for taking the time!

conorsch:
Summarizing out of band discussion: we went ahead and merged the CI fixes in a separate PR (https://github.com/freedomofpress/securedrop-workstation/pull/752), so @rocodes is going to rebase here and ping when it's ready for a second look.

launcher/sdw_updater_gui/UpdaterApp.py
rocodes: Thank you -addressed in d5f2d1d (Note this log message will not be encountered in a no-network scenario; in that case nmcli will still return 0 but the error log will read `Network connectivity check failed`).

launcher/sdw_updater_gui/UpdaterApp.py
rocodes: I left this as-is because this is currently the only place we have a need for connectivity checking in the Updater code, and I have a vague future wish to disentangle some of the dependencies in UI classes (so eg Updater.py can import Util.py, but maybe some day in the distant future the UpdaterApp, more UI-focused, won't have dependencies outside of the `sdw_updater_gui` package).  However, I agree that some methods (this `netcheck`, and maybe `launch_securedrop_client`) could belong elsewhere.

eaon:
This looks and works great! Followed the test plan, success rate was :100: Logging also works as expected, showing connectivity or lack thereof in dom0 and failure to detect Qubes OS outside of dom0, including netcheck skips.

Nice work!

conorsch (approved):
Looks great! Tested mostly in dom0, and confirmed that log messages were recorded correctly when network was disconnected. Online via wifi or ethernet was correctly understood as "online". Dabbled with the qt4/pqt5 differences in the dev env, and both appeared to work well (with qt5 unsurprisingly looking subtly spiffier). 

Thanks for your patience and fortitude on getting this change to land, @rocodes!

PR #750, opened 2022-01-12
Correct template names in comments

eloquence:
## Status

Ready for review. These templates were renamed as part of template consolidation / Buster migration, respectively.


conorsch (approved):
Thank you, @eloquence!

PR #751, opened 2022-01-13
Initial support for Qubes 4.1

eaon:
## Status

Ready for review. Towards #600.

## Description of changes

* Enables installation of SDW on Qubes 4.1.
* Updates qrexec policies to use the [new R5.0 format](https://github.com/QubesOS/qubes-core-qrexec/blob/master/doc/multifile-policy.markdown) where appropriate. We now write some policies to the new location, but continue to maintain others in the legacy location, in order to ensure our policies remain defensive (e.g. `qubes.Gpg`).
* Removes unused duplicate RPM repo config from `sys-firewall`; closes #702. Technically this change affects 4.0, as well, but was convenient given that sys-firewall is a DispVM in Q4.1.
* Liberal test updates throughout. Mostly that's for RPC policies, but also made sure to update the gpg calls for validating key material.

## Known problems

Export and print functionality hasn't been implemented for Qubes 4.1 yet. Under 4.1, the `sys-usb` VM is disposable by default, meaning our custom udev rules do not persist. We should handle that change in a separate PR.

The `sd-log` setup will complain (via GUI notifications of qrexec denial) about lack of loopback support; see #755. We may be able to leverage `notify=false` in the R5 policy syntax to resolve.

The dom0 repos currently hardcode the 4.0 locations: https://yum-test.securedrop.org/workstation/dom0/f25/ That's good enough for now, but we'll want to create f32 locations on the yum server(s), and update the URLs here accordingly.

## Test plan
We want to verify that the Workstation behaves well under both Qubes 4.0 & Qubes 4.1. To that end, we'll need to test on both platforms. Let's divvy up the testing obligations across the team, to minimize duplication of effort. Tests we'll need to investigate on each platform:

- Clean install works: `make clone && make clean && make dev` and `make test` has only `test_sys_usb` related tests failing

- Perform full suite of [Qubes scenarios](https://github.com/freedomofpress/securedrop-workstation/wiki/Workstation-Acceptance-Tests#qubes-scenarios)
- Perform full suite of [Client QA checks](https://github.com/freedomofpress/securedrop-workstation/wiki/Workstation-Acceptance-Tests#client-scenarios)

See the wiki page for formatting your reports: https://github.com/freedomofpress/securedrop-workstation/wiki/Workstation-test-case-report-template

Issue #600, opened 2020-08-06
Suport for Qubes 4.1


eloquence:
Qubes 4.1 is not out and no release date has been announced yet. However, given the scope of expected changes, we should begin testing early to ensure a smooth update for SecureDrop Workstation users. Test builds can be downloaded from [OpenQA](https://openqa.qubes-os.org/), e.g., https://openqa.qubes-os.org/assets/iso/Qubes-4.1-20200726-x86_64.iso

Major changes that have been announced include:
- [changes to the Qrexec policy system](https://www.qubes-os.org/news/2020/06/22/new-qrexec-policy-system/), including a new policy format, performance improvements, support for socket services, and policy notifications;
- an experimental and optional [GUI domain](https://www.qubes-os.org/news/2020/03/18/gui-domain/) ("a qube separate from dom0 that handles all the display-related tasks and some system management")
- [upgrade from fedora-25 to fedora-32 in dom0](https://www.qubes-os.org/doc/supported-versions/#dom0)

eloquence:
During the 8/6-8/20 sprint, @emkll has offered to take a test build for a spin, and @rmol has previously suggested doing the same. We've allocated some time to the sprint to capture initial findings, so we can get a better sense of the scope of changes that will be needed fro the SecureDrop Workstation codebase.

eloquence:
We've not made progress on this yet, but both @rmol and @emkll want to poke at this during the 8/20-9/2 sprint. Beyond installing Qubes 4.1, this may include attempting to build/install the RPM if time permits.

conorsch:
Although we've discussed it several times, it's worth reiterating that running an early build of Qubes 4.1 will prohibit development and testing of the Workstation on that hardware, since the repository channels and several APIs have changed significantly. That's OK, though, and precisely the info we hope to gather by running it.

Rather than installing fresh from the ISO, it might be worthwhile to test-drive the burgeoning plans for an in-place upgrade: https://github.com/QubesOS/qubes-issues/issues/5685 , so we can share test results with upstream. As always, backups strongly advised. :smile:

emkll:
Timeboxed `make dev` in a freshly installed Qubes 4.1 install (using the ISO in https://openqa.qubes-os.org/tests/10827/asset/iso/Qubes-4.1-20200726-x86_64.iso)

Here were the steps needed in order to unblock the install:

* `make` was not present in dom0: `sudo qubes-dom0-update make`
[...]

PR #764, opened 2022-04-21
4.1 support: `sys-usb` DispVM compatibility and `sd-log` qrexec loopback denial notification suppression

### LINKED ISSUE #755: sd-log tries to log to itself, but loopback qrexec connections aren't supported
While looking into policy changes/details this stuck out to me: `qrexec: securedrop.Log: sd-log -> sd-log: denied: loopback qrexec connection not supported`. @conorsch confirmed this also affects 4.0 when we briefly touched on this in passing in #751, but I think this is worth breaking out for a different reason:

On 4.0 this denial is silent for users, but 4.1 ramped up user facing reporting of such denials, so this triggers a fairly alarming :no_entry: notification visible when running the preflight updater:

Screenshot_2022-01-21_17-54-04

While it's functionally not a problem as far as I've gathered, this may warrant handling to improve the optics.

eaon:
## Description of Changes

Fake "stacked" PR waiting for #751 being merged first. Towards #600.

The Qubes OS 4.1 default configuration makes the `sys-usb` qube
disposable, which interferes with our udev rule to automatically attach
flash devices to `sd-devices`. This change clones the `fedora-34-dvm`
qube as `sd-fedora-dvm`, where we henceforth add the necessary
modifications, so `sys-usb` stays disposable, but with our modifications
sticking around permanently.
    
In 4.1 policy denials trigger a notification visible to the user -
qexec loopback isn't supported in 4.0 either, but they're silent.
Suppressing these notifications is only supported in the new policy
format, so `securedrop.Log` rules are now there.

Fixes #755 

## Testing

* QubesOS version:
 * If 4.1, `sys-usb` is a DispVM:
* Workstation environment:
* Server environment:

### dom0

- [ ] During the initial `sdw-admin --apply` run, there are no error notifications along the lines of #755
- [ ] `sdw-admin --apply` successfully terminates
- [ ] All tests (`make test`) pass (prerequisite: https://github.com/freedomofpress/securedrop-debian-packaging/issues/312 is fixed)

### Updater (4.1 only)

- [ ] During the update, no error notifications along the lines of #755 are shown

### Client

- [ ] When a USB flash drive is connected to the machine, it is automatically attached to `sd-devices` (which boots if it is not running)
- [ ] Export UX works as expected (see [export instructions in Workstation Acceptance Tests(https://github.com/freedomofpress/securedrop-workstation/wiki/Workstation-Acceptance-Tests#export)])

## Deployment

For new 4.1 installs both `sys-usb` DispVMs (default configuration) and AppVMs are supported, though I think we should [advertise the default configuration in the docs](https://github.com/freedomofpress/securedrop-workstation/pull/764#issuecomment-1112350309).

If migrated 4.0 installs keep modifications to their `sys-usb` AppVM during/after upgrade, the salt states should also continue to work as expected on 4.1.

### DIFF
--- a/dom0/sd-clean-all.sls
+++ b/dom0/sd-clean-all.sls
@@ -9,8 +9,43 @@ set-fedora-as-default-dispvm:
 
 {% set gui_user = salt['cmd.shell']('groupmems -l -g qubes') %}
 
+{% if grains['osrelease'] == '4.1' and salt['pillar.get']('qvm:sys-usb:disposable', true) %}
+restore-sys-usb-dispvm-halt:
+  qvm.kill:
+    - name: sys-usb
+
+restore-sys-usb-dispvm-halt-wait:
+  cmd.run:
+    - name: sleep 5
+    - require:
+      - qvm: restore-sys-usb-dispvm-halt
+
+restore-sys-usb-dispvm:
+  qvm.prefs:
+    - name: sys-usb
+    - template: fedora-35-dvm
+    - require:
+      - cmd: restore-sys-usb-dispvm-halt-wait
+      - cmd: set-fedora-as-default-dispvm
+
+restore-sys-usb-dispvm-start:
+  qvm.start:
+    - name: sys-usb
+    - require:
+      - qvm: restore-sys-usb-dispvm
+
+# autoattach modifications are only present in sd-fedora-dvm
+# so no more sd-usb-autoattach-remove necessary
+remove-sd-fedora-dispvm:
+  qvm.absent:
+    - name: sd-fedora-dvm
+    - require:
+      - qvm: restore-sys-usb-dispvm
+{% else %}
+# If sys-usb is not disposable, clean up after ourselves
 include:
   - sd-usb-autoattach-remove
+{% endif %}
 
 # Reset desktop icon size to its original value
 dom0-reset-icon-size-xfce:
--- a/dom0/sd-dom0-qvm-rpc.sls
+++ b/dom0/sd-dom0-qvm-rpc.sls
@@ -193,6 +193,8 @@ dom0-rpc-qubes.r5-format-deny:
   file.managed:
     - name: /etc/qubes/policy.d/70-securedrop-workstation.policy
     - contents: |
+        securedrop.Log          *           @anyvm @anyvm deny
+
         qubes.FeaturesRequest   *           @anyvm @tag:sd-workstation deny
         qubes.FeaturesRequest   *           @tag:sd-workstation @anyvm deny
 
@@ -221,6 +223,10 @@ dom0-rpc-qubes.r5-format-ask-allow:
   file.managed:
     - name: /etc/qubes/policy.d/60-securedrop-workstation.policy
     - contents: |
+        # required to suppress unsupported loopback error notifications
+        securedrop.Log          *           sd-log sd-log deny notify=no
+        securedrop.Log          *           @tag:sd-workstation sd-log allow
+
         qubes.Filecopy          *           sd-log @default ask
         qubes.Filecopy          *           sd-log @tag:sd-receive-logs ask
         qubes.Filecopy          *           sd-proxy @tag:sd-client allow
--- a/dom0/sd-log.sls
+++ b/dom0/sd-log.sls
@@ -31,6 +31,7 @@ sd-log:
     - require:
       - qvm: sd-small-buster-template
 
+{% if grains['osrelease'] == '4.0' %}
 # Allow any SecureDrop VM to log to the centralized log VM
 sd-log-dom0-securedrop.Log:
   file.prepend:
@@ -38,6 +39,15 @@ sd-log-dom0-securedrop.Log:
     - text: |
         @tag:sd-workstation sd-log allow
         @anyvm @anyvm deny
+{% elif grains['osrelease'] == '4.1' %}
+# In 4.1 this policy is handled in the more central app policy
+# files added by sd-dom0-qvm-rpc.sls, no need to keep this
+# around in 4.0 if we migrated
+sd-log-dom0-remove-old-securedrop.Log-policy:
+  file.absent:
+    - names:
+      - /etc/qubes-rpc/policy/securedrop.Log
+{% endif %}
 
 {% import_json "sd/config.json" as d %}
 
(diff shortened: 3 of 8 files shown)


eloquence:
Question: Could it make sense to instruct users to set up sys-usb as persistent, and therefore avoid having to customize it in this manner?

eaon:
> Could it make sense to instruct users to set up sys-usb as persistent …

I'd tend towards supporting default configurations in any case - that choice is made on one screen and changing any `sys-` qube from DispVM to AppVM (or vice versa) is involved enough to not ask users to do that, and requiring a reinstall just for that would be a shame too (from a time-sink perspective).

As far as I can tell, there's no real downside to this way of handling it, we get updates from the base template, we have our "permanent" customisations but get the benefits of `sys-usb` being a little bit more protected.

eloquence:
(As discussed, I'll take this for a quick spin on my in-place-upgraded setup which uses 4.1 repos in the SecureDrop Workstation templates.)

eloquence:
* QubesOS version: 4.1 (in-place upgrade)
 * If 4.1, `sys-usb` is a DispVM: no
* Workstation environment: staging, built from this branch, SecureDrop Client 0.7.0rc3 (latest on apt-test)
* Server environment: production (2.3.2)

On this system, both exports and logging don't work; I've not double-checked, but I doubt it's related to the changes on this branch. `sdw-admin --apply` with an RPM from this branch runs without errors and makes no changes to `sys-usb` since it's not disposable. All export attempts fail with ": See your administrator for help". I don't see any RPC policy errors in dom0, or AppArmor denials in `sd-app` or `sd-devices`. 

USB devices _are_ auto-attached, and I'm able to mount them manually in `sd-devices`. Logs from `sd-devices` indicate that the preflight check returns a `USB_CONNECTED` result but does not continue from there. In debug mode, SecureDrop Client just logs a preflight check failure without further information. Will need to dig further into the code to see where it's not getting the result it expects.

As for logging, I'm continuing to observe the issue I reported here: https://github.com/freedomofpress/securedrop-workstation/pull/751#issuecomment-1101841764

I've split this out into its own issue (#773).

creviera:
@eaon - does this sound like the right order of next steps:
1. Test and merge https://github.com/freedomofpress/securedrop-client/pull/1485 against a 4.0 system to fix the issue @eloquence mentioned above
2. Test and merge this PR once a reviewer tests on a 4.1 fresh install (it's fine if the system still points to Qubes 4.0 packages 
3. Merge https://github.com/freedomofpress/qubes-template-securedrop-workstation/pull/24 and rebuild all 9 workstation packages for Bullseye 
4. Do a full round of QA on a fresh 4.1 install with the new template and kernel

There are some details will need to discuss around steps 3 and 4, which we can flesh out in the next sprint or two.

creviera:
> On this system, both exports and logging don't work; I've not double-checked, but I doubt it's related to the changes on this branch. `sdw-admin --apply` with an RPM from this branch runs without errors and makes no changes to `sys-usb` since it's not disposable. All export attempts fail with ": See your administrator for help". I don't see any RPC policy errors in dom0, or AppArmor denials in `sd-app` or `sd-devices`. 
> 
> USB devices _are_ auto-attached and show up in Nautilus, and I'm able to mount them manually in `sd-devices`. Logs from `sd-devices` indicate that the preflight check returns a `USB_CONNECTED` result but does not continue from there. In debug mode, SecureDrop Client just logs a preflight check failure without further information. Will need to dig further into the code to see where it's not getting the result it expects.

I think you're right that this seems to be unrelated to this PR after reviewing the code changes here. But we'll have another opportunity to do a full round of testing against a prod-like environment in step 4.

eloquence:
I dug a bit further into why exports aren't working on my 4.1 system. `export.py` runs this `qvm-open-in-vm` command:

https://github.com/freedomofpress/securedrop-client/blob/main/securedrop_client/export.py#L115-L118

The command appears to return no output, when the script expects output such as `USB_CONNECTED` to determine the preflight status. I'll do a bit more digging into 4.0/4.1 differences to see if this is another `qvm-open-in-vm` behavior change. @eaon I'd be very curious if exports work for you with 4.1 packages in all templates.

eloquence:
Dug a bit further into the export issue and indeed it's again a difference between the 4.0/4.1 version of `qvm-open-in-vm`, see #775. So this is definitely an issue we have to resolve for full 4.1 support, but not related to this PR.

cfm:
Assigning to myself to test as soon as I have Qubes 4.1 installed (which I'll prioritize next week).

creviera:
> - [ ] Export UX works as expected (see [export instructions in Workstation Acceptance Tests(https://github.com/freedomofpress/securedrop-workstation/wiki/Workstation-Acceptance-Tests#export)])

Could someone confirm that we won't be able to see this PR's test plan pass until https://github.com/freedomofpress/securedrop-workstation/issues/775 is fixed?

creviera:
Based on my reading, this explains why export failed on my 4.1 fresh install. So I'm assuming this PR is blocked until #775 is fixed, probably using the solution @eloquence and marek hashed out in the issue (using `qrexec-client-vm` directly instead of `qvm-open-in-vm` within the client). I can volunteer to make this change in the client on Monday or Tuesday next week, but don't want to step on anyones toes (so @eaon or @eloquence, we should sync up on Monday :)).

Also, do we want to remove this test from this PR's test plan and not block on this issue? Or shall we consider it blocked? Looking for your feedback @eaon.

eloquence:
FWIW, it doesn't look to me that any reviewer has successfully stepped through the full test plan on 4.1 yet, so it might be best to wait until the the #775 fix lands to ease testing this PR. It should be a pretty small fix.

creviera:
Okay, this should be unblocked now. Not sure when I'll be able to test next week, but probably on Tuesday, unless @cfm gets to it first with his recent fresh install of 4.1.

cfm:
I have Qubes 4.1 successfully installed and will test this tomorrow.

cfm:
(Review postponed at @creviera's request; removing my assignment and deferring to her for the current status update. :-)

creviera (approved):
LGTM based on visual review and @eloquence's confirmation that there is no effect when `sys-usb` is not disposable (we're not making `main` unreleasable by merging this).

Once this PR and https://github.com/freedomofpress/securedrop-debian-packaging/pull/339 are merged, we'll be able to test packages in `nightlies` on apt-test. Yay! :)

PR #768, opened 2022-05-03
Use proper name to describe salt state installing config.json on sd-app

### LINKED ISSUE #767: `sd-app-config.sls` uses wrong salt state name
The state that creates `config.json` for `sd-app` has the name "install-securedrop-proxy-yaml-config`, likely an oversight from basing the file on `sd-proxy-files.sls`.

It's not referenced by any other saltstate but still a bit confusing

eaon:
## Description of Changes

This just changes the name to be accurate, but it's more for developer convenience as it's never referenced by name

Fixes #767

## Testing

No, the change is far too trivial :grinning:

### DIFF
--- a/dom0/sd-app-config.sls
+++ b/dom0/sd-app-config.sls
@@ -13,7 +13,7 @@
 
 {% import_json "sd/config.json" as d %}
 
-install-securedrop-proxy-yaml-config:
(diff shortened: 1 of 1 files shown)


creviera (approved):
yuup, makes sense to me! :)

PR #780, opened 2022-06-02
Add signature check and warning for instructions on signing an RPM

creviera:
The signature check is also added for the reviewer, but let's catch any mistakes we make early on with the new signature check step. There's also now a warning to remind signers who are switching between signing keys to update `.rpmmacros`.

## Status

Ready for review

### DIFF
--- a/README.md
+++ b/README.md
@@ -527,6 +527,8 @@ rpm -Kv  # Signature lines will now contain OK instead of NOKEY
 
 You can then proceed with distributing the package, via the "test" or "prod" repo,
 as appropriate.
+   
+:warning: Remember to update your `~/.rpmmacros` file with the correct GPG key ID each time you switch between signing an RPM with the test key (required for `yum-test`) and the prod key (required for `yum`).
 
 #### `~/.rpmmacros` file
 
--- a/README.md
+++ b/README.md
@@ -517,18 +517,19 @@ Now we'll sign the RPM:
 
 ```
 rpm --resign <rpm>.rpm  # --addsign would allow us to apply multiple signatures to the RPM
-rpm -qi<file.rpm>  # should now show that the file is signed
+rpm -qi <file.rpm>  # should now show that the file is signed
 rpm -Kv  # should contain NOKEY errors in the lines containing Signature
 # This is because the the (public) key of the RPM signing key is not present,
 # and must be added to the RPM client config to verify the signature:
-sudo rpm --import <publicKey>.asc
+sudo rpm --import <publicKey>.asc  # Keys are in the lfs repos
 rpm -Kv  # Signature lines will now contain OK instead of NOKEY
+rpm qi <file.rpm> | grep -oP "key ID \K\w+"" | xargs gpg -k  # Verify the RPM was signed with the correct key
 ```
 
 You can then proceed with distributing the package, via the "test" or "prod" repo,
 as appropriate.
    
-:warning: Remember to update your `~/.rpmmacros` file with the correct GPG key ID each time you switch between signing an RPM with the test key (required for `yum-test`) and the prod key (required for `yum`).
+:warning: Remember that you hard-coded a GPG key ID in your `~/.rpmmacros` file to be used any time you sign an RPM from your Fedora machine/VM. If you have to switch between keys, you'll want to either update that file or use a different environment for the other key.
 
 #### `~/.rpmmacros` file
 


creviera:
@gonzalo-bulnes - I think you said you wanted to review this one. I marked it ready for review after our discussion in the retro today.

gonzalo-bulnes:
Yes! Thanks @creviera!

README.md
```diff
@@ -517,16 +517,20 @@ Now we'll sign the RPM:
 
 ```
 rpm --resign <rpm>.rpm  # --addsign would allow us to apply multiple signatures to the RPM
-rpm -qi<file.rpm>  # should now show that the file is signed
+rpm -qi <file.rpm>  # should now show that the file is signed
 rpm -Kv  # should contain NOKEY errors in the lines containing Signature
 # This is because the the (public) key of the RPM signing key is not present,
 # and must be added to the RPM client config to verify the signature:
-sudo rpm --import <publicKey>.asc
+sudo rpm --import <publicKey>.asc (keys are in the lfs repos)
```
zenmonkeykstop, line 524: nit: would delimit  comment part with `#` for cut-n-paste goodness.

README.md
```diff
@@ -517,16 +517,20 @@ Now we'll sign the RPM:
...
 rpm -Kv  # Signature lines will now contain OK instead of NOKEY
+rpm qi <file.rpm>  # Copy the KEY ID so you can paste it into the next command
```
zenmonkeykstop, line 526: nit: can the key ID be extracted via bash instead of copy-pasting?

README.md
```diff
@@ -517,16 +517,20 @@ Now we'll sign the RPM:
...
+gpg -k <KEY ID>  # Verify the RPM was signed with the correct key
 ```
 
 You can then proceed with distributing the package, via the "test" or "prod" repo,
 as appropriate.
+   
+:warning: Remember to update your `~/.rpmmacros` file with the correct GPG key ID each time you switch between signing an RPM with the test key (required for `yum-test`) and the prod key (required for `yum`).
```
zenmonkeykstop, line 533: This speaks to signing process so no need for detail, but it might be worth having test- and prod-only signing environments with the correct `.rpmmacros` to avoid the potential for user error here.

zenmonkeykstop (changes requested):
Additional clarifications are helpful, would be good to remove the need to swap `.rpmmacros` to reduce the risk of user error.

README.md
creviera: I updated the comment to be more generic.

README.md
creviera: looking into it...

creviera: Unfortunately it doesn't look like there's a direct rpm query to get just the key ID, but we could do something like `rpm -Kv <file.rpm> | grep -o "key ID .*:" | grep -Ei '[[:alnum:]]{8}*'` to get the key ID (gonna see if there's a better way)

creviera: Okay I think `rpm -Kv <file.rpm> | grep -oP 'key ID \K\w+'` works well... will update

creviera:
@zenmonkeykstop - your comments should be addressed, see https://github.com/freedomofpress/securedrop-workstation/pull/780/commits/5cbed4bd7c96868ec19c7b92d1d555bd7cd525f7

README.md
```diff
@@ -517,16 +517,19 @@ Now we'll sign the RPM:
...
-sudo rpm --import <publicKey>.asc
+sudo rpm --import <publicKey>.asc  # Keys are in the lfs repos
 rpm -Kv  # Signature lines will now contain OK instead of NOKEY
+rpm qi <file.rpm> | grep -oP "key ID \K\w+"" | xargs gpg -k  # Verify the RPM was signed with the correct key
 ```
 
 You can then proceed with distributing the package, via the "test" or "prod" repo,
 as appropriate.
+   
+:warning: Remember that you hard-coded a GPG key ID in your `~/.rpmmacros` file to be used any time you sign an RPM from your Fedora machine/VM. If you have to switch between keys, you'll want to either update that file or use a different environment for the other key.
```
creviera, line 532: @zenmonkeykstop - The most helpful catch will be the added step above around running `rpm qi <file.rpm> | grep -oP "key ID \K\w+"" | xargs gpg -k` to verify the RPM was signed with the expected key, but this general warning also should help remind folks that the id is hardcoded in that file. Does this seem like an adequate improvement to how we currently have no correct key check step for the signer?

creviera: actually, since we are moving the docs, I'll close this PR and move it into the wiki

creviera:
closing in favor or https://github.com/freedomofpress/securedrop-workstation/pull/828

PR #781, opened 2022-06-02
Updated config validation to compare fingerprint against the provided key

### LINKED ISSUE #779: sdw-admin script should check the fingerprint in config.json against the provided submission private key.
If an incorrect but syntactically valid fingerprint is provided in config.json, `sdw-admin apply` will set up the workstation using the incorrect key. The client application will still partially work: submissions can be downloaded and decrypted. But attempts to reply will fail immediately, with no useful feedback.

To prevent this, the SDWConfigvalidator class should check the fingerprint against the key's fingerprint. This would catch any transcription/cut-n-paste errors during setup.

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #779 .

Adds a validation check for the fpr in config.json, comparing it to the fpr of `sd-journalist.sec`.

## Testing
in a dev env on this branch:
- modify `config.json`, changing `submission_key_fpr` value to a syntactically valid (40 hex char) GPG fingerprint that does not match the one for the `sd-journalist.sec` key in use
- [ ] run `make dev` and confirm that you see a validation error indicating that the fingerprints don't match
- fix `config.json` by changing back to the correct fingerprint
- [ ] run make dev and confirm that validation succeeds, the install proceeds, and replies work in the client.

## Deployment

This change will not retroactively fix fingerprints in broken installs. Users will need to fix config.json manually and run `sdw-admin --apply` to correct any config errors.

### DIFF
--- a/scripts/validate_config.py
+++ b/scripts/validate_config.py
@@ -6,6 +6,7 @@ Checks for
 import json
 import re
 import os
+import sys
 import subprocess
 import tempfile
 from qubesadmin import Qubes
@@ -102,6 +103,14 @@ class SDWConfigValidator(object):
     def confirm_submission_privkey_fingerprint(self):
         assert "submission_key_fpr" in self.config
         assert re.match("^[a-fA-F0-9]{40}$", self.config["submission_key_fpr"])
+        gpg_cmd = ["gpg2", "--show-keys", self.secret_key_filepath]
+        try:
+            out = subprocess.check_output(gpg_cmd).decode(sys.stdout.encoding)
+            match = "      {}".format(self.config["submission_key_fpr"])
+            assert re.search(match, out), "Configured fingerprint does not match key!"
+
+        except subprocess.CalledProcessError as e:
+            assert False, "Fingerprint validation failed with error: {}".format(e.output)
 
     def read_config_file(self):
         with open(self.config_filepath, "r") as f:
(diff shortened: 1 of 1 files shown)


eloquence:
(Rebased to pick up version bump for RPM build)

scripts/validate_config.py
```diff
@@ -102,6 +103,14 @@ def confirm_submission_privkey_file(self):
     def confirm_submission_privkey_fingerprint(self):
         assert "submission_key_fpr" in self.config
         assert re.match("^[a-fA-F0-9]{40}$", self.config["submission_key_fpr"])
+        gpg_cmd = ["gpg2", "--show-keys", self.secret_key_filepath]
+        try:
+            out = subprocess.check_output(gpg_cmd).decode(sys.stdout.encoding)
```
eloquence, line 108: `check_output` defaults to `stderr=None`. Errors printed to `STDERR` by `gpg2` will therefore not be passed along for the exception handler below. To combine them, you could use `stderr=subprocess.STDOUT` here, or you could capture it via `stderr=subprocess.PIPE` and access it via `e.stderr`.

scripts/validate_config.py
```diff
@@ -102,6 +103,14 @@ def confirm_submission_privkey_file(self):
...
+            match = "      {}".format(self.config["submission_key_fpr"])
+            assert re.search(match, out), "Configured fingerprint does not match key!"
+
+        except subprocess.CalledProcessError as e:
+            assert False, "Fingerprint validation failed with error: {}".format(e.output)
```
eloquence, line 113: This string will be empty for typical `gpg` errors due to the issue above. It's also in binary string format, so I would suggest doing `.decode(sys.stdout.encoding)` here as well.

eloquence (changes requested):
Functionality looks good, noted a couple of nits re: error handling.

eloquence (approved):
Test plan passes; comments re: error handling have been addressed. Thank you! :)

PR #784, opened 2022-06-15
Use bullseye templates on 4.1 and and use apt-test+nightlies repos for dev apt-test+main packages for staging

eaon:
## Description of Changes

Towards #600

Draft PR for now, but this will include a collection of 4.1 bullseye template related changes to salt state logic, as well as @legoktm's work from the `bullseye-nightlies` branch. It also removes the `qa-switch.sh` utility (at least for now) as it would introduce unnecessary complexity for keeping 4.0 compatibility, and at least for this upcoming stage of testing it isn't useful to us and could probably be implemented in a cleaner way if we need it again in the future.

Fixes #787

## Testing

Qubes OS R4.1 only:

- [ ] `make dev` installs without errors, with `bullseye` based templates which in turn use `bullseye`+`nightlies` repositories
- [ ] `make test` in dom0 passes except for 1 failure about a `securedrop-workstation-grsec` that is not up to date (seems to be a nightly package/apt problem, the package hasn't been changed, there's no negative effects to this failure)

## Checklist

- [ ] I have updated `MANIFEST.in` and `rpm-build/SPECS/securedrop-workstation-dom0-config.spec`
- [x] I would appreciate help with the documentation

PR #790, opened 2022-06-23
update test to ensure new 2023 expiration

creviera:
## Status

Ready for review

## Description of Changes

In the 4.1 branch we added an update to the keyring test: https://github.com/freedomofpress/securedrop-workstation/pull/784/commits/d3c00c04af66eb67a5dca5eebdcce07ed4b2e3b1. Instead of cherry-picking that commit which would include a new dependency assertion that would fail on 4.0, I am adding the change directly to the `qubes-4.0` branch. 

## Testing

- In Qubes 4.0...
- [ ] See that the test that checks the keyring output now passes


eaon (approved):
Fresh install with the `qubes-4.0` branch, ran `python3 -m unittest discover -v tests -p test_vms_platform.py` and all tests passed!

PR #798, opened 2022-06-29
Fix Fedora template installation failures

### LINKED ISSUE #797: Provision-all error when Fedora-35 template already installed
Running through our install docs, we advise users to [upgrade to Fedroa 35](https://workstation.securedrop.org/en/stable/admin/install.html#install-fedora-35-template) before installing the workstation, to ensure that their `sys` VMs are up to date.

This creates a problem during our provisioning of sys vms, because `qvm-template install $template` returns a non-zero exit status (2) when the template is already installed, leading [this salt state](https://github.com/freedomofpress/securedrop-workstation/blob/b7eebfe3f54db6422afc3ba7e7d502bbc2a24cc6/dom0/sd-sys-vms.sls#L15-L19) (and those that depend on it) to error out.

Using `qvm.present` or `qvm.template_installed` instead of `cmd.run` would solve this issue.

gonzalo-bulnes:
## Description

Fixes #797

## Implementation

As far as I can tell, providing the name of the template to `qvm.template_installed` is enough. (See example code in the commit message.)

## Test plan

- Install Qubes OS R4.1 if needed
- Install the `fedora-35` template if necessary
- [ ] Apply this Salt state and confirm that it succeeds (`make dev`? - or you preferred way of applying the Workstation Salt states)

### DIFF
--- a/dom0/sd-sys-vms.sls
+++ b/dom0/sd-sys-vms.sls
@@ -12,11 +12,10 @@ include:
 {% set sd_supported_fedora_version = 'fedora-35' %}
 
 
-# Install latest templates required for SDW VMs.
-dom0-install-fedora-template:
-  cmd.run:
-    - name: >
-        qvm-template install {{ sd_supported_fedora_version }}
+# Ensure the latest templates required for SDW VMs are installed.
+dom0-fedora-template-is-installed:
+  qvm.template_installed:
+    - name: {{ sd_supported_fedora_version }}
 
 # Update the mgmt VM before updating the new Fedora VM. The order is required
 # and listed in the release notes for F32 & F33.


dom0/sd-sys-vms.sls
```diff
@@ -12,11 +12,10 @@ include:
 {% set sd_supported_fedora_version = 'fedora-35' %}
 
 
-# Install latest templates required for SDW VMs.
-dom0-install-fedora-template:
-  cmd.run:
-    - name: >
-        qvm-template install {{ sd_supported_fedora_version }}
+# Ensure the latest templates required for SDW VMs are installed.
+dom0-fedora-template-is-installed:
```
eloquence, line 16: Multiple subsequent commands `require` this one under the old name (`dom0-install-fedora-template`), so this Salt state will fail with "requisites were not found" error unless all those `require` conditions are updated, e.g.:

https://github.com/freedomofpress/securedrop-workstation/blob/b7eebfe3f54db6422afc3ba7e7d502bbc2a24cc6/dom0/sd-sys-vms.sls#L28-L29

eloquence (changes requested):
Problem with `require` resolution; note this may not resolve #797 (see discussion there).

dom0/sd-sys-vms.sls
eloquence: (Also note that all those `cmd` instances will need to be changed to `qvm`.)

gonzalo-bulnes: Thanks for the review and suggestions @eloquence :+1: 

I may wait until I'm in a position to test this before looking into it further, especially if #797 is unconfirmed/unclear at this point.

eloquence: Sounds good. In the meantime I've confirmed that with those fixes, it runs without errors and is idempotent. So I think this would be a neutral change.

PR #824, opened 2022-08-31
Updated README data flow diagram

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Updates data flow diagram:
- Updated qubes version to 4.1
- removed logging flow from sd-gpg

## Testing

Does the dfd display correctly in the readme? Does it accurately reflect the current data flow?


gonzalo-bulnes (commented):
Just for the record, I haven't verified that the `.draw` file matches the PNG, but that doesn't worry me much.

- [x] The diagram update matches the description. :heavy_check_mark: 
- [ ] Your commit is not signed @zenmonkeykstop :grimacing: (I assume that wasn't intentional?)

eloquence (changes requested):
As Gonzalo pointed out the commit is unsigned - since that's probably unintentional, explicitly requesting changes.

eloquence (approved):
Changes look good (I've verified in `draw.io` app as well).

Issue #710, opened 2021-06-09
Replace `grains.id` with `qubesdb-read` to support concurrent updates


creviera:
# Description

Follow-up issue for: https://github.com/freedomofpress/securedrop-workstation/issues/705

We were able to make a hot fix for #705 and now we need to decide, do we want to replace our usage of `grains`, specifically `grains.id` with `qubesdb-read` everywhere? Does the issue consistently go away when you turn off concurrency during updates, as @eloquence tested here: https://github.com/freedomofpress/securedrop-workstation/issues/705#issuecomment-857366241? Is `qubesdb-read` required in order to reliably use concurrency? If so, perhaps we should increase concurrency from 2 to 4?

rocodes:
Potentially relevant: we may instead be able to [adjust the way grains are cached](https://docs.saltproject.io/en/latest/ref/configuration/minion.html#grains-cache) (partly dropping this link here to remind myself to investigate this more later)

PR #827, opened 2022-09-15
Replace salt/jinja execution modules

### LINKED ISSUE #826: `sdw-admin --apply` fails with Jinja errors on Salt calls
## Description

`sdw-admin --apply` currently fails with multiple errors like this one:

```
Rendering SLS 'base:sd-mime-handling' failed: Jinja variable 'salt.utils.templates.AliasedLoader object' has no attribute 'cmd.shell'; line 23
```

## Steps to reproduce

On an up-to-date system, uninstall the SecureDrop Workstation, re-install, and run `sdw-admin --apply` or `make dev`

## Expected behavior

Installation completes

## Actual behavior

Installation errors out as noted above.

## Analysis

Salt 3005 appears to have a bug with Salt calls from within Jinja: https://github.com/saltstack/salt/issues/62636. Downgrading Salt resolves the issue.

rocodes:
## Status

Ready for review

## Description of Changes

Fixes #826

Avoid use of jinja/salt execution modules in default-mgmt-dvm. (Upstream: https://github.com/saltstack/salt/issues/62636).
 
Notes:
- Use of execution modules (for example, `salt['cmd.get']`) is unchanged in dom0, since Salt version there is more stable and is unaffected by this issue.
- The approach I took with @eaon initially relied on writing a custom grain to set the hostname; however, it's likely subject to the same bugginess as the original implementation with `grains['id']`, which we changed in favour of the execution module in 2021. The approach I've taken now involves separate sls files targeting each vm, but to keep things DRY they all use the same underlying macro.  It's a little bit clunky (cause there are small extra files), but it probably means we can boost our `max_concurrency` again (see https://github.com/freedomofpress/securedrop-workstation/issues/710), which the custom grains approach would not let us do. 
- We may not need this entire PR if the upstream fix gets merged soon.

## Testing

- [ ] Ensure you can repro the bug in #826 or the upstream bug on your target machine if you haven't done so already.
- [ ] Ensure provisioning works correctly: check out this branch then `make clean`, `make clone`, `make dev`
- [ ] Provisioning completes successfully. USB autoattach udev rules are present in `/etc/udev/rules.d/99-sd-devices.rules` on sd-fedora-dvm, and mimeapps.list symlink is correct in sd-app (sd-app), sd-devices-dvm (sd-devices-dvm), sd-proxy (default), sd-viewer (sd-viewer).
- [ ] `make test` passes on dom0

## Deployment

2. New installs

- [ ] All tests (`make test`) pass in `dom0`  **todo**

- [ ] I have updated `MANIFEST.in` and `rpm-build/SPECS/securedrop-workstation-dom0-config.spec`  **todo**

### DIFF
--- a/dom0/sd-usb-autoattach-add.sls
+++ b/dom0/sd-usb-autoattach-add.sls
@@ -6,15 +6,6 @@
 # USB devices to sd-devices.
 ##
 
-# If sys-usb is disposable, we have already set up sd-fedora-dvm to make our
-# modifications in, so we only want to modify sys-usb if it is a regular AppVM
-
-{% set apply = True %}
-{% if grains['id'] == 'sys-usb' and salt['pillar.get']('qvm:sys-usb:disposable', true) %}
-  {% set apply = False %}
-{% endif %}
-
-{% if apply %}
 sd-udev-rules:
   file.managed:
     - name: /rw/config/sd/etc/udev/rules.d/99-sd-devices.rules
@@ -48,4 +39,3 @@ sd-attach-export-device:
     - user: root
     - group: root
     - mode: 0555
-{% endif %}
--- a/dom0/sd-app-mime-handling.sls
+++ b/dom0/sd-app-mime-handling.sls
@@ -1,0 +1,6 @@
+include:
+  - sd-mime-handling
+
+{% from 'sd-mimeapps-macro.sls' import link_mimeapps with context %}
+
+{{ link_mimeapps('sd-app') }}
--- a/dom0/sd-devices-mime-handling.sls
+++ b/dom0/sd-devices-mime-handling.sls
@@ -1,0 +1,6 @@
+include:
+  - sd-mime-handling
+
+{% from 'sd-mimeapps-macro.sls' import link_mimeapps with context %}
+
+{{ link_mimeapps('sd-devices-dvm') }}
--- a/dom0/sd-mime-handling.sls
+++ b/dom0/sd-mime-handling.sls
@@ -8,8 +8,8 @@
 # Overrides mimetype handling for certain VMs. Instead of relying on the
 # /usr/share/applications (system volume), we instead use /home/user/.local/share/
 # (private volume). The various mimeapps.list files linked are provided by the
-# securedrop-workstation-config package in /opt/, and are symlinked here in their
-# respective AppVMs.
+# securedrop-workstation-config package in /opt/, and are symlinked in their
+# respective AppVMs by calling the mimeapps-macro with the correct symlink target.
 ##
 
 sd-private-volume-mimeapps-config-dir:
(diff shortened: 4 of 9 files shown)


rocodes:
Just leaving this in draft mode cause it's not ready for anyone to test/step through yet - I just wanted to push some of @eaon's and my work and wind down for the day.  I think what will actually make sense, as mentioned above, is to put this in a pillar, not in a grain, so I'll update tomorrow.

gonzalo-bulnes (commented):
For what is worth, both changes look good to me so far, understanding that we need to figure out a way to ensure the grains (or pillar) are fresh by the time they're read.

rocodes:
(one test failure in dom0, regarding `mailcap`, investigating)

dom0/sd-usb-autoattach-add.sls
```diff
@@ -6,15 +6,6 @@
...
 
-# If sys-usb is disposable, we have already set up sd-fedora-dvm to make our
-# modifications in, so we only want to modify sys-usb if it is a regular AppVM
-
-{% set apply = True %}
-{% if grains['id'] == 'sys-usb' and salt['pillar.get']('qvm:sys-usb:disposable', true) %}
-  {% set apply = False %}
-{% endif %}
-
-{% if apply %}
```
rocodes, line 17: the reason we aren't worried about this anymore is it's basically saying "only apply these changes if sys-usb is non disposable."  However, if sys-usb is disposable, the changes don't matter anyway since they are identical to the ones in the template (and are applied in the template).

We can probably remove all of this and drop support for non-disposable sys-usb, which we may not be fully supporting in 4.1 anyway.

gonzalo-bulnes:
The macro approach feels so much cleaner! I'll review the changes properly next week, but I love what I see so far! 💡💡💡

dom0/sd-usb-autoattach-add.sls
creviera: This decision makes sense to me. Thanks for the clear explanation.

creviera:
Thanks @gonzalo-bulnes! I'll have time to functional test this next week as well (still keeping an eye on upstream).

dom0/sd-usb-autoattach-add.sls
eaon: I was just re-reading `sd-sys-vms.sls` because when we went over it we (or I) thought it didn't support non-disposable `sys-usb` anymore anyway… but after revisiting I think it still technically does.

It doesn't _need_ to be addressed in this PR, but just wanted to note that this PR would technically alter what the code unofficially supports. To go the whole way, we could remove the `sd-sys-vms.sls` check [whether a VM is disposable or not](https://github.com/freedomofpress/securedrop-workstation/blob/main/dom0/sd-sys-vms.sls#L85-L96) and just assume it is.

rocodes:
Since an upstream fix has [landed recently](https://github.com/QubesOS/qubes-mgmt-salt/commit/f5fe0268f9d50c678e85521c31ec990e85475d43), we don't technically need to make the macro changes (or any of these changes on an urgent timeline).

I think I will close this PR if no one objects. We can be inspired by some of it (eg combination of macro and execution module) later if we choose to get fancy, but we needn't be in a hurry to make these changes (esp the ones that @eaon rightly pointed out would affect unofficial non-disp sys-usb support).  We can consider our overall Salt/Qubes strategy and make changes deliberately down the line.

PR #830, opened 2022-09-23
Correct license reported by the rpm and Python packages

### LINKED ISSUE #829: `LICENSE` is the AGPL, `.spec` says the project licensed with the GPL
Looks like there was an oversight: `LICENSE` is a copy of the AGPL while … [the spec implies it should be the GPL](https://github.com/freedomofpress/securedrop-workstation/blob/main/rpm-build/SPECS/securedrop-workstation-dom0-config.spec#L22).

Given that this repo does not cover any network accessible applications, I think it'd be reasonable to go with the GPL, plus other components like `securedrop-proxy` are licensed under the GPL as well.

eaon:
## Description of Changes

Does not change functionality, just puts license reporting in line with the actual license of the project.

Fixes #829


legoktm (approved):
Thanks for noticing and fixing!

creviera (commented):
lgtm

creviera:
oops... only 3 days behind

Issue #843, opened 2022-11-28
Nightly build of SecureDrop Client won't start after updates are applied successfully


eaon:
Due to https://github.com/freedomofpress/securedrop-client/pull/1600 and https://github.com/freedomofpress/securedrop-client/pull/1601 being merged, the updater now refers to an outdated `.desktop` file name to be launched with `gtk-launch`. As a consequence, the launch fails _silently_.

1. `gtk-launch` should not fail silently but the updater/launcher ought to display a the error to the user
2. It should refer to the new freedesktop spec compliant `.desktop` name

Probably warrants a new `securedrop-workstation` dev build package to fix for `dev` environment users, even if the next real release might include the [securedrop-updater](https://github.com/freedomofpress/securedrop-updater/) changes.

eaon:
This was fixed by #844

PR #844, opened 2022-11-28
Launch correct SecureDrop Client .desktop application

eaon:
## Description of Changes

Due to a file name change in [securedrop-client#1600](https://github.com/freedomofpress/securedrop-client/pull/1600) and [securedrop-client#1601](https://github.com/freedomofpress/securedrop-client/pull/1601), the client wouldn't get launched because the `securedrop-client.desktop` does not exist any longer. 

Towards #843.

## Testing

- [ ] Using the `dev` environment in `config.json`, the SecureDrop Client gets launched after using the launcher on the desktop

### DIFF
--- a/launcher/sdw_updater_gui/UpdaterApp.py
+++ b/launcher/sdw_updater_gui/UpdaterApp.py
@@ -18,7 +18,7 @@ def launch_securedrop_client():
     """
     try:
         logger.info("Launching SecureDrop client")
-        subprocess.Popen(["qvm-run", "sd-app", "gtk-launch securedrop-client"])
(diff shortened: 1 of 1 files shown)


gonzalo-bulnes:
A side-effect of the name change is that it will now be easier to know where the desktop entry is used!

gonzalo-bulnes (approved):
- [x] I can reproduce the error:
   - after checking out `main`
   - and applying updates
   - double-clicking on the desktop shortcut doesn't start the Client
- [x] The fix works as expected
   - after checking out this branch in `sd-dev`
   - then running `make dev`
   - when I double-click on the desktop shortcut, the Client starts

PR #851, opened 2023-01-24
Containerize build and test environment, add some developer tools

eaon:
## Description of Changes

* Change build environment to use containers of the package's target OS
  * Use "native" reproducibility macros as much as possible, removing need for custom scripting
  * Install `diffoscope` and `reprotest` from PyPI for minimal OS package requirements in CI and containers
  * Replace `sed` hack for `reprotest` container support with a custom (inert) `setarch` script, allowing seamless upgrades in the future
* Reorganise repo a bit …
  * … to distinguish between `files/` we ship and `scripts/` used for development
  * … move Python dependencies into `requirements/` in their own folder (as is the case in [`securedrop-client`](https://github.com/freedomofpress/securedrop-client/))
* Add `isort`, `bandit` and `shellcheck`, however, there's no generic `make lint`. CI is just running `black` as it did up until now
* Removes launcher from the `/srv/salt/` directory where it hasn't been used since `make clone` was introduced

Fixes #849
Fixes #806

Towards #837

## Testing

- [x] CI is happy :heavy_check_mark: 
- [x] The package built with `make build-rpm` installs as expected and SDW can be set up as expected
- [x] `make dev` and `make clone` still work

On an already-provisioned Workstation, you'll need to manually apply the following patch in `/home/user/securedrop-workstation` before `make clone` will succeed:

```patch
--- a/scripts/clone-to-dom0
+++ b/scripts/clone-to-dom0
@@ -25,7 +25,7 @@ dom0_dev_dir="$HOME/securedrop-workstation"
 # in the subsequent tarball, which is fetched to dom0.
 function build-dom0-rpm() {
     printf "Building RPM on %s ...\n" "${dev_vm}"
-    qvm-run -q "$dev_vm" "make -C $dev_dir dom0-rpm"
+    qvm-run -q "$dev_vm" "make -C $dev_dir build-rpm"
 }
 
 # Call out to target AppVM to create a tarball in dom0
```

- [x] `make test` passes in `dom0`

## Checklist

- [x] I have updated `MANIFEST.in` and `rpm-build/SPECS/securedrop-workstation-dom0-config.spec`
- [x] I would appreciate help with the documentation
  - Though I believe I've covered everything in this repo. `securedrop-dev-docs` would still need a PR

PR #856, opened 2023-02-14
Run flake8 normally and re-enable in CI; add mypy

### LINKED ISSUE #854: Add mypy type checker
*This is a good first issue for new contributors to take on, if you have any questions, please ask on the task or in our [Gitter room](https://gitter.im/freedomofpress/securedrop)!*

We should use the [`mypy`](https://mypy.readthedocs.io/en/stable/) type checker in this repository. You can probably copy most of the configuration from https://github.com/freedomofpress/securedrop-client. It would be good to fix simple issues in the same PR, but if some types are complicated, it's fine to mark them with `# type: ignore` to suppress the mypy warning for now so it can be cleaned up at a later time.

### LINKED ISSUE #855: flake8 isn't being run by CI
It was accidentally removed in https://github.com/freedomofpress/securedrop-workstation/commit/0878b172c27f631886089b49cdb9ceefa4e9255c.

Even if it was in the CircleCI config, it wouldn't really work because the `lint-all` script looks for files with a mime-type of `text/x-python`, except the version of file on bullseye outputs `text/x-script.python`.

(Unsurprisingly there are a bunch of flake8 errors when I run it locally).

The main purpose of `lint-all` is to catch files that don't have a `.py` extension. But there are only 4 of those files:
```
./scripts/destroy-vm
./scripts/configure-environment
./dom0/remove-tags
./dom0/securedrop-login
```

Two other Python scripts in scripts/ already have a `.py` extension so maybe we can just rename those two?

I'm not really sure why securedrop-login is a Python script, it just (uselessly?) configures logging and then shells out to another Python process...

Anyways, I would rather rename them or just explicitly list these 4 files in the flake8 config and move on.

legoktm:
## Status

Ready for review

## Description of Changes

Fixes #854.
Fixes #855.

Changes in this pull request:

The lint-all script was buggy (the mime is now text/x-script.python), so
it was actually excluding files that would've normally been detected.
Instead explicitly list the 4 extension-less Python files in the flake8
configuration.
    
To prevent this from being missed again, add a `make lint` target that
both CI and developers can use. Additional tools like isort and mypy can
be added to this as well.

The change to UpdaterApp.py is significant, as the `e.output` wasn't
being included in the logging.
    
The rest are unused and missing imports. I suppressed W503 since it goes
against newer PEP8 style, per
<https://www.flake8rules.com/rules/W503.html>.

Finally, configure mypy while we're at it. This is slightly different than other repositories because all the code is in individual scripts rather than a typical Python package.

## Testing

* [ ] CI passes

## Deployment

Any special considerations for deployment? Not really.


legoktm:
I tacked on mypy to this PR since it ended up being just adding more configuration/deps (no code changes), I can split it out to a separate one if that's preferred.

rocodes (approved):
Thanks for doing this and no need to split out mypy imo.
- CI looks good
- `make lint` passes locally 
- small thing: mypy references python 3.8, whereas dev-requirements is generated with 3.7, and our README says python 3.5+. We should clarify if we have any additional restrictions since this advice is a few years old, but we can do that separately/after.

legoktm:
The use of 3.7 seems to be because CI is testing on bullseye rather than Fedora 32. https://github.com/freedomofpress/securedrop-workstation/pull/851 fixes that and regenerates dev-requirements.txt, probably we can fix setup.py there too.

PR #802, opened 2022-07-01
Update to support/migrate to Fedora 36 templates

eaon:
## Description of Changes

Fixes #801

Update supported Fedora version to 36.

## Testing

- [ ] There are no references to Fedora 35 left in the repository
- `make dev` or `make clone` then `sdw-admin --apply`
- [ ] `sd-fedora-dvm` exists and is based on `fedora-36`
- [ ] `make test` passes in dom0

### Only relevant on 4.1

- [ ] New Fedora-36 DVM is created and displays in appmenu; other fedora dispvms can still be opened
- [ ] `default-mgmt-dvm` is `fedora-36`, as are templates for sys-net, sys-usb, sys-firewall

## Checklist

- [x] I would appreciate help with the documentation

PR #859, opened 2023-02-27
Bump supported Fedora release to 37

### LINKED ISSUE #841: Switch to Fedora 37
The release of Fedora 37 is around the corner, but Fedora 36 isn't EOL yet, so we're not particularly in a rush to switch to it even if templates become available soon (https://github.com/QubesOS/qubes-issues/issues/7807). I'd even go further and argue to hold off for a couple weeks anyway as new Fedora releases are often a bit bumpier of a ride when they're fresh.

But it will have to happen so here's the issue for it :+1:

eaon:
## Description of Changes

Fedora 37 will be shipped with Qubes OS R4.1.2. As soon as Qubes R4.1.2 has been released, this should probably be merged so that the next release doesn't download Fedora 36 unnecessarily.

Fixes #841

## Testing

- [ ] There are no references to Fedora 36 left in the repository
- [ ] `sd-fedora-37-dvm` is used in place of `sd-fedora-dvm` through the repo (except for the salt state removing the latter)
- Check out this branch in your sd-dev vm and ensure you [set the SDW environment variables](https://developers.securedrop.org/en/latest/workstation_setup.html#download-configure-copy-to-dom0). Ensure SD VMs (lookin at you, sd-whonix) are all shut down (otherwise, during `make test`, you could encounter tests that report that VMs aren't fully updated). In dom0 `make clone` then `make dev`
   - [ ] `make dev` completes without error
   - [ ] `sd-fedora-37-dvm` exists and is based on `fedora-37`
   - [ ] `sd-fedora-dvm` is removed
   - [ ] `make test` passes in dom0

Note that removing the f36 template/f36-dvm template is _not_ part of the template update process.

### DIFF
--- a/dom0/sd-sys-vms.sls
+++ b/dom0/sd-sys-vms.sls
@@ -61,10 +61,7 @@ set-fedora-default-template-version:
   {% set _ = required_dispvms.append("sd-" + sd_supported_fedora_version + "-dvm") %}
 {% endif %}
 
-{% set existing_vms = salt['cmd.shell']('qvm-ls --raw-list').split('\n') %}
-
 {% for required_dispvm in required_dispvms %}
-{% if required_dispvm not in existing_vms %}
 create-{{ required_dispvm }}:
   qvm.vm:
     - name: {{ required_dispvm }}
@@ -79,9 +76,7 @@ create-{{ required_dispvm }}:
 {% endif %}
     - require:
       - cmd: dom0-install-fedora-template
-{% endif %}
 {% endfor %}
-
 
 # Now proceed with rebooting all the sys-* VMs, since the new template is up to date.
 
--- a/dom0/sd-sys-vms.sls
+++ b/dom0/sd-sys-vms.sls
@@ -9,7 +9,7 @@ include:
   # DispVM is created
   - qvm.default-dispvm
 
-{% set sd_supported_fedora_version = 'fedora-36' %}
+{% set sd_supported_fedora_version = 'fedora-37' %}
 
 
 # Install latest templates required for SDW VMs.
@@ -61,7 +61,10 @@ set-fedora-default-template-version:
   {% set _ = required_dispvms.append("sd-fedora-dvm") %}
 {% endif %}
 
+{% set existing_vms = salt['cmd.shell']('qvm-ls --raw-list').split('\n') %}
+
 {% for required_dispvm in required_dispvms %}
+{% if required_dispvm not in existing_vms %}
 create-{{ required_dispvm }}:
   qvm.vm:
     - name: {{ required_dispvm }}
@@ -76,6 +79,7 @@ create-{{ required_dispvm }}:
 {% endif %}
     - require:
       - cmd: dom0-install-fedora-template
+{% endif %}
 {% endfor %}
 
 
@@ -94,7 +98,9 @@ create-{{ required_dispvm }}:
 {% else %}
   {% set sd_supported_fedora_template = sd_supported_fedora_version %}
 {% endif %}
-{% if salt['cmd.shell']('qvm-prefs ' + sys_vm + ' template') != sd_supported_fedora_template %}
+{% set vm_needs_updating = salt['cmd.shell']('qvm-prefs ' + sys_vm + ' template') != sd_supported_fedora_template %}
+{% set dvm_needs_updating = sys_vm == 'sys-usb' and 'sd-fedora-dvm' in existing_vms and salt['cmd.shell']('qvm-prefs sd-fedora-dvm template') != sd_supported_fedora_version %}
+{% if vm_needs_updating or dvm_needs_updating %}
 sd-{{ sys_vm }}-fedora-version-halt:
   qvm.kill:
     - name: {{ sys_vm }}
@@ -107,6 +113,17 @@ sd-{{ sys_vm }}-fedora-version-halt-wait:
     - require:
       - cmd: dom0-install-fedora-template
 
+# Since sys-usb uses sd-fedora-dvm, we can only change the latter's template
+# when it's not running
+{% if sys_vm == 'sys-usb' %}
+  qvm.vm:
+    - name: sd-fedora-dvm
+    - prefs:
+      - template: {{ sd_supported_fedora_version }}
+    - require:
+      - cmd: sd-{{ sys_vm }}-fedora-version-halt-wait
+{% endif %}
+
 sd-{{ sys_vm }}-fedora-version-update:
   qvm.vm:
     - name: {{ sys_vm }}
@@ -114,7 +131,7 @@ sd-{{ sys_vm }}-fedora-version-update:
       - template: {{ sd_supported_fedora_template }}
     - require:
       - cmd: sd-{{ sys_vm }}-fedora-version-halt-wait
-{% if sd_supported_fedora_template.endswith("-dvm") %}
+{% if not dvm_needs_updating and sd_supported_fedora_template.endswith("-dvm") %}
       - qvm: create-{{ sd_supported_fedora_template }}
 {% endif %}
 
--- a/dom0/sd-sys-vms.sls
+++ b/dom0/sd-sys-vms.sls
@@ -125,7 +125,7 @@ sd-{{ sys_vm }}-fedora-version-update:
 # We're numbering our VMs now even though its contents wouldn't change, to work
 # around Salt's limitations, so the non-numbered VM should be removed.
 {% if sys_vm == "sys-usb" %}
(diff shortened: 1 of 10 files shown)


eloquence:
(I've taken the liberty to push a small commit to update the `dom0` tests to the new template version.)

eaon:
The Fedora 37 template is available now so technically this is already ready for review.

eloquence:
I consistently get the following error when attempting to migrate from `fedora-36`:

```
qubesadmin.exc.QubesVMNotHaltedError: Cannot change template while there are running DispVMs based on this DVM template
```

To test the migration, I built an RPM from the branch, installed it in `dom0`, and then ran the updater with ` /opt/securedrop/launcher/sdw-launcher.py`, including once after a clean reboot. (I wanted to see the migration behavior a user would see.)

eloquence (changes requested):
Thanks for working through this! Unfortunately I'm still hitting an error during the migration. That said, it looks like the migration _did_ complete -- the non-disposable `sys-` templates are using fedora-37, and `sd-fedora-dvm` uses the new template. :tada: 

This Gist has the Salt error I'm getting: https://gist.github.com/eloquence/e5bca6fcd12ed6d7b7364c180bfe8bac

You may have an immediate hypothesis on what's causing this, but I'll leave one note at the line level on a part of the logic that's confusing to me.

dom0/sd-sys-vms.sls
```diff
@@ -107,14 +113,25 @@ sd-{{ sys_vm }}-fedora-version-halt-wait:
...
+
 sd-{{ sys_vm }}-fedora-version-update:
   qvm.vm:
     - name: {{ sys_vm }}
     - prefs:
       - template: {{ sd_supported_fedora_template }}
     - require:
       - cmd: sd-{{ sys_vm }}-fedora-version-halt-wait
-{% if sd_supported_fedora_template.endswith("-dvm") %}
+{% if not dvm_needs_updating and sd_supported_fedora_template.endswith("-dvm") %}
```
eloquence, line 134: I'm confused by the `not dvm_needs_updating` addition here -- what's the purpose of this additional check?

That said, what _appears_ to be happening on my system is that this evaluates to true, but the `create-` block itself is never defined, because the VM is already present and the `{% if required_dispvm not in existing_vms %}` block evaluates to false.

eaon:
I'm assuming you ran everything again without doing any manual Fedora 37 related cleaning operations, correct? If that's the case, your machine is in a state that the new Salt would not get to anymore, which is why this patch currently does not cover that situation. That being said, I think we shouldn't assume that SDW is used in a completely isolated environment, power users may have done some manual updating, so I'll take another look.

Just to spell out the permutations of cases we would run into:

- **Fresh install**: SDW is not installed, neither `fedora-37` nor `sd-fedora-dvm` exist
- **Regular upgrade**: SDW is installed, `fedora-37` does not exist, `sd-fedora-dvm` needs updating
- **Manual intervention or previous broken run**: SDW is installed, `fedora-37` exists, `sd-fedora-dvm` needs updating
- **All up-to-date**: SDW is installed, `fedora-37` exists, `sd-fedora-dvm` does not need updating

eloquence:
Yes, that's correct, I did not manually remove `fedora-37` before trying again. What's odd about my run, though, is that it seemed to eventually fit the "All up-to-date" description, but still kept failing with the error above.

The only other state I can think of is with Qubes 4.1.2, which was released last week. A fresh install on 4.1.2 would be:

-  Fresh install: SDW is not installed, fedora-37 exists but sd-fedora-dvm does not.

rocodes:
(Stepping through now to put some more eyes on this, will first try with no fedora-37 preinstall, then with fedora-37 template preinstalled)

eaon:
I pushed some changes after a note from @gonzalo-bulnes came to mind again.

I have not tested this, but I think this should cover all cases including the one @eloquence pointed out.

This is all significantly more complex than https://github.com/freedomofpress/securedrop-workstation/pull/802/files in part due what I've introduced with https://github.com/freedomofpress/securedrop-workstation/pull/764 and https://github.com/freedomofpress/securedrop-workstation/pull/784. I consider the numbered approach now taken by this PR a workaround as removing and recreating an AppVM with the same contents wastes time and resources. However, this is a restriction of our deployment strategy which is IMO in dire need of fundamental restructuring. More notes on the direction I think would remove classes of pain-points is documented in https://github.com/freedomofpress/securedrop-engineering/pull/22

zenmonkeykstop:
Given 4.1.2 is out and recommended, the fresh install case should actually be "SDW not installed, f37 exists, sd-fedora-dvm doesn't exist"

eloquence:
On my system which has `fedora-37` already installed, I am still encountering the error below. Note that `fedora-37-dvm` also exists on this system. This is why I think [this Jinja check](https://github.com/freedomofpress/securedrop-workstation/blob/5f55c279682a396b810769715a7a9586c7233200/dom0/sd-sys-vms.sls#L67) fails to include the `create-fedora-37` step, which is then later required. 

I'm going to try removing the `fedora-37-dvm` AppVM to see if that causes a successful run.

```
----------
          ID: sd-sys-firewall-fedora-version-update
    Function: qvm.vm
        Name: sys-firewall
      Result: False
                                 require:
                                     qvm: create-fedora-37-dvm
     Started: 18:33:02.926407
    Duration: 0.002 ms
     Changes:   
----------
          ID: sd-sys-firewall-fedora-version-start
    Function: qvm.start
        Name: sys-firewall
      Result: False
     Comment: One or more requisite failed: sd-sys-vms.sd-sys-firewall-fedora-version-update
     Started: 18:33:02.926714
    Duration: 0.003 ms
     Changes:   

```
However, I do end up

eloquence:
Yes, deleting `fedora-37-dvm` resulted in a successful run with all expected updates completed. :tada: 

If Qubes 4.1.2 ships with a `fedora-37-dvm` being created, I do think we have to update the Jinja logic to avoid failing on that case.

rocodes:
Thanks for the sleuthing - you're right, I thought initially you were encountering the issue after only having downloaded the f37 template (as opposed to having set up the dvm as well).  

This is fixable, thanks for flagging.

rocodes:
(A little bit annoying to remove the jinja optimization that was in there, but for now it's fine. We'll want to revisit our overall strategy here, as was [suggested earlier](https://github.com/freedomofpress/securedrop-workstation/pull/859#issuecomment-1479682157))

rocodes:
I'm stepping through one last time for good measure on a brand new 4.1.2 install. If someone wants to review on an existing 4.1 workstation by following the orignal test plan, since I'm the last one to push some code so I shouldn't review, that would be great.

cfm (approved):
Without *any* `fedora-37` yet installed, and with the caveat that this test plan wasn't written when I started the `make clone && sdw-admin --apply` sequence :-):

- Check out this branch in your sd-dev vm and ensure you [set the SDW environment variables](https://developers.securedrop.org/en/latest/workstation_setup.html#download-configure-copy-to-dom0). ~Ensure SD VMs (lookin at you, sd-whonix) are all shut down (otherwise, during `make test`, you could encounter tests that report that VMs aren't fully updated).~ In dom0 `make clone` then ~`make dev`~ `sdw-admin --apply`
   - [x] ~`make dev`~ `sdw-admin-apply` completes without error
   - [x] `sd-fedora-37-dvm` exists and is based on `fedora-37`
   - [x] `sd-fedora-dvm` is removed
   - [ ] `make test` passes in dom0
       * ...modulo qrexec policy changes I've made because this is my lowercase-W workstation too...  :-)
       * ...with one meaningful exception:

https://github.com/freedomofpress/securedrop-workstation/blob/d7045683aec8f0437841428fe3d53fdf26dd9436/tests/test_vms_exist.py#L9
```sh-session
[user@dom0 securedrop-workstation]$ qvm-run --pass-io sd-app uname -r
5.15.89-grsec-workstation
```

rocodes:
@cfm I still have the 5.15.41 grsec kernel when I've only built from dev on the SDW laptop. I'm guessing you had this machine on staging or prod before, since we don't do nightly kernel builds - that would explain the test failure.

Thank you for the review :) I will finish clean install QA then merge.

eloquence (approved):
Didn't do a full re-run, but did do another `make dev` with latest changes and all LGTM, so adding another +1 for good measure :)

rocodes:
Thanks both. Tested on fresh 4.1.2 install; fedora 37 template handled just fine, although the first time, the installation errored out because our own template (workstation-bullseye) was not successfully downloaded.  Will keep an eye out for that issue in QA, but I think we're OK to proceed here.

PR #864, opened 2023-03-28
Add -r to xargs invocation in case no .rpm files are found

maeve-fpf:
## Status

Ready for review

## Description of Changes

If no matching `.rpm` files exist, the `find | xargs` in `scripts/prep-dev` will run `ls` with no arguments, listing files in the current directory, and then attempt to install whatever other file showed up there as most recently modified as an RPM. This adds `-r` so that there will be no output, and the test to bail out in that case works as intended.

## Testing

Remove any RPMs in `rpm-build/RPMS` and try running `./scripts/prep-dev`. It should print `Could not find RPM!` and exit with failure, instead of attempting to run `rpm` on `config.json` or the like.

## Deployment

N/A, dev-related change.


rocodes (approved):
Thank you @maeve-fpf, tested and works as intended :)

PR #867, opened 2023-04-03
Bump version to 0.8.0-rc1

rocodes:
## Status

Ready for review

## Description of Changes

Bump version to rc1, add rc1 changelog 

- [x] Visual review makes sense
- [x] Base branch is release/0.8.0 not main ;) 
- [x] CI passing


rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -15,13 +15,13 @@
 
 Name:		%{srcname}
 Version:	%{version}
-Release:	1%{?dist}
+Release:	0.rc1.1%{?dist}
```
legoktm, line 18: It seems like this is how we've always been doing it, but I think the correct way is:
```
%global version 0.8.0~rc1
Release: 1%{?dist}
```

(see https://rpm-packaging-guide.github.io/#what-is-a-spec-file where Version is "The upstream version number of the software." and Release is just a plain integer)

I'll file a bug, not something to block on

legoktm (approved):

Issue #866, opened 2023-03-29
Release securedrop-workstation-dom0-config-0.8.0


rocodes:
This is a tracking issue for the release of the securedrop-workstation-dom0-config-0.8.0 RPM. This release moves the base Fedora template from Fedora 36 to Fedora 37.

## Roles:
- RM: @rocodes 
- Deputy RM: @legoktm 
- Comms: handled by RM/deputy RM
- QA: @cfm, @eloquence (?), others, as needed

### Pre-QA:
- [x] last QA+ merge of [F37 PR](https://github.com/freedomofpress/securedrop-workstation/pull/859)
- [x] Create test plan
- [x] Create release branch

### RC1 QA: 
- [x] [In release branch] bump rc version using `update_version` script, update changelog in rpm .spec file
- [x] [In release branch] create rc tag (signed with a maintainer's key)
- [x] [In release branch] build rc, commit build logs, and publish artifact to yum-test
- [x] QA; if additional RCs needed, repeat this section, amending test plan as needed  

### Release:
- [x] [In release branch] Bump version via `update_version` script, and update [changelog in rpm .spec file](https://github.com/freedomofpress/securedrop-workstation/blob/main/rpm-build/SPECS/securedrop-workstation-dom0-config.spec)
- [x] [In release branch]  Create prod tag (signed by release signing key)
- [x] Build prod artifact, sign with release key, commit build logs 
- [x] QA/smoketest of prod artifact (stuff rpm in dom0 of SDW prod machine)
- [x] Publish prod artifact to yum repo  

### Post-release
- [ ] Backport version and changelog updates to main branch from release branch
- [ ] Update [release docs](https://developers.securedrop.org/en/latest/workstation_release_management.html#release-securedrop-workstation-dom0-config) as needed 
- [x] Retro if/as needed

rocodes:
Test Plan
----------------------
### Clean install scenario (staging package)
Prep: Install Qubes 4.1.2. Provision workstation with `make staging`. Update workstation.
When package is pushed to staging, run updater.
  - [ ] Updater completes successfully
  - [ ] dom0-ws-config package is 0.8.0-rc1
  - [ ] sd-fedora-37-dvm created and assigned to sys-usb
  - [ ] all vms boot / make test in dom0 passes

### Upgrade scenario (staging)
Prep: Existing up-to-date SDW staging setup using fedora-36 vms.
When package is pushed to staging, run updater.
  - [ ] Updater completes successfully
[...]

PR #871, opened 2023-04-04
[0.8.0] Force a full migration for Fedora 37 template upgrade

legoktm:
## Status

Ready for review ?

## Description of Changes

The files needed for USB auto-attachment are only copied into sd-fedora-37-dvm during a full `sdw-admin --apply` run, which doesn't happen by default because it's slow.

Set a flag file to tell the updater to do a full migration. This is basically the same as was used in d4a8d1ba60f7.

Refs <https://github.com/freedomofpress/securedrop-workstation/issues/866#issuecomment-1496205177>.

## Testing

## Deployment

2. New installs

## Checklist

- [x] CI passes
- [x] Target branch set to `release/0.8.0`


rocodes:
Thank you @legoktm, will be folded into RC QA process

rocodes (approved):

PR #875, opened 2023-04-05
Make `gpg` tests more resilient

### LINKED ISSUE #874: Make gpg key test resilient
Our `gpg` test is currently pretty naive and fails easily if there is more than one key present in `sd-gpg`, which can happen during repeated provisioning runs and shouldn't cause a test failure if the expected key is present:

https://github.com/freedomofpress/securedrop-workstation/blob/main/tests/test_gpg.py#L20

eloquence:
## Status

Ready for review 

## Description of Changes

Fixes #874 

- Tests no longer fail if more than one key present
- Test fails if no key present in local keyring

## Testing

- Prior to testing, `make clone` from this branch in `dom0`. 
- Ensure you have a configured workstation and `sd-journalist.sec` and `config.json` present in `dom0`.

- [x] `make test-gpg` passes (if not, initial setup may have issues)
- Replace `dom0` `sd-journalist.sec` with a different key not provisioned in `sd-gpg`
- [x] `make test-gpg` fails
- Import the new key manually on `sd-gpg` via `gpg --import`
- [x] `make test-gpg` passes
- Remove the new key on `sd-gpg` via `gpg --delete-secret-key` and `gpg--delete-key`
- [x] `make test-gpg` fails
- Restore original `dom0` config
- [x] `make test-gpg` passes

### DIFF
--- a/tests/test_gpg.py
+++ b/tests/test_gpg.py
@@ -6,8 +6,9 @@ import tempfile
 from base import SD_VM_Local_Test
 
 
-def find_fp_from_gpg_output(gpg):
+def find_fps_from_gpg_output(gpg):
 
+    fingerprints = []
     lines = gpg.decode("utf-8").split("\n")
 
     for line in lines:
@@ -18,7 +19,9 @@ def find_fp_from_gpg_output(gpg):
         m = re.match(regex, line)
         if m:
             fp = m.groups()[1]
-            return fp
+            fingerprints.append(fp)
+
+    return fingerprints
 
 
 def get_local_fp():
@@ -31,15 +34,21 @@ def get_local_fp():
             stderr=subprocess.DEVNULL,
         )
         results = subprocess.check_output(["gpg", "-k", "--with-fingerprint"], env=gpg_env)
-        return find_fp_from_gpg_output(results)
+        fingerprints = find_fps_from_gpg_output(results)
+        if len(fingerprints) == 1:
+            return fingerprints[0]
 
 
-def get_remote_fp():
+def get_remote_fp(expected_fp=None):
     cmd = ["qvm-run", "-p", "sd-gpg", "/usr/bin/gpg --list-secret-keys --fingerprint"]
 
     p = subprocess.check_output(cmd)
 
-    return find_fp_from_gpg_output(p)
+    fingerprints = find_fps_from_gpg_output(p)
+
+    if expected_fp and expected_fp in fingerprints:
+        return expected_fp
+
 
 
 class SD_GPG_Tests(SD_VM_Local_Test):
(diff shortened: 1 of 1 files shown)


tests/test_gpg.py
```diff
@@ -31,15 +31,24 @@ def get_local_fp():
             stderr=subprocess.DEVNULL,
         )
         results = subprocess.check_output(["gpg", "-k", "--with-fingerprint"], env=gpg_env)
-        return find_fp_from_gpg_output(results)
+        fingerprints = find_fps_from_gpg_output(results)
+
+        # Because we imported this key into a temporary directory,
+        # we should only have one key in the keyring.
+        if len(fingerprints) == 1:
```
legoktm, line 38: Is there a reason you didn't make this an assertion? Otherwise it seems like there needs to be an else case here that throws an exception or pytest.fails

tests/test_gpg.py
```diff
@@ -31,15 +31,24 @@ def get_local_fp():
...
+def get_remote_fp(expected_fp):
     cmd = ["qvm-run", "-p", "sd-gpg", "/usr/bin/gpg --list-secret-keys --fingerprint"]
 
     p = subprocess.check_output(cmd)
 
-    return find_fp_from_gpg_output(p)
+    fingerprints = find_fps_from_gpg_output(p)
+
+    # Especially during development, sd-gpg may contain more than one key
+    if expected_fp and expected_fp in fingerprints:
```
legoktm, line 50: assert or else here as well

tests/test_gpg.py
```diff
@@ -52,7 +61,11 @@ def test_sd_gpg_timeout(self):
         self.assertFileHasLine("/home/user/.profile", line)
 
     def test_we_have_the_key(self):
-        self.assertEqual(get_local_fp(), get_remote_fp())
+        local_fp = get_local_fp()
+        remote_fp = get_remote_fp(expected_fp=local_fp)
+
+        self.assertIsNotNone(local_fp, "Local key not found")
```
legoktm, line 67: This covers what I mentioned above but I think it would be better for the assertions to be closer to what they are asserting.

legoktm (changes requested):
(per inline comments)

eloquence:
Thanks Kunal, will pick this back up but need to reprov a 4.1 env first -- doing that part today, so may not get back to the PR til tomorrow.

eloquence:
From-scratch 4.1 reprovisioning completed without issues or errors, will poke this tomorrow :)

eloquence:
Thanks @legoktm - I've reorganized the file a bit, tell me what you think. I did not see a good reason to have the GPG test setup outside the class. I've still tried to group the assertions in the test method but avoided the use of conditionals in the test helper.

zenmonkeykstop (approved):
LGTM, test plan passes

PR #877, opened 2023-04-05
Don't verify a specific kernel version

### LINKED ISSUE #872: Make sure `make test` expects correct kernel package depending on build flavour
Currently `make test` in a dom0 staging setup can produce failures when the workstation [grsec kernel version on apt-test](https://apt-test.freedom.press/pool/main/l/linux-upstream/) is newer than expected.

legoktm:
## Status

Ready for review

## Description of Changes

We are now shipping kernels more frequently, so it's not useful to assert an exact kernel version is being used, continually needing to be manually updated.

As of <https://github.com/freedomofpress/securedrop/commit/9cd5e8d107a1997ffcb9e061113932911ab20bfe> in SecureDrop server, we only verify that the running kernel is using the correct grsec flavor, which is what I've updated this test to match.

Fixes #872.

## Testing

* [ ] Visual review
* [ ] If desired, do a `make test` run, should have no failures (or at least this test should no longer be failing!)

## Deployment

Any special considerations for deployment? No, test only.

## Checklist

### DIFF
--- a/tests/test_vms_exist.py
+++ b/tests/test_vms_exist.py
@@ -6,7 +6,6 @@ from base import WANTED_VMS
 
 
 DEBIAN_VERSION = "bullseye"
-EXPECTED_KERNEL_VERSION = "5.15.41-grsec-workstation"
 
 
 class SD_VM_Tests(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


cfm (approved):
Looks good, and `make test` shows that the `SD_VM_Tests` pass, including individual tests' calls to `SD_VM_Tests._check_kernel()`.  Thanks, @legoktm!

PR #883, opened 2023-05-16
Don't allow running sdw-admin.py as root

### LINKED ISSUE #861: sdw-admin should error out when running as root
`sdw-admin --apply` will fail if run as root, so let's error out early.

mig5:
## Status

Ready for review

## Description of Changes

Fixes #861 

Changes proposed in this pull request: simple low-hanging fruit to prevent `sdw-admin.py` from being run as root.

(This was a useful change to test my CI environment after migrating it to Fedora 37)

### DIFF
--- a/files/sdw-admin.py
+++ b/files/sdw-admin.py
@@ -134,6 +134,9 @@ def perform_uninstall(keep_template_rpm=False):
 
 
 def main():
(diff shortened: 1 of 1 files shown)


rocodes (approved):
LGTM, tested in dom0 - thank you @mig5 :)

Issue #884, opened 2023-05-17
[Qubes 4.2] Test an SDW installation against development build


rocodes:
Tracking issue to document testing clean install of SDW against Qubes 4.2 build

Known to-dos:
- [ ] Build fedora-37 version of dom0 packages, including the base template
- [ ] Split templates into their own repo

Potential to-dos:
- [ ] Migrate to Debian Bookworm

eloquence:
I've begun installation of a weekly from https://qubes.notset.fr/iso-testing/ (per [this thread](https://forum.qubes-os.org/t/qubes-os-4-2-signed-weekly-builds/16929)) on a T480. Make sure you have the `qubes-weekly-builds-signing-key.asc` from the same directory (fingerprint should be `9B7E61D3BB70C4B1335CE5B67B72A119CCCA57BB`) installed if you want to verify the signature.

Will poke a bit more tomorrow.

eloquence:
First observations:
- 4.2 uses `fedora-37` as the base template in `dom0`, so as may be expected we'll need to build for a new architecture. The `fedora-32` RPM is installable in `dom0`, however.
- Seeing a lot of very disruptive `setuptools is replacing distutils` deprecation warnings all over the place, not sure yet if that's due to anything on our end or if it's just the Python env in `dom0`
- Install aborts early because `fedora-37` template is not yet available in https://yum.qubes-os.org/r4.2/templates-itl/rpm/, will try installing it from `testing` or patch to downgrade to fedora-36

eloquence:
Enabling the testing repo in `/etc/qubes/repo-templates/qubes-templates.repo` resolved the issue of installing `fedora-37`; however, our f32 SDW template does not appear to be installable via `qvm-template` after https://yum.securedrop.org/workstation/dom0/f32/ is configured as a repo.

The template installation stalls and fails with `Error canonicalizing file: failed to fill whole buffer`. This occurs during the use of `rpmcanon`, a Rust tool Qubes uses to mitigate security risks from installing RPMs (https://github.com/QubesOS/qubes-rpm-oxide/blob/main/rpm-writer/bin/rpmcanon.rs and https://github.com/QubesOS/qubes-issues/issues/6485).

This could also simply be because our existing published template is only built for `f32` right now, but CC-ing @marmarek in case there other potential reasons for this error that are worth investigating.

marmarek:
In theory, it should work I think. @DemiMarie, any idea?

[...]

PR #896, opened 2023-06-01
Check if template is installed before trying to install it

eaon:
## Description of Changes

Addresses https://github.com/freedomofpress/securedrop-workstation/issues/884#issuecomment-1570597339

This allows an installation to succeed if a template was installed manually even if a template isn't available from `qvm-template` repositories. Applies to both system and `securedrop-workstation-*` templates.

## Testing

- Stage 1, fresh install
   - [ ] Installation succeeds as normal
   - [ ] Running the updater succeeds
- Stage 2, fresh install
  - ~Manually rename `fedora-37` template to a name of your choice, change `sd_supported_fedora_version` to that name~
  - [ ] ~Installation succeeds as normal, despite that template name not referencing available templates in Qubes' or our repositories~
  - [ ] Manual installation of a template from disk (per [#884](https://github.com/freedomofpress/securedrop-workstation/issues/884#issuecomment-1574154240) before SDW installation allows for successful install

### DIFF
--- a/dom0/sd-dom0-files.sls
+++ b/dom0/sd-dom0-files.sls
@@ -68,7 +68,7 @@ dom0-workstation-templates-repo:
 dom0-install-securedrop-workstation-template:
   cmd.run:
     - name: >
-        qvm-template install securedrop-workstation-{{ sdvars.distribution }}
+        qvm-template info --machine-readable securedrop-workstation-{{ sdvars.distribution }} | grep -q "installed|securedrop-workstation-{{ sdvars.distribution }}|" || qvm-template install securedrop-workstation-{{ sdvars.distribution }}
     - require:
       - file: dom0-workstation-rpm-repo
 
--- a/dom0/sd-sys-vms.sls
+++ b/dom0/sd-sys-vms.sls
@@ -16,7 +16,7 @@ include:
 dom0-install-fedora-template:
   cmd.run:
     - name: >
-        qvm-template install {{ sd_supported_fedora_version }}
+        qvm-template info {{ sd_supported_fedora_version }} | grep -q Installed || qvm-template install {{ sd_supported_fedora_version }}
 
 # Update the mgmt VM before updating the new Fedora VM. The order is required
 # and listed in the release notes for F32 & F33.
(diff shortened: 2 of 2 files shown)


rocodes:
Sorry for the hold-up; it's a small change but I wanted to try out a couple edge cases with how qvm-template reports back just to make sure what we expect of qvm-template is what's happening.

- Template previously installed then removed: `qvm-template` correctly reports it as only `available`
- Template upgraded in place from $oldver to $newver but no changes to template name: `qvm-template info --machine-readable $oldver | grep installed` still shows `oldver` as installed
- Template upgraded in place from `$oldver` to `$newver` and then template renamed to `$newver`: `qvm-template --info` shows _neither_ `$oldver` or `$newver` as installed in the machine-readable output. (This is also true if you rename the template to a random name, see test plan below)
- Template upgraded in place, renamed to random name etc, and then renamed back to original name (`$oldver`) -- `qvm-template info` shows it as `installed` again.
- Template renamed to a random name (test plan):  renamed a fedora template to `fedora-thirtynonsense` then ran `qvm-template info --machine-readable fedora-thirtynonsense` and got `qvm-template: error: No matching templates to list` and a usage message.

So tl;dr people shouldn't be renaming templates or migrating in place, which is good/according to our advice, and if someone upgrades a sys template in place instead of downloading a new one, they shouldn't be able to proceed (which is ideal, also lines up with to our advice- we'd prefer freshly-downloaded sys template to upgraded-in-place ones).  Other than wondering why the discrepancy (between my xp with renaming and the one as described in the test plan), no issues with this change :)

rocodes (commented):
(see [comment](https://github.com/freedomofpress/securedrop-workstation/pull/896#issuecomment-1629869551))

rocodes (approved):
No changes to the PR, but I have modified the test plan to indicate that renaming templates may have unexpected behaviour and is something we should discourage (along with upgrading the templates in place, for our own purposes anyway). These instructions can go into our docs an/or ideally make it into our general installation acceptance testing / QA scenarios.  But otherwise the change wfm and will probably save us a headache or two down the line (like all your work on this project). Thanks @eaon <3  :shipit:

PR #908, opened 2023-07-19
Clarify closed beta notice and add it to installer

eloquence:
## Status

Ready for review 

## Description of Changes

Clarifies warning in README and adds a similar (interactive) warning to the installer script. This does add a speedbump to the installation process, which is IMO appropriate, but if it's annoying during development, we could add a `--skip-warning` flag, or just suppress it if the environment is set to staging or dev.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ Qubes uses SaltStack internally for VM provisionining and configuration manageme
 
 Installing this project is involved. It requires an up-to-date Qubes 4.1 installation running on a machine with at least 16GB of RAM (32 GB recommended).
 
-The project is currently in a closed beta, and we do not recommend installing it for production purposes. If you are participating in our beta program, please consult our *[end user documentation](https://workstation.securedrop.org)* for detailed setup instructions, and do not hesitate to reach out for assistance.
+The project is currently in a closed beta, and we do not recommend installing it independently for production purposes. If you are participating in our beta program, please consult our *[end user documentation](https://workstation.securedrop.org)* for detailed setup instructions, and do not hesitate to reach out for assistance. If you are interested in joining the beta, please contact us via securedrop@freedom.press or our support portal.
 
 To install a development version (using test data on the server and a test encryption key to decrypt it), in summary, you will need to:
 
--- a/files/sdw-admin.py
+++ b/files/sdw-admin.py
@@ -133,15 +133,29 @@ def perform_uninstall(keep_template_rpm=False):
     )
 
 
+def show_warning():
+    print(
+        "This script will provision the SecureDrop Workstation.\n\n"
+        "The SecureDrop Workstation is currently in a closed beta.\n"
+        "We do not recommend installing it in production independently until\n"
+        "the project has reached general availability.\n\n"
+        "If you are not part of the beta program, but are interested in joining,\n"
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -83,7 +83,7 @@ Qubes uses SaltStack internally for VM provisionining and configuration manageme
 
 Installing this project is involved. It requires an up-to-date Qubes 4.1 installation running on a machine with at least 16GB of RAM (32 GB recommended).
 
-The project is currently in a closed beta, and we do not recommend installing it for production purposes. If you are participating in our beta program, please consult our *[end user documentation](https://workstation.securedrop.org)* for detailed setup instructions, and do not hesitate to reach out for assistance.
+The project is currently in a closed beta, and we do not recommend installing it indepependently for production purposes. If you are participating in our beta program, please consult our *[end user documentation](https://workstation.securedrop.org)* for detailed setup instructions, and do not hesitate to reach out for assistance. If you are interested in joining the beta, please contact us via securedrop@freedom.press or our support portal.
```
rocodes, line 86: [nit: indepependently -> independently]

I'm not sure about adding this line on joining the beta. Don't think we have sufficiently talked about rollout strategy to be ready for that.

rocodes (commented):
Hey @eloquence, thank you for suggesting this change.

I think we should discuss this the next time we talk about our install story and/or meet as a team, since I don't think we are actually positioned to invite more support / install inquiries yet without a few more milestones under our belt, which to me is the natural reading of this kind of wording.  (I also have an admitted bias against the interaction pattern of "warning [select enter to continue anyway]" because I think it encourages ignoring warnings :) )

README.md
eloquence: Yeah, I'm ambivalent as well. However, I think an email to us is probably better than an org or individual setting up the system without ever talking to us -- and it also creates opportunities for us to engage in needs assessment and provide advice.

Perhaps a middle ground could be:
- in the README, a more general "discuss your needs and learn more" invitation
- in the installer, a more explicit "we advise against installing without talking to us" warning.

What do you/others think?

eaon: > and it also creates opportunities for us to engage in needs assessment and provide advice

I think there's a capacity question in there, that is I don't think we have enough to do _all the things_ at once. For that reason I'd prefer not adding any invitation.

files/sdw-admin.py
```diff
@@ -133,15 +133,29 @@ def perform_uninstall(keep_template_rpm=False):
...
+def show_warning():
+    print(
+        "This script will provision the SecureDrop Workstation.\n\n"
+        "The SecureDrop Workstation is currently in a closed beta.\n"
+        "We do not recommend installing it in production independently until\n"
+        "the project has reached general availability.\n\n"
+        "If you are not part of the beta program, but are interested in joining,\n"
+        "please contact us at securedrop@freedom.press or via our support portal."
+    )
+    input("Press [Enter] to proceed…")
```
eaon, line 145: I think this would keep our custom testing CI from running right now. I don't mind adding it but I think we'll need a `--skip-warning` argument along with this, and I guess changes in the CI as well?

eloquence:
Upon consideration & discussion, gonna close this out for now (easy to re-create if we want to) and just do a docs-only PR for now.

PR #913, opened 2023-08-03
Pull fedora:32 image from Fedora's own registry, always

### LINKED ISSUE #912: "fedora:32" is an ambigous container name
docker and podman interpret "fedora:32" differently. 

docker pulls it from dockerhub (https://hub.docker.com/_/fedora):
```
user@debian-test:~$ docker pull fedora:32
32: Pulling from library/fedora
de2fa1c557bb: Pull complete 
Digest: sha256:a40f00021f4051568e45e73b669b696ee9ceaced8deb4afc7cc431d3d6cb8ac8
Status: Downloaded newer image for fedora:32
docker.io/library/fedora:32
```

while podman aliases it to registry.fedoraproject.org/fedora:
```
user@dev ~> podman pull fedora:32
Resolved "fedora" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull registry.fedoraproject.org/fedora:32...
Getting image source signatures
Copying blob 480478cefb5d done  
Copying config 27a9790209 done  
Writing manifest to image destination
Storing signatures
27a9790209529e14132d05f032228f339a4fe1dd1e93a588f5af2c06c0234969
```

The two images are supposedly maintained by the same people but are different? :|

Anyways, we should explicitly specify which registry we want to pull from. My preference would be to get them from Fedora's own registry.

legoktm:
## Status

Ready for review

## Description of Changes

docker and podman interpret "fedora:32" in different ways, docker pulls it from Docker Hub and podman gets it from
registry.fedoraproject.org.

Despite presumably being maintained by the same people, those two 32 images are different. Let's be explicit and have docker use Fedora's own registry too.

Fixes #912.

## Testing

* [x] Visual review
* [x] CI passes

## Deployment

Any special considerations for deployment? No

### DIFF
--- a/bootstrap/Dockerfile
+++ b/bootstrap/Dockerfile
@@ -1,4 +1,4 @@
-FROM fedora:32
+FROM registry.fedoraproject.org/fedora:32
 LABEL org="Freedom of the Press"
 LABEL image_name="securedrop-workstation-qubes-4.1"
 


eloquence (approved):
LGTM; was able to build same RPM with Docker & podman with this change in place.

PR #922, opened 2023-11-27
Just set `Version` in RPM spec

legoktm:
## Status

Ready for review

## Description of Changes

This is mostly a copy of @eaon's change in securedrop-updater[1], with one key change of normalizing the version down to a PEP 440 compatible format.

Since we use setuptools's sdist to create a tarball and then expect rpmbuild to be able to find it, we need to use a common version format across both setuptools and RPM. Encoding the RC version in the `Release` field has always been wrong, since that's used for changes to packaging, not when the upstream (i.e. sdist tarball) changes.

This also means we no longer need to update `Source` in the release branch to point at the tarball, since it can be inferred correctly from the existing `%{name}-%{version}.tar.gz` macros.

Leave a TODO to switch to rpmdev-bumpspec which is the RPM `dch` equivalent to bump the version and insert a changelog entry.

[1] https://github.com/freedomofpress/securedrop-updater/commit/42a55c7f10546710e99fa73ae291a69b04dc4eeb

## Testing

* Run `./update_version.sh` with `0.9.0-rc1`, run `make rpm-build` and you get a package of what you expect
* Repeat with `0.9.0`

## Deployment

Any special considerations for deployment? Applies to everything

## Checklist

- [x] I would appreciate help with the documentation

### DIFF
--- a/update_version.sh
+++ b/update_version.sh
@@ -3,55 +3,16 @@
 
 set -e
 
-readonly NEW_VERSION=$1
-
-if [ -z "$NEW_VERSION" ]; then
+if [ -z "$1" ]; then
   echo "You must specify the new version!"
   exit 1
 fi
 
-OLD_VERSION=$(cat VERSION)
+# We want the Python and RPM versions to match, so we'll use a PEP 440
+# compatible version, e.g. 0.9.0rc1 or 0.9.0.
+NEW_VERSION=$(echo $1 | sed 's/-//g' | sed 's/~//g' )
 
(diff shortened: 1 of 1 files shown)


zenmonkeykstop:
@rocodes this matches up with your recco elsewhere as well, does the change above work for you?

zenmonkeykstop (approved):
Ran through test plan and compared "0.9.0" and "0.9.0-rc1" RPMs via diffoscope for good measure. All good!

PR #924, opened 2023-11-28
Don't install dev deps in the RPM build container

### LINKED ISSUE #921: Un-audited dev dependencies shouldn't be installed in rpm-build container
When we switched to the new dev container setup, we also inadvertently started installing dev dependencies (from pypi, that are hash-pinned but not audited) into the rpm build container. We should not be doing this, the only stuff in the build container should be from system packages. This might require us to split the build container into two.

legoktm:
## Status

Ready for review

## Description of Changes

The RPM build container needs to be highly trustworthy, so it should only contain things from Fedora itself and any audited code. We don't audit development dependencies, so we shouldn't be installing them into the build container.

We split the container used by `./scripts/container.sh` into two, a base, build container and then a container layered on top with dev dependencies.

Functionally this should be a no-op since none of the dependencies are used at build time but it cuts down on the risk of malicious code injection.

Fixes #921.

## Testing

* [x] visual review
* [x] Run `make rpm-build` with PR, note the checksum, run without this PR, should get the same checksum.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ dev staging: assert-dom0 ## Configures and builds a dev or staging environment
 
 .PHONY: build-rpm
 build-rpm: ## Build RPM package
-	@$(CONTAINER) ./scripts/build-rpm.sh
+	USE_BUILD_CONTAINER=true $(CONTAINER) ./scripts/build-rpm.sh
 
 .PHONY: reprotest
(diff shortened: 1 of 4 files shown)


zenmonkeykstop (approved):
Visual review LGTM and checksums of packages from this branch and main are the same.
big :+1: from me.

PR #955, opened 2024-02-28
Push our own nightlies to securedrop-yum-test

legoktm:
## Status

Ready for review

## Description of Changes

Take this responsibility over from securedrop-builder.

Refs <https://github.com/freedomofpress/securedrop-builder/issues/482>.

## Testing

* [x] infra has set PUSH_TOKEN
* [x] demo run of workflow passes without pushing - see https://github.com/freedomofpress/securedrop-workstation/actions/runs/8654879721

## Deployment

### DIFF
--- a/.github/workflows/nightlies.yml
+++ b/.github/workflows/nightlies.yml
@@ -1,0 +1,68 @@
+name: Nightlies
+on:
+  schedule:
+    - cron: "0 6 * * *"
+  push:
+    branches:
+      - main
+
+# Only allow one job to run at a time because we're pushing to git repos;
+# the string value doesn't matter, just that it's a fixed string.
+concurrency:
+  group: "just-one-please"
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build-rpm:
+    runs-on: ubuntu-latest
+    container:
+      image: registry.fedoraproject.org/fedora:37
+    steps:
+      - run: dnf install -y make git
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: make install-deps
+      - name: Build RPM
+        run: |
+          git config --global --add safe.directory '*'
+          # Version format is "${VERSION}-0.YYYYMMDDHHMMSS.fXX", which sorts lower than "${VERSION}-1"
+          rpmdev-bumpspec --new="$(cat VERSION)-0.$(date +%Y%m%d%H%M%S)%{?dist}" rpm-build/SPECS/*.spec
+          make build-rpm
+      - uses: actions/upload-artifact@v4
+        id: upload
+        with:
+          name: rpm-build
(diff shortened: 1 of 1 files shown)


legoktm (commented):
Flagging updates needed now

legoktm:
Workflow seems ready; pending infra generating a PAT, requested at https://github.com/freedomofpress/infrastructure/issues/4749.

.github/workflows/nightlies.yml
```diff
@@ -0,0 +1,69 @@
+name: Nightlies
+on:
+  schedule:
+    - cron: "0 6 * * *"
+  push:
```
legoktm, line 5: This is based on https://github.com/freedomofpress/securedrop-client/pull/1960, which hasn't been merged yet.

legoktm:
Token has been set by infra, I tested and fixed the workflow and moved it to ready for review \o/

.github/workflows/nightlies.yml
eloquence: OK, so we expect to get both nightlies _and_ a new build everytime we see a push on `main`. The nightlies themselves are expected to trigger a double-build, because they themselves will kick off a push to `main`. Is that correct?

legoktm: > The nightlies themselves are expected to trigger a double-build, because they themselves will kick off a push to `main`. Is that correct?

No, because the nightlies just push to securedrop-yum-test's main and not securedrop-workstation's main.

legoktm: Or maybe I'm misinterpreting what you meant by "both nightlies _and_ a new build"

eloquence: No, you're right. The order of operations as I understand it would be:

1) Push occurs to `securedrop-workstation`
2) This triggers the `on: push` workflow here, causing a nightly build to be pushed to `securedrop-yum-test`
3) _That_ push kicks off the `on: push` cleanup job in `securedrop-yum-test` to preserve a maximum of 4 builds, which recursively triggers itself once with no expected side effect

And the same occurs on a nightly basis. Does that sound right?

legoktm: Yep!

eloquence: Excellent.The workflow LGTM, but I'll take another look on Monday for a formal review if nobody beats me to it. :)

eloquence (approved):
LGTM, let's give it a try. TIL about artifact actions.

legoktm:
> As a side note, should we split `make install-deps`? It's installing Qt-related dependencies -- are these actually needed for this workflow? If it makes sense to split, I'll open a separate issue to track.

Yeah, probably time to split it, as it's currently a mix of dev/test and build dependencies.

PR #967, opened 2024-03-27
WIP Intial Qubes 4.2 Updater Integration

### LINKED ISSUE #899: Investigate using Qubes updater in 4.2
Qubes 4.2 ships with a new GUI and CLI updater that no longer relies on Salt to update VMs. This means that our existing update method no longer works, but also presents an opportunity to more heavily rely on Qubes' built-in update facilities. The purpose of this issue is to:

- investigate and flesh out what a migration to direct use of the Qubes 4.2 updater could look like;
- identify decisions we need to make about the desired updater behavior
- cross-link upstream issues that could present blockers for such a migration.

*(Issue placeholder fleshed out by @eloquence, 2023-06-08)*

deeplow:
## Status
Work in progress. Based on 4_2-compat-deb12-base, only @deeplow's commits are relevant to this PR.

TODO:
  - [ ] Evaluate if error states are still appropriate
  - [ ] Add / fix tests
  - [ ] Consider if it makes sense to update even after a migration was applied (this was not the case before)

## Description of Changes

Fixes #899 . Keeps most of the existing updater logic.

- updates templates via the CLI Qubes Updater instead of via salt
- removes functions deprecated by the new Qubes updater

## Testing
1. `make dev`
2. launch updater
3. Observe templates starting as they are updated
4. Observe app qubes getting restarted in background
5. See "successfully completed"
6. Inspect SDW-launcher logs to ensure correct updating of templates.

## Deployment

Any special considerations for deployment? Will only apply to new installs

> **Note**: tests don't pass due to the WIP nature of the base 4.2 branch. However, it seems that launcher-specific tests weren't passing already and it was unclear how to run the tests. 


zenmonkeykstop:
@deeplow, I can't reopen this from here - I think you need to switch the base to `main` first.

legoktm:
I tried restoring the branch in case it would help - it did not. I think you'll just have to create a new PR for it. :|

PR #970, opened 2024-04-02
Update SecureDrop workstation to support 4.2 and use an install-time base template.

### LINKED ISSUE #969: Drop securedrop-workstation-bullseye base template, use a debian-12 system template
(initial attempt underway in https://github.com/freedomofpress/securedrop-workstation/tree/4_2-compat-deb12-base)

Pre-Qubes 4.2, SDW used a custom base templateVM for AppVMs, https://github.com/freedomofpress/qubes-template-securedrop-workstation . This was a maintenance burden, and now that qubes-builder is deprecated in favour of qubes-builderv2, we would need to start from scratch to stick with this approach.

An alternative would be to use the system debian-12-xfce template as a base at install time. This would reduce maintenance overhead and eliminate the 4 GiB custom template download at install time. One major downside would be a loss of control over the state of SDW AppVMs' template.

We can mitigate said downside in a couple of ways:
- Require that production SDW instances be installed on fresh Qubes installs (docs-only)
- Install and use another Qubes-provided template not installed with the OS, such as debian-12-minimal or debian-12-gnome.
- Test to see if the debian-12-xfce template is already in use at install time:
  - if so, cancel the install and instruct the user to use a fresh Qubes install
  - if not, uninstall and reinstall the template, resetting it (open question: can we do this without downloading it again?), then proceed with the install.

The last option is currently being prototyped.

zenmonkeykstop:
## Status

Work in progress:
- `sdw-admin --apply` command detects existing appVM dependencies on `debian-12-minimal` and alerts the user, but should also reinstall `debian-12-minimal` if necessary
- some tests failing - needs a grsec kernel with the expected identifier (including `-workstation-grsec`) and needs updates to the RPC checks.
- launcher tests are failing in CI

## Description of Changes

Fixes #969 .

## Testing

Much larger plan TK, but roughly
- [ ] Provision dev env via `make clone`, `make dev`, verify that the latter completes successfully
- [ ] reboot
- [ ] Verify that `make test` passes
- [ ] Verify that the launcher is triggered and update is successful
- [ ] Verify basic client functionality (log in, retrieve and view submissions,, reply, export/print)

## Deployment

Migration from 4.1 is not supported - fresh installs on 4.2 only.

vi 


Makefile
```diff
@@ -1,9 +1,11 @@
 DEFAULT_GOAL: help
-# We prefer to use python3.8 if it's availabe, as that is the version shipped
-# with Fedora 32, but we're also OK with just python3 if that's all we've got
-PYTHON3 := $(if $(shell bash -c "command -v python3.8"), python3.8, python3)
-# If we're on anything but Fedora 32, execute some commands in a container
-CONTAINER := $(if $(shell grep "Thirty Two" /etc/fedora-release),,./scripts/container.sh)
+# We prefer to use python3.11 if it's availabe, as that is the version shipped
+# with Fedora 37, but we're also OK with just python3 if that's all we've got
+PYTHON3 := $(if $(shell bash -c "command -v python3.11"), python3.8, python3)
```
legoktm, line 4:
```suggestion
PYTHON3 := $(if $(shell bash -c "command -v python3.11"), python3.11, python3)
```

Makefile
```diff
@@ -34,9 +36,12 @@ build-rpm: ## Build RPM package
 
 .PHONY: reprotest
 reprotest: ## Check RPM package reproducibility
-	TERM=xterm-256color $(CONTAINER) bash -c "sudo ln -s $$PWD/scripts/fake-setarch.py /usr/local/bin/setarch && sudo reprotest 'make build-rpm' 'rpm-build/RPMS/noarch/*.rpm' --variations '+all,+kernel,-fileordering,-domain_host'"
+	TERM=xterm-256color $(CONTAINER) bash -c "sudo ln -s $$PWD/scripts/fake-setarch.py /usr/local/bin/setarch && sudo reprotest 'make build-rpm' 'rpm-build/RPMS/noarch/*.rpm' --variations '+all,+kernel,-time,-fileordering,-domain_host'"
+	@echo
+	@echo Warning! Temporarily removed time variations for reprotest.
```
legoktm, line 41: This seems like a blocker. Or if its just reprotest/libfaketime we can switch to the approach of doing two builds and then diffing them.

dom0/sd-base-template-files.sls
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# vim: set syntax=yaml ts=2 sw=2 sts=2 et :
+include:
+  - fpf-apt-repo
+
+# TODO: test if this works, if not it may need to be a qvm-run command
+sd-base-template-install-additional-packages:
+  pkg.installed:
+    - pkgs:
+      - qubes-core-agent-passwordless-root
```
legoktm, line 10: All of these should be dependencies in -config or other packages. But I'm not sure why we're not starting with the GNOME (or even XFCE) templates to begin with, I thought that was what we had discussed initially?

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -53,6 +51,9 @@ configuration over time.
 
 %install
 %{python3} -m pip install --no-compile --no-index --no-build-isolation --root %{buildroot} .
+# direct_url.json is is not reproducible and not strictly needed
+rm %{buildroot}/%{python3_sitelib}/*%{version}.dist-info/direct_url.json
+sed -i "/\.dist-info\/direct_url\.json,/d" %{buildroot}/%{python3_sitelib}/*%{version}.dist-info/RECORD
```
legoktm, line 56: from the client:
```
override_dh_strip_nondeterminism:
	find ./debian/ -type f -name '*.pyc' -delete
	find ./debian/ -type d -name '__pycache__' -delete
	find ./debian/ -type f -name 'pip-selfcheck.json' -delete
	find ./debian/ -type f -name 'direct_url.json' -delete
	find ./debian/ -type f -name 'RECORD' -delete
	dh_strip_nondeterminism $@
```

let's just delete these instead of trying to fix them up

dom0/sd-base-template-files.sls
zenmonkeykstop: If we start with -gnome or -xfce, then the base template will include more than we want. The initial discussion was actually around using -minimal, I did flip to -xfce temporarily, but the resulting attack surface in terms of unused packages is too great, and removing them would be more hassle than starting with minimal and adding just what we need.

zenmonkeykstop: But I'm down to add them in the securedrop-workstation-config package.

legoktm:
Summary of the dom0 test failures:

* test_sd_client_apparmor - apparently not set to enforce
* test_log_dirs_properly_named / test_logs_are_flowing - unclear what the error is, but probably related to the next two
* test_redis_service_running - it's not running
* test_sd_log_service_running - it's not running
* test_sd_proxy_rpc_spec - file doesn't exist maybe?
* test_Policies - RPC policy doesn't exist
* test_sd_workstation_template - template doesn't exist
* test_all_fedora_vms_uptodate - `Command 'sudo dnf check-update' returned non-zero exit status 100.`
* test_gpg_domain_configured - returned '' instead of sd-gpg
* test_logging_configured, and a number of related dependency-related tests all saying not installed
* test_mime_types - selecting wrong opener
* _check_kernel - known that the kernel is reporting the wrong version
* test_all_sd_vm_apt_sources - missing apt source

dom0/sd-dom0-qvm-rpc.sls
```diff
@@ -146,10 +70,23 @@ dom0-rpc-qubes.r5-format-ask-allow:
         securedrop.Log          *           sd-log sd-log deny notify=no
         securedrop.Log          *           @tag:sd-workstation sd-log allow
 
+        securedrop.Proxy        *           sd-app sd-proxy allow
+
+        qubes.Gpg               *           @tag:sd-client sd-gpg allow
```
rocodes, line 75: (are we missing the allow rule for GpgImportKey?  `GpgImportKey * @tag:sd-client sd-gpg allow`)

dom0/sd-dom0-qvm-rpc.sls
```diff
@@ -146,10 +70,23 @@ dom0-rpc-qubes.r5-format-ask-allow:
...
+
+        qubes.USBAttach         *           sys-usb sd-devices allow  user=root
+        qubes.USBAttach         *           @anyvm @anyvm ask
+
+        qubes.USB               *           sd-devices sys-usb allow
+
+        # TODO: should this be handled with the new Global Config UI instead?
```
rocodes, line 82: I think this is appropriate for the policy file, so that the config is set by files we control (and if `uninstall` is run it will be removed cleanly)

dom0/sd-logging-setup.sls
```diff
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
 # vim: set syntax=yaml ts=2 sw=2 sts=2 et :
 
-{% if grains['id'] in ["securedrop-workstation-bullseye", "sd-small-bullseye-template", "sd-large-bullseye-template"] %}
+# TODO: parametrise this
+{% if grains['id'] in ["sd-small-bookworm-template", "sd-large-bookworm-template"] %}
```
rocodes, line 5: nit: if we stick to ` "sd-small-{}-template".format(grains['oscodename']) %`, as in below conditional, and set it in sdvars, we will have one source of truth for debian version name

dom0/securedrop-check-migration
```diff
@@ -22,7 +22,7 @@ function request_migration() {
 
 # Template consolidation. If old template names are found,
 # then we must rerun the full states to re-apply.
-if [[ -n "$(qvm-ls --tags sd-workstation --raw-list | perl -nE '/sd-(?!small|large).*-template/ and print $_')" ]] ; then
+if [[ -n "$(qvm-ls --tags sd-workstation --raw-list | perl -nE '/sd-(?!small|large|base).*-template/ and print $_')" ]] ; then
     reason="template-consolidation"
```
rocodes, line 26: Can we remove this, since we're starting clean on 4.2?

dom0/sd-dom0-files.sls
```diff
@@ -46,31 +46,11 @@ dom0-workstation-rpm-repo:
...
-        enabled=1
-        baseurl={{ sdvars.dom0_yum_repo_url }}
-        name=SecureDrop Workstation Templates repository
-    - require:
-      - file: dom0-rpm-test-key
-
-dom0-install-securedrop-workstation-template:
+# Ensure debian-12-minimal is present for use as base template
+dom0-install-debian-minimal-template:
   cmd.run:
```
rocodes, line 51: I _think_ we can just use `[qvm.template_installed](https://github.com/QubesOS/qubes-mgmt-salt-dom0-qvm?tab=readme-ov-file#qvmtemplate_installed)` instead of `cmd.run`.  If memory serves, we had an issue with the qvm-template tooling because we were using the same repo to serve templates and the dom0 config rpm, and qvm-template was upset by this.

legoktm:
I squashed it down from 41 commits to 11. 3 are annotated as FIXMES, I think we'd want to drop them before merging. I'll look into the rpm reproducibility issue tomorrow, and the shutdown VMs one can be dropped once https://github.com/freedomofpress/securedrop-workstation-ci/pull/44 is deployed. I'm not sure about the qvm-check one.

mig5:
> the shutdown VMs one can be dropped once [freedomofpress/securedrop-workstation-ci#44](https://github.com/freedomofpress/securedrop-workstation-ci/pull/44) is deployed.

https://github.com/freedomofpress/securedrop-workstation-ci/pull/44 is now merged and deployed to the runner bastion, should now be active for all future CI runs.

dom0/securedrop-check-migration
zenmonkeykstop: Yup, I think so.

dom0/sd-sys-vms.sls
```diff
@@ -117,18 +119,12 @@ sd-{{ sys_vm }}-fedora-version-update:
...
-  qvm.absent:
-    - name: sd-fedora-dvm
-    - require:
-      - qvm: sd-sys-usb-fedora-version-update
-
 # Finally, remove the old supported fedora DVM we created. We won't uninstall
 # the template, in case it's being used elsewhere, but the `sd-` VMs we can
 # reasonably manage (remove) ourselves.
+# TODO: does this need to be updated to refer to sd-fedora-<previous>-dvm? or was it
+# a one-time thing?
```
rocodes, line 126: Yes, if we're still going with `sd-fedora-xx-dvm` for the sys-usb dvm template, it would be friendly to clean up the old ones when we're not using them anymore.  But since we're installing clean on 4.2, users won't have an out-of-date sd-fedora dvm on their machines, so this won't be relevant til the next fedora base template upgrade.

files/provision-all
```diff
@@ -27,12 +36,17 @@ qvm-shutdown --wait sd-log && qvm-start sd-log
...
-qvm-check --quiet sd-fedora-38-dvm 2> /dev/null && \
-  sudo qubesctl --show-output --skip-dom0 --targets sd-fedora-38-dvm state.highstate && \
+# If sd-fedora-39-dvm exists it's because salt determined that sys-usb was disposable
+qvm-check --quiet sd-fedora-39-dvm 2> /dev/null && \
+  sudo qubesctl --show-output --skip-dom0 --targets sd-fedora-39-dvm state.highstate && \
   qvm-shutdown --wait sys-usb && qvm-start sys-usb || \
   sudo qubesctl --show-output --skip-dom0 --targets sys-usb state.highstate
+
+# Shut down all VMs to ensure new configuration takes place, if the user doesn't reboot.
+qvm-ls --tags sd-workstation --raw-list | xargs qvm-shutdown --wait
```
rocodes, line 52: You might want `qvm-shutdown --force --wait`, but you might decide this is too niche a usecase and opt to leave it as is. The use case would be: if the user had a running appvm that used sd-whonix for networking (eg they were checking the admin interface), then sd-whonix wouldn't reboot without the `force` command.

files/sdw-admin.py
```diff
@@ -139,11 +167,31 @@ def main():
...
-        print("Validating...")
+        print("Validating...", end="")
         validate_config(SCRIPTS_PATH)
+        print("OK")
     elif args.apply:
+        print(
+            "SecureDrop Workstation should be installed on a fresh Qubes OS install.\n"
+            "The installation process will overwrite any user modifications to the\n"
+            f"{BASE_TEMPLATE} TemplateVM, and will disable old-format qubes-rpc\n"
+            "policy directives.\n"
```
rocodes, line 178: I don't think this latter part is true anymore - I think we have decided that we won't disable them, we'll just make sure all our rules take precedence before the old-style rules can even load.

files/sdw-admin.py
```diff
@@ -139,11 +167,31 @@ def main():
         sys.exit(0)
     args = parse_args()
     if args.validate:
-        print("Validating...")
+        print("Validating...", end="")
         validate_config(SCRIPTS_PATH)
+        print("OK")
```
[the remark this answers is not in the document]
zenmonkeykstop: This and that sticky note saying PUSH I put on a glass door in 444 Front West are my crowning achievements in life.

files/sdw-admin.py
zenmonkeykstop: sdw-admin will get another pass to clean up output and add more stringent prod checks, we can bake that final decision in (and maybe check the rpc rules config as well as checking for VMs) when we do.

files/provision-all
legoktm: I removed this commit, https://github.com/freedomofpress/securedrop-workstation-ci/pull/44 uses force so not an issue, but if we want to tweak it further, probably should go in the CI repo.

legoktm:
I don't want to disuade people from pointing out things that can safely be removed, because I think there's a bunch, but I think we can defer them for follow-up PRs that will be trivial to merge?

----

In my testing locally, with the two commits I just pushed, the only test failures I have left are related to the kernel version. There are a few different failures remaining in CI that I wonder if they're related to the VM shutdown issue (but I can see they're being shut down in the log?)

zenmonkeykstop:
> I don't want to disuade people from pointing out things that can safely be removed, because I think there's a bunch, but I think we can defer them for follow-up PRs that will be trivial to merge?

Yup, I agree - I'd like to get a functional 4.2 main ASAP and work on fixes/major changes/tweaks, in that order.

legoktm:
We're down to two failing tests (excluding kernel version suffix):
```
INFO:[2024-04-04-00:02:31:790536] ======================================================================
INFO:[2024-04-04-00:02:31:790573] ERROR: test_all_fedora_vms_uptodate (test_vms_platform.SD_VM_Platform_Tests.test_all_fedora_vms_uptodate)
INFO:[2024-04-04-00:02:31:790610] Asserts that all Fedora-based templates, such as sys-net, have all
INFO:[2024-04-04-00:02:31:790647] ----------------------------------------------------------------------
INFO:[2024-04-04-00:02:31:790682] Traceback (most recent call last):
INFO:[2024-04-04-00:02:31:790718]   File "/home/user/securedrop-workstation/tests/test_vms_platform.py", line 174, in test_all_fedora_vms_uptodate
INFO:[2024-04-04-00:02:31:790752]     self._ensure_packages_up_to_date(vm, fedora=True)
INFO:[2024-04-04-00:02:31:790788]   File "/home/user/securedrop-workstation/tests/test_vms_platform.py", line 104, in _ensure_packages_up_to_date
INFO:[2024-04-04-00:02:31:790821]     stdout, stderr = vm.run(cmd)
INFO:[2024-04-04-00:02:31:790858]                      ^^^^^^^^^^^
INFO:[2024-04-04-00:02:31:790891]   File "/usr/lib/python3.11/site-packages/qubesadmin/vm/__init__.py", line 320, in run
INFO:[2024-04-04-00:02:31:790927]     raise e
INFO:[2024-04-04-00:02:31:790960]   File "/usr/lib/python3.11/site-packages/qubesadmin/vm/__init__.py", line 316, in run
INFO:[2024-04-04-00:02:31:790996]     return self.run_service_for_stdio('qubes.VMShell',
INFO:[2024-04-04-00:02:31:791029]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INFO:[2024-04-04-00:02:31:791064]   File "/usr/lib/python3.11/site-packages/qubesadmin/vm/__init__.py", line 296, in run_service_for_stdio
INFO:[2024-04-04-00:02:31:791098]     raise exc
INFO:[2024-04-04-00:02:31:791135] subprocess.CalledProcessError: Command 'sudo dnf check-update' returned non-zero exit status 100.
```

and

```
INFO:[2024-04-04-00:02:31:792645] ======================================================================
INFO:[2024-04-04-00:02:31:792681] FAIL: test_all_sd_vm_apt_sources (test_vms_platform.SD_VM_Platform_Tests.test_all_sd_vm_apt_sources)
INFO:[2024-04-04-00:02:31:792716] Test all VMs fpf apt source list iteratively.
INFO:[2024-04-04-00:02:31:792753] ----------------------------------------------------------------------
INFO:[2024-04-04-00:02:31:792787] Traceback (most recent call last):
INFO:[2024-04-04-00:02:31:792823]   File "/home/user/securedrop-workstation/tests/test_vms_platform.py", line 221, in test_all_sd_vm_apt_sources
INFO:[2024-04-04-00:02:31:792857]     self._validate_apt_sources(vm)
INFO:[2024-04-04-00:02:31:792893]   File "/home/user/securedrop-workstation/tests/test_vms_platform.py", line 83, in _validate_apt_sources
INFO:[2024-04-04-00:02:31:792927]     self.assertTrue(self.apt_url in contents)
INFO:[2024-04-04-00:02:31:792963] AssertionError: False is not true
```

I pushed some better error messages to surface which VMs are at fault. I suspect it's sys-net, etc. that have pending Fedora updates because we're not restarting them (since they don't have the sd-workstation tag). I also don't know if we can restart them, because we need a network connection to communicate with the ws-ci-bastion. If this is the case I would suggest we skip these tests if we're in the CI environment and leave a fixme for a better solution.

zenmonkeykstop:
One option (for the package updates at least) might be to trigger the cli updater as a first/early step in the install process. For the pilot we told folk to manually update before kicking things off, but if we can do it for them so much the better.

legoktm:
Also the VM shutdown we did in the CI repo doesn't seem sufficient, I think we might need to properly wait until they're shutdown. So we can probably ship the current FIXME commit (with a different summary) and revert it once CI is fixed.

dom0/sd-default-config.sls
```diff
@@ -17,7 +17,7 @@
 {% if d.environment == "dev" %}
   # use apt-test and nightlies
   {% set sdvars = sdvars_defaults["test"] %}
-  {% set _ = sdvars.update({"component": "nightlies"}) %}
+  {% set _ = sdvars.update({"component": "main nightlies"}) %}
```
cfm, line 20: This breaks: <https://github.com/freedomofpress/securedrop-workstation/blob/941f051f2e7fb627935dc661d01e7504b3b7c884/tests/test_vms_platform.py#L36-L37>

dom0/sd-dom0-files.sls
zenmonkeykstop: This can probably be an issue.

dom0/sd-logging-setup.sls
zenmonkeykstop: Yup that's what I meant.

legoktm:
> I'll look into the rpm reproducibility issue tomorrow

I forgot to report back that I did investigate the reproducibility issue - the rpms are reproducible *except* when built under reprotest (I suspect this is some libfaketime interaction gone wrong). I've updated the commit/comments to reflect this and don't consider it a blocker to moving forwards.

cfm (commented):
Just to stress-test things, uninstalling an existing installation on Qubes 4.2 from 4d6ca1a8cf329fccec8f2a81bca50e34ae45e0d0...5873c638d25094f734d5359c2f01005d360dde85 and then installing from here:

- [x] Provision dev env via `make clone`, `make dev`, verify that the latter completes successfully

```sh-session
[user@dom0 ~] sdw-admin --uninstall
[user@dom0 ~] cd securedrop-workstation
[user@dom0 securedrop-workstation] make clone
[user@dom0 securedrop-workstation] make dev
```

Failed because the signing key has changed per <https://github.com/freedomofpress/securedrop-workstation/pull/972>.  Apply that locally, then:

```sh-session
[user@dom0 securedrop-workstation] sdw-admin --uninstall  # our templates deleted: ✓
[user@dom0 securedrop-workstation] make clone
[user@dom0 securedrop-workstation] make dev
```

- [x] reboot

Updates, and reboot again.

- [ ] Verify that `make test` passes

Exceptions:

- review link TK
- VM kernels fail <https://github.com/freedomofpress/securedrop-workstation/blob/941f051f2e7fb627935dc661d01e7504b3b7c884/tests/test_vms_exist.py#L36>

- [x] Verify that the launcher is triggered and update is successful
- [ ] Verify basic client functionality (log in, retrieve and view submissions,, reply, export/print)

TK

dom0/sd-default-config.sls
zenmonkeykstop: the kernel string test failure will be addressed with next 6.6 kernels, @legoktm already fixed that in a PR pending to `kernel-builder`.

tests/test_qubes_rpc.py
```diff
@@ -1,50 +1,13 @@
...
-                print("Policy for {} should be:\n{}".format(filename, expectedPolicy))
-                print("\n\n#### END RPC policy error report ####\n\n")
-                return False
-
-    def _loadVars(self):
-        filepath = os.path.join(os.path.dirname(os.path.abspath(__file__)), "vars", "qubes-rpc.yml")
-        with io.open(filepath, "r") as f:
-            data = yaml.safe_load(f)
-        return data
+    def test_policies(self):
```
zenmonkeykstop, line 7: This is a good first step IMO

files/provision-all
```diff
@@ -47,3 +47,8 @@ qvm-check --quiet sd-fedora-39-dvm 2> /dev/null && \
   sudo qubesctl --show-output --skip-dom0 --targets sd-fedora-39-dvm state.highstate && \
   qvm-shutdown --wait sys-usb && qvm-start sys-usb || \
   sudo qubesctl --show-output --skip-dom0 --targets sys-usb state.highstate
+
+# Shut down all VMs to ensure new configuration takes place, if the user doesn't reboot.
```
zenmonkeykstop, line 51: If we actually need a reboot, we could probably change this to a user Y/N prompt - "The system must be rebooted to complete the installation. Reboot now? Y/N"  or similar. If we don't *actually* need a reboot then kicking the VMs is fine.

zenmonkeykstop (commented):
@legoktm's changes LGTM. I can't approve having created the PR tho, someone else will need to.

cfm (approved):
Approving based on all testing and review so far, per <https://github.com/freedomofpress/securedrop-workstation/pull/970#pullrequestreview-1990178359>.

PR #971, opened 2024-04-03
[WIP] reabsorb `securedrop-updater`

cfm:
## Status

Work in progress

## Description of Changes

Towards #966, this branch aims to replicate the head of `reabsorb-updater` without pulling in (via `git merge -X find-renames --allow-unrelated-histories`) the original divergence of `securedrop-updater` from this repository.  Preserving history, undoing that divergence, _and_ integrating with other work in progress towards Qubes 4.2 compatibility increasingly seem like incompatible goals, at least in practice.

Per <https://github.com/freedomofpress/securedrop-workstation/issues/966#issuecomment-2035435574>, for the version immediately *after* the first release with Qubes 4.2 compatibility, the team should make a go/no-go decision about adopting this work.  Remaining tasks include:
- defining the packaging boundaries (per `make test-install-rpm`); and
- defining the testing boundaries (TK).

## Testing

TK

## Deployment

Any special considerations for deployment? Consider both:

1. Upgrading existing pilot instances
2. New installs

### If you have made changes to the provisioning logic

- [ ] All tests (`make test`) pass in `dom0`

### If you have added or removed files

- [ ] I have updated `MANIFEST.in` and `rpm-build/SPECS/securedrop-workstation-dom0-config.spec`

### If documentation is required

- [ ] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-workstation-docs) for these changes, or will do so later
- [ ] I would appreciate help with the documentation

Issue #965, opened 2024-03-26
Add basic 4.2 compatibility


zenmonkeykstop:
Support for Qubes 4.2 was investigated in #898, which updated policies and split out the launcher script to a separate repo, among other changes. A new branch has been created cherry-picking desired changes from this PR, and reinstating the launcher as part of the securedrop workstation repo.

<details>
<summary>table of cherry-picks and rationales</summary>

original commit | status in new branch | notes |  
-- | -- | -- | --
https://github.com/freedomofpress/securedrop-workstation/pull/898/commits/052e5a6fd8255f8ed5b4d811b5082319f1cfc0d1 | NOT INCLUDED | superseded by https://github.com/freedomofpress/securedrop-workstation/commit/4d841fa3c096a3ef904bf1dbbc4701f516ef7bb0 |  
https://github.com/freedomofpress/securedrop-workstation/pull/898/commits/4d841fa3c096a3ef904bf1dbbc4701f516ef7bb0 | NOT INCLUDED | already in main |  
https://github.com/freedomofpress/securedrop-workstation/pull/898/commits/98b5bb3690bf4c83c519e863252d32c442c94f25 | INCLUDED | mostly does the bump to f37 |  
https://github.com/freedomofpress/securedrop-workstation/pull/898/commits/45341575a9f7ab5b48521f5d0ae6e5bf1a1b8528 | NOT INCLUDED | launcher no longer being split out |  
https://github.com/freedomofpress/securedrop-workstation/pull/898/commits/c2ce8d42a7ce480fc45f87fc54a5719b44d26011 | NOT INCLUDED | already in main |  
https://github.com/freedomofpress/securedrop-workstation/pull/898/commits/a708a60163d40e079a1f6f22571ded97a4405d8d | NOT INCLUDED | merge commit, already in main |  
https://github.com/freedomofpress/securedrop-workstation/pull/898/commits/99b72a5eb7d1dab90db1b175587c5e2041e052bc | INCLUDED | RPC migration -  needs further review tho, to make sure legacy policies aren't overriding the rules here |  
https://github.com/freedomofpress/securedrop-workstation/pull/898/commits/b127b667edef7f5b2d1d55b0149d2e510699edd0 | NOT INCLUDED | launcher no longer being split out |  
[...]

PR #974, opened 2024-04-10
reabsorb `securedrop-updater` without migrations

### LINKED ISSUE #966: Port updates from defunct securedrop-updater repo to securedrop-workstation
Code cleanup and general fixes for the current launcher were added in https://github.com/freedomofpress/securedrop-updater - these should be evaluated and pulled back into https://github.com/freedomofpress/securedrop-workstation

- [x] review commits from the last common point: https://docs.google.com/spreadsheets/d/1he6yeZ2frwhMLw7N4bfZQgx61E7-vz3la9axriY-oAk/edit
    - 400f6828bdef49f0ae2f8b937712afe4caf0808c
    - freedomofpress/securedrop-updater@18cacc89fb7c23beeea8f6c36234393d68730ebd
- [x] document what should be kept and why (see #965 for precedent)
- [x] copy desired changes into securedrop-workstation (most likely using the `4_2-compat-keep-launcher` branch as a base)
    - [ ] #966

cfm:
## Status

Ready for review

## Description of Changes

Closes #966 (superseding #971) by cherry-picking freedomofpress/securedrop-updater@ba99a93314837d0378d75cf966e5427c7355dfce...
- without CircleCI; and
- without versioned migration, as decided in <https://github.com/freedomofpress/securedrop-workstation/issues/966#issuecomment-2035435574>.

That is, most of these commits just cherry-pick commits from <https://github.com/freedomofpress/securedrop-updater/compare/c72c7e17ee33a85b9ac5bf78356974471a523357%5E...ba99a93314837d0378d75cf966e5427c7355dfce>, and most of the work here was resolving conflicts following from freedomofpress/securedrop-updater@c72c7e17ee33a85b9ac5bf78356974471a523357.

## Testing

- [ ] Visual review.  Nothing in the diff should be either terribly interesting or terribly surprising.  If it is, flag it for discussion!
- [ ] Locally: `cd launcher && source .venv/bin/activate && make test`
- [ ] CI passes.

On Qubes 4.2 either (a) freshly installed or (b) after `sdw-admin --uninstall`:
- [ ] `make clone` successfully builds the dom0 RPM package.
- [ ] `make dev` successfully provisions the system.

After reboot...
- [ ] `systemctl --user list-timers` shows `sdw-notify.timer` with ≤ 2 hours remaining.
- [ ] The `SecureDrop` icon appears...
    - [ ] ...and has the correct image image (after reboot).
    - [ ] ...and clicking it does not trigger a permissions warning.
- [ ] Via the `SecureDrop` icon, the updater (`sdw-updater`) runs successfully.
- [ ] After the updater completes, it (via `sdw-login`) starts `sd-app` and runs the Client.
- [ ] dom0: `make test` passes.

After `sdw-admin uninstall` and a reboot:
- [ ] `systemctl --user list-timers` does *not* show `sdw-notify.timer`.

## Deployment

As part of work towards #965, this assumes a fresh installation (except during testing) and so has no special considerations.


cfm:
@zenmonkeykstop and I discussed this work in progress and would like to try two different approaches instead of this one:

1. **New plan A:** Squash-merge just the [commits of interest](https://docs.google.com/spreadsheets/d/1he6yeZ2frwhMLw7N4bfZQgx61E7-vz3la9axriY-oAk/edit#gid=0), with the migration-related commits reverted (rather than rebased out) to preserve them from `freedomofpress/securedrop-updater@main`.  This should fail quickly if it's not going to work, in which case:
2. **New plan B:** Manually reconstruct (via `git cherry-pick -x`?) the commits of interest, excluding the migration-related commits.

cfm:
Plan B is coming along nicely via `git cherry-pick -x -Xfind-renames`.  I'm about a third of the way through the [commits of interest](https://docs.google.com/spreadsheets/d/1he6yeZ2frwhMLw7N4bfZQgx61E7-vz3la9axriY-oAk/edit#gid=0).

cfm:
@ro, we should sync up about these changes v. your #994, since some of `securedrop-updater`'s reorganization (even once reabsorbed here) anticipates your refactoring.

@deeplow: Based on <https://github.com/freedomofpress/securedrop-workstation/compare/reabsorb-updater-no-migrations...deeplow:securedrop-workstation:899-4.2-updater>, I believe following renames should be all you need to do once this reaches `main`.

rocodes:
@cfm thanks for flagging - I'd say that you should go ahead cause this is priority, and consider my changes as something that will work around yours - I'll resolve conflicts on my side once your changes land.

rocodes:
@cfm thanks for flagging - My assumption was that your changes here take priority and I will accommodate my work around them once this lands. But happy to touch base tomorrow :)

cfm:
Rebased from `main` after #990 and #997, from which together the diff is <https://github.com/freedomofpress/securedrop-workstation/compare/d1ae5a3d99b80076ecdd44a3f7bdc87248141e99..21b6288>.

rocodes:
@cfm So far so good - just a quick note to mention that a nice addition would be an additional sentence or two in the launcher README explaining the symlink and any relevant instructions for our future selves

cfm:
Let me know what you think of 5d439649597a0ce56e94b5913a637690a0980a57.

rocodes (commented):
## Testing

- [x] Visual review.  Nothing in the diff should be either terribly interesting or terribly surprising.  If it is, flag it for discussion!
- [x] Locally: `cd launcher && source .venv/bin/activate && make test`
- [x] CI passes.

On Qubes 4.2 either (a) freshly installed or (b) after `sdw-admin --uninstall`:
- [x] `make clone` successfully builds the dom0 RPM package.
- [x] `make dev` successfully provisions the system.
    -  :x:   `systemctl --user list-timers` shows `sdw-notify.timer` **No**, not even after reboot, but `/usr/lib/systemd/user/sdw-notify.service` and `sdw-notify.timer` are  present (do we need a `systemctl --user enable` for the timer somewhere?)

- [x] The `SecureDrop` icon appears...
    - [x] ...and has the correct image image (after reboot).
    - [x] ...and clicking it does not trigger a permissions warning.
- [x] Via the `SecureDrop` icon, the updater (`sdw-updater`) runs successfully.
- [x] After the updater completes, it (via `sdw-login`) starts `sd-app` and runs the Client.
- [x] dom0: `make test` passes.

rocodes:
I have to step afk, but I'll look more into the systemd timer thing when I return, or if you have any thoughts I can retest when I am back at keyboard later this eve.

rocodes:
Ok, so based on https://github.com/freedomofpress/securedrop-updater/pull/29#issuecomment-1462684087 it looks like enabling the systemd timer was out of scope for the original work by design. I'm fine with that (it's similar to how the dom0 config rpm puts salt files in place in `/srv/salt`, but doesn't enable them).  But it does mean that we have to decide where to add that enabling (and starting) logic. For now probably a salt state.

rocodes (changes requested):
This is looking very good, but since it removes `sd-dom0-crontab`, let's include a Salt state to enable the systemd service or timer so that we don't lose functionality. (If however you want to make the argument for putting that in a separate followup PR for the sake of getting this merged, that would also be ok with me).

We can also remove the crontab line from our uninstall logic (https://github.com/freedomofpress/securedrop-workstation/pull/974/files#diff-26037dbf9e3eacc9db8b2932bf7b532cd61e78b529f05b552622840c0bf4c155L93) unless you prefer to keep it in for developer convenience and remove it before 4.2-compat release, but maybe with a comment so we remember to clean it up?

I have to step afk but I will come back later tonight to try launcher tests again. For some reason the launcher test suite isn't happy with my OS-installed pyqt5 version, I'll debug that when I have a little more time. otherwise lgtm.

cfm:
On Tue, Apr 23, 2024 at 03:30:02PM -0700, rocodes wrote:
> This is looking very good, but since it removes `sd-dom0-crontab`,
> let's include a Salt state to enable the systemd timer so that we
> don't lose functionality. (If however you want to make the argument
> for putting that in a separate followup PR for the sake of getting
> this merged, that would also be ok with me).
> 
> We can also remove the crontab line from our uninstall logic
> (https://github.com/freedomofpress/securedrop-workstation/pull/974/files#diff-26037dbf9e3eacc9db8b2932bf7b532cd61e78b529f05b552622840c0bf4c155L93)
> unless you prefer to keep it in for developer convenience and remove
> it before 4.2-compat release, but maybe with a comment so we remember
> to clean it up?

Great catches, both done in 2d98bd2a2e03bcfcc023f857cd08b0134445231e
(pending CI).

dom0/sd-clean-all.sls
```diff
@@ -112,3 +112,12 @@ sd-cleanup-sys-firewall:
       - qvm-run sys-firewall 'sudo rm -f /rw/config/RPM-GPG-KEY-securedrop-workstation-test'
       - qvm-run sys-firewall 'sudo rm -f /etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation'
       - qvm-run sys-firewall 'sudo rm -f /etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation-test'
+
+disable-systemd-units:
+  cmd.run:
+    - name: systemctl --user disable sdw-notify.timer
```
rocodes, line 118: Noting for searchability/for our future selves: I think this is fine, but if we run into any errors we can try ` systemctl [...] || true` in `cmd.run` situations where we want to be sure we continue regardless, eg in cleanup. [saltstack-formula](https://github.com/saltstack-formulas/mongodb-formula/blob/fbe383183ff7742292e24f24ff734effe1a87384/mongodb/service/clean.sls#L24-L26) example

dom0/sd-dom0-systemd.sls
```diff
@@ -21,3 +25,14 @@ apply-systemd-changes:
   cmd.run:
     - name: sudo systemctl restart systemd-logind
 {% endif %}
+
+enable-user-units:
+  cmd.run:
+    - names:
+        - systemctl --user daemon-reload
```
rocodes, line 32: I wonder if cmd.run will reliably work here with these two commands back to back, or if we need to have the enable command wait for the service to reload. (ie `module.wait` for systemd to restart and then have the timer enabling `watch` the previous state). Will look more into this briefly tm.

cfm: @rocodes, are you questioning whether `systemctl daemon-reload` is blocking?  [It is.](https://github.com/systemd/systemd/blob/c2e6ed612a11d363dfca71c9a78ea005bfc0cc3a/src/systemctl/systemctl-daemon-reload.c#L40-L41)  But [it looks like `cmd.run(names=[])` is not executed in deterministic order](https://github.com/saltstack/salt/issues/16879#issuecomment-60663390), even though `names` is a list.  This isn't the only state declaration in which we implicitly deterministic ordering of `names`.  Let me tack on a commit to fix that.

rocodes: Sorry, my question was about whether cmd.run would reliably chain commands like that or not, and it looks like you found the answer. Looking at the changes now :)

rocodes: Nice one. It's a bit hard to find documentation from Salt about this syntax but there are a few [examples](https://github.com/saltstack/salt-ci-images/blob/327950581ba196f2cd10e992deab7b9c43133f80/golden-state-tree/pkgs/vault.sls#L14-L18), so lgtm

rocodes (approved):
## (re) Testing

- [x] Visual review.  Nothing in the diff should be either terribly interesting or terribly surprising.  If it is, flag it for discussion!
- [x] Locally: `cd launcher && source .venv/bin/activate && make test`
- [x] CI passes.

On Qubes 4.2 either (a) freshly installed or (b) after `sdw-admin --uninstall`:
- [x] `make clone` successfully builds the dom0 RPM package.
- [x] `make dev` successfully provisions the system.

After reboot...
- [x] `systemctl --user list-timers` shows `sdw-notify.timer` with ≤ 2 hours remaining.
- [x] The `SecureDrop` icon appears...
    - [x] ...and has the correct image image (after reboot).
    - [x] ...and clicking it does not trigger a permissions warning.
- [x] Via the `SecureDrop` icon, the updater (`sdw-updater`) runs successfully.
- [x] After the updater completes, it (via `sdw-login`) starts `sd-app` and runs the Client.
- [x] dom0: `make test` passes

After `sdw-admin uninstall` and a reboot:
- [x] `systemctl --user list-timers` does *not* show `sdw-notify.timer`.

Thanks @cfm for your careful work on this. :green_circle:

PR #976, opened 2024-04-10
[4.2] Cleanup - fix readme, remove old template consolidation check.

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

- Updates readme to reference 4.2
- Removes template cleanup logic used in original large/small template consolidation migration.

## Testing
- [ ] CI is passing
- [ ] Visual review.


dom0/securedrop-check-migration
```diff
@@ -19,11 +19,3 @@ function request_migration() {
     mkdir -p "$flag_dir"
```
legoktm, line 19: The `request_migration` function is no longer used; I think this entire file can be deleted?

zenmonkeykstop: Probably more than that - securedrop-check-migration is used by sd-upgrade-templates, which in turn is running the securedrop-handle-upgrade script, which is also run by sd-remove-unused-templates. There's a whole chunk that can be dropped

rocodes: LGTM but the file also needs to be removed from dom0 spec: https://github.com/freedomofpress/securedrop-workstation/blob/759cbc5686fce192421a00408d00d835439bad65/rpm-build/SPECS/securedrop-workstation-dom0-config.spec#L77

rocodes (changes requested):
lgtm pending fixup to dom0 .spec file (see comment)

rocodes (approved):

PR #979, opened 2024-04-10
Initial Qubes 4.2 Updater Integration

### LINKED ISSUE #899: Investigate using Qubes updater in 4.2

deeplow:
## Status
[edit] Ready for Review Opened PR due to the closure of https://github.com/freedomofpress/securedrop-workstation/pull/967.

TODO:
  - [ ] Consider if it makes sense to update even after a migration was applied (this was not the case before)

Blockers:
  - https://github.com/QubesOS/qubes-issues/issues/8493

## Description of Changes

Fixes #899 . Keeps most of the existing updater logic.

- updates templates via the CLI Qubes Updater instead of via salt
- removes functions deprecated by the new Qubes updater

## Testing
1. `make dev`
2. launch updater
3. Observe templates starting as they are updated
4. Observe app qubes getting restarted in background
5. See "successfully completed"
6. Inspect SDW-launcher logs to ensure correct updating of templates.

## Deployment

Any special considerations for deployment? Will only apply to new installs

> **Note**: tests don't pass due to the WIP nature of the base 4.2 branch. However, it seems that launcher-specific tests weren't passing already and it was unclear how to run the tests. 

### DIFF
--- a/launcher/tests/test_updater.py
+++ b/launcher/tests/test_updater.py
@@ -8,7 +8,7 @@ from unittest.mock import call
 import pytest
 
 from sdw_updater import Updater
-from sdw_updater.Updater import UpdateStatus, current_templates, current_vms
+from sdw_updater.Updater import UpdateStatus
 
 debian_based_vms = [
     "sd-app",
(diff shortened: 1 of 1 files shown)


legoktm:
> ## Open Implementation Questions
> ### What to do with the progress bar?
> 
> Currently this PR functionally removed a lot of meaningful progress reporting. I see several alternatives:
> 
> 1. Keep it as is. Even with reduced reporting (10%, 15%, 75%, 100%)
> 2. Remove progress bar entirely.
> 3. Remove progress bar and launch GUI updater for visual progress indication
> 4. Parse CLI updater for progress reporting. With the `--show-output` we should be able to know when template has finished.

I think the status quo is fine for now, but in the long term we probably want 3 or 4. If it's not too much work to implement in this PR, then I think it's probably worth doing. Just nicer to keep PRs on the smaller side and do it incrementally when possible.

> ### Should we use the GUI Updater?
> 
> Launching the Qubes GUI updater could be helpful in two aspects

My understanding from https://github.com/freedomofpress/securedrop-workstation/issues/899#issuecomment-2015739466 is that this is the long-term plan right? Are you just asking what should be done for this PR?

rocodes:
Note from conversation: Early testers will have to patch `/usr/lib/python3.11/site-packages/vmupdate/qube_connection.py` per  https://github.com/QubesOS/qubes-core-admin-linux/commit/61b4dc73c50941056d7ca7c615a0af79d035e7ae

rocodes:
[have requested a branch in the main repo so we can see full CI suite]

Issue #939, opened 2024-02-02
Pre-flight check: smoke test on VM / RPC configuration before starting


deeplow:
As I understand it, the updater runs before every startup. However, I learned from @zenmonkeykstop that only in certain occasions does it run the full salt deployment (I think this [is happening here](https://github.com/freedomofpress/securedrop-workstation/blob/356b289390005d59eaa0efd2cd7647c4f75e3b9d/launcher/sdw_updater_gui/UpdaterApp.py#L250)).

This makes sense as otherwise this would take too much time. But I wonder if it would be worth adding some sort of smoke test before startup just to make sure all the expected qubes are there and the proper policies in place to protect against the threat of user user mistake.

As it stands Qubes hardly has any foot-shooting protection. So it could be good to explore this extra guard rail. Furthermore, the results of this smoke test could be logged so in case of troubleshooting we could understand if there was some architecture deviation.

PR #984, opened 2024-04-12
Skip tests that check packages are up to date in CI

legoktm:
## Status

~~Ready for review~~ ~~in progress~~ Ready for review

## Description of Changes

In CI, the base templates are only updated once a day so it can easily be out of date.

## Testing

* [ ] Visual review
* [ ] test log shows tests are skipped

## Deployment

### DIFF
--- a/tests/test_vms_platform.py
+++ b/tests/test_vms_platform.py
@@ -1,4 +1,5 @@
 import json
+import os
 import unittest
 import subprocess
 
@@ -17,6 +18,8 @@ apt_url = ""
 FPF_APT_TEST_SOURCES = "deb [arch=amd64] https://apt-test.freedom.press {dist} {component}"
 FPF_APT_SOURCES = "deb [arch=amd64] https://apt.freedom.press {dist} {component}"
 APT_SOURCES_FILE = "/etc/apt/sources.list.d/securedrop_workstation.list"
+
+IS_CI = os.environ.get("CI") == "true"
 
 
 class SD_VM_Platform_Tests(unittest.TestCase):
@@ -165,6 +168,7 @@ sub   rsa4096 2021-05-10 [E] [expires: 2024-07-08]"""
         fpf_gpg_pub_key_fp = "2359E6538C0613E652955E6C188EDD3B7B22E6A3"
         self.assertFalse(fpf_gpg_pub_key_fp in results)
 
+    @unittest.skipIf(IS_CI, "Skipping on CI")
     def test_all_sd_vms_uptodate(self):
         """
         Asserts that all VMs have all available apt packages at the latest
@@ -174,6 +178,7 @@ sub   rsa4096 2021-05-10 [E] [expires: 2024-07-08]"""
             vm = self.app.domains[vm_name]
             self._ensure_packages_up_to_date(vm)
 
+    @unittest.skipIf(IS_CI, "Skipping on CI")
     def test_all_fedora_vms_uptodate(self):
         """
         Asserts that all Fedora-based templates, such as sys-net, have all


zenmonkeykstop:
Could we mark it xfail or to not be run in CI instead? The expectation when doing a dev install is that the system has been updated recently. I don't think we have to worry much about our packages at all, since they won't update that frequently.

IMO the optimal thing to do here would be to update the system templates automatically as part of the install process - which we can probably do as a shell callout with the Qubes updater now.

legoktm:
> Could we mark it xfail or to not be run in CI instead?

Yes, both are possible. I don't think xfail is useful, and would rather go the "check if CI" route or remove it altogether.

> The expectation when doing a dev install is that the system has been updated recently.

I agree, however, I don't think it is useful to test for this property. As a developer, every time I do a `make test` run, I should also not need to run the updater. Or to put it another way, I don't think we should be verifying that the system under test is up to date, rather we should be verifying that the updater works and is able to keep the system up to date.

rocodes:
IMO we do want overall system checks/tests, but probably run periodically against a prod/prodlike "real world conditions" system, not during CI or on dev machines (unless part of a release preflight CI somehow).  I think these tests are a relic from when `make dom0 test` was a `make sanity` check before publishing an rpm, as opposed to something meant to be scheduled and run periodically. 

A system test that runs the updater and then makes assertions about system state, separate from regular CI, is a good idea. (fwiw: kicksecure/whonix provides a `systemcheck` package: https://www.kicksecure.com/wiki/Systemcheck that gets at a similar goal, although primarily it's checking consistent system attributes plus their own metapackage versions).

tldr I'm fine with removing these tests from this part of CI since it's not a CI issue per se - we're more worried about the state of end user systems than anything else with these tests- but I would like to see what these tests represent being tested somewhere, eg as related to #939.

legoktm:
> I think these tests are a relic from when `make dom0 test` was a `make sanity` check before publishing an rpm, as opposed to something meant to be scheduled and run periodically.

Gotcha, that makes sense.

> A system test that runs the updater and then makes assertions about system state, separate from regular CI, is a good idea.

👍🏾 

> tldr I'm fine with removing these tests from this part of CI since it's not a CI issue per se - we're more worried about the state of end user systems than anything else with these tests- but I would like to see what these tests represent being tested somewhere, eg as related to https://github.com/freedomofpress/securedrop-workstation/issues/939.

OK, once https://github.com/freedomofpress/securedrop-workstation-ci/pull/55 lands I'll update this to skip this test if run in CI. And when the updater work happens, we can work on better tests.

rocodes:
[moving temporarily to `blocked` to reflect dependency on https://github.com/freedomofpress/securedrop-workstation-ci/pull/55]

legoktm:
This is ready for review again :)

legoktm:
From the log:
```
...
INFO:[2024-04-23-17:49:13:536805] test_all_fedora_vms_uptodate (test_vms_platform.SD_VM_Platform_Tests.test_all_fedora_vms_uptodate)
INFO:[2024-04-23-17:49:13:536835] Asserts that all Fedora-based templates, such as sys-net, have all ... skipped 'Skipping on CI'
INFO:[2024-04-23-17:49:13:537020] test_all_sd_vm_apt_sources (test_vms_platform.SD_VM_Platform_Tests.test_all_sd_vm_apt_sources)
INFO:[2024-04-23-17:49:13:537060] Test all VMs fpf apt source list iteratively. ... ok
INFO:[2024-04-23-17:49:13:537092] test_all_sd_vm_platforms (test_vms_platform.SD_VM_Platform_Tests.test_all_sd_vm_platforms)
INFO:[2024-04-23-17:49:13:537126] Test all VM platforms iteratively. ... ok
INFO:[2024-04-23-17:49:13:537160] test_all_sd_vms_uptodate (test_vms_platform.SD_VM_Platform_Tests.test_all_sd_vms_uptodate)
INFO:[2024-04-23-17:49:13:537191] Asserts that all VMs have all available apt packages at the latest ... skipped 'Skipping on CI'
...
```

zenmonkeykstop (approved):
LGTM based on visual review, CI results

PR #986, opened 2024-04-15
Install grub2-xen-pvh through RPM dependencies

legoktm:
## Status

Not ready for review

## Description of Changes

In a normal production install, we install the
securedrop-workstation-dom0-config RPM through `qubes-dom0-update`, which will fetch any dependencies since it has network access.

However in a dev setup, we install the RPM we just built using plain `dnf`, which has no network access, and is therefore unable to download missing dependencies. We can workaround this by passing the dependency list to `qubes-dom0-update` prior to RPM installation. I was unable to figure out how to pass a local RPM file qubes-dom0-update in the same way you can to dnf.

## Testing

* [ ] CI passes

## Deployment

Any special considerations for deployment? No, prod install behavior should already work

## Checklist

### DIFF
--- a/files/sdw-admin.py
+++ b/files/sdw-admin.py
@@ -63,17 +63,6 @@ def parse_args():
     return args
 
 
-def install_pvh_support():
-    """
-    Installs grub2-xen-pvh in dom0 - required for PVH with AppVM local kernels
-    TODO: install this via package requirements instead if possible
-    """
-    try:
-        subprocess.run(["sudo", "qubes-dom0-update", "-y", "-q", "grub2-xen-pvh"])
-    except subprocess.CalledProcessError:
-        raise SDWAdminException("Error installing grub2-xen-pvah: local PVH not available.")
-
-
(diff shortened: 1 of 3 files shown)


zenmonkeykstop:
Do we use qubes-dom0-update? By current docs the initial install is also via `dnf`.

legoktm:
> Do we use `qubes-dom0-update` for prod installs? By current docs the initial install is also via `dnf`.

Nope, I just figured that out myself (and marked as draft accordingly). I am not sure why we do it that way, I'll spend a bit of time poking but probably will postpone this until we spend time on the install process.

zenmonkeykstop:
If we had a key+repo-config-only rpm (ideall,y that was available via a regular qubes dom0 repo) then you could probably install *that* via `dnf` and then you'd be home free on the prod side

legoktm:
This needs to be properly revisited when we discuss/rework the install workflow. Closing and making sure an issue is open to track our desire.

PR #956, opened 2024-02-29
configure VMs via `vm-config.*` features

cfm:
## Status

Work in progress

## Description of Changes

Towards #936
- Based/blocked on #970

Changes proposed in this pull request:

## Testing

If you made non-trivial code changes, include a test plan and validated it for this PR.

## Deployment

Any special considerations for deployment? Consider both:

1. Upgrading existing pilot instances
2. New installs

### If you have made changes to the provisioning logic

- [ ] All tests (`make test`) pass in `dom0`

### If you have added or removed files

- [ ] I have updated `MANIFEST.in` and `rpm-build/SPECS/securedrop-workstation-dom0-config.spec`

### If documentation is required

- [ ] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-workstation-docs) for these changes, or will do so later
- [ ] I would appreciate help with the documentation

Issue #936, opened 2024-01-16
[4.2] Write vm-specific config values to qubesdb


rocodes:
Tracking/discussion. Let's make use of `.vm-config` part of qubesdb (internal to a VM) to store vm-specific configuration. Eventually this will allow us to get rid of config.json.

Upstream documentation:
https://dev.qubes-os.org/projects/core-admin-client/en/latest/manpages/qvm-features.html#vm-config

## Configurable components

| Status | Value | `dom0`[1] | `sd-app` | `sd-gpg` | `sd-log` | `sd-proxy` | `sd-whonix` |
| --- | --- | --- | --- | --- | --- | --- | --- |
| Drafted | `$QUBES_GPG_DOMAIN` | | [`$QUBES_GPG_DOMAIN`](https://github.com/freedomofpress/securedrop-client/blob/520e6a7490ebd0f1580b52e40127cc913c55e646/client/files/sd-app-qubes-gpg-domain.sh)[2] |
| n/a | `config.json:.environment` | →`/etc/apt/sources.list.d/securedrop_workstation.list` |
| Drafted | `config.json:.hidserv.hostname` | | | | | `/home/user/.securedrop_proxy/sd-proxy.yaml`[2] | `/var/lib/tor/keys/app_journalist.auth_private`[3] |
| Drafted | `config.json:.hidserv.key` | | | | | | `/var/lib/tor/keys/app_journalist.auth_private`[3] |
| Drafted | `config.json:.submission_key_fpr` | | `/home/user/.securedrop_client/config.json`[2] |
| n/a | `config.json:.vmsizes.sd_app` | `qvm-volume resize sd-app:private ...` |
| n/a | `config.json:.vmsizes.sd_log` | `qvm-volume resize sd-log:private ...` |
| | `salt://sd/sd-journalist.sec` | | | `/home/user/.gnupg/sd-journalist.sec`[4] |

1. Out of scope of this ticket; just documenting here for completeness.
2. [x] Program reads directly
3. [x] Helper program templates into file contents
4. [x] **Secret:** install directly into domain and remove from `dom0`

**NB.**  Only testable on Qubes 4.2, where QubesDB starts to return `vm-config.*` as `/vm-config/*`.

## Related
- https://github.com/freedomofpress/securedrop-workstation/issues/853

cfm:
[...]

PR #1001, opened 2024-04-24
deploy: set `vm-config` features for `sd-{app,proxy,whonix}`

cfm:
## Status

Ready for review

## Description of Changes

Towards #936 (superseding #956 due to its `skip-ci/` prefix):
1. sets values from the dom0 `config.json` to keys under the `vm-config/` prefix in QubesDB (under the `vm-config.` prefix in Salt); and
2. tests *all* VMs for expected configuration keys, eliminating the special-casing of `QUBES_GPG_DOMAIN`;
3. tests configured VMs for their expected values from `config.json`; and
4. leaves `TODO` hints in state files that can be removed as each VM's applications are updated to read configuration from this store, as in freedomofpress/securedrop-client#1883, and their Salt-managed configuration files can be removed.

## Testing

With this branch checked out in your `$SECUREDROP_DEV_VM`:

```sh-session
[user@dom0 securedrop-workstation]$ make clone
[user@dom0 securedrop-workstation]$ make dev
```

Now you can poke around (e.g.):

```sh-session
[user@dom0 securedrop-workstation]$ qvm-run --pass-io sd-app qubesdb-read /vm-config/QUBES_GPG_DOMAIN
sd-gpg
[user@dom0 securedrop-workstation]$ qvm-run --pass-io sd-app qubesdb-read /vm-config/SD_SUBMISSION_KEY_FPR
5F4BB12BAF811C789347DDFF148B87347CB3DDA1
```

That's it!  Since nothing downstream consumes from this configuration store yet, it has no side effects.

### Extra credit

Test the enforcement of expected configuration keys:

```sh-session
[user@dom0 securedrop-workstation]$ qubesdb-write -d sd-proxy /vm-config/QUBES_GPG_DOMAIN "you shouldn't have this"
[user@dom0 securedrop-workstation]$ make test-proxy  # Expected: "AssertionError: Items in the first set but not the second..."
```

## Deployment

As part of work towards https://github.com/freedomofpress/securedrop-workstation/issues/965, this assumes a fresh installation (except during testing) and so has no special considerations.

### DIFF
--- a/dom0/sd-app.sls
+++ b/dom0/sd-app.sls
@@ -33,6 +33,15 @@ sd-app:
 
 {% import_json "sd/config.json" as d %}
 
+sd-app-config:
+  qvm.features:
+    - name: sd-app
+    - set:
+        # TODO: freedomofpress/securedrop-client:client/files/sd-app-qubes-gpg-domain.sh
+        - vm-config.QUBES_GPG_DOMAIN: sd-gpg
+        # TODO: sd-app:/home/user/.securedrop_client/config.json
+        - vm-config.SD_SUBMISSION_KEY_FPR: {{ d.submission_key_fpr }}
+
 # The private volume size should be defined in the config.json
 sd-app-private-volume-size:
   cmd.run:
--- a/dom0/sd-proxy.sls
+++ b/dom0/sd-proxy.sls
@@ -29,3 +29,12 @@ sd-proxy:
     - require:
       - qvm: sd-whonix
       - qvm: sd-small-{{ sdvars.distribution }}-template
+
+{% import_json "sd/config.json" as d %}
+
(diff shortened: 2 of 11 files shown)


legoktm (approved):
Awesome, this looks great. Test plan checked out (qubesdb-read works properly, setting another key causes tests to fail).

Explicitly noting that the PGP secret key is not included in this PR, per https://github.com/freedomofpress/securedrop-workstation/issues/936#issuecomment-1930590304 and follow-up discussion.

One other thing I was curious about is whether we need to explicitly delete these or if they're removed as part of VM deletion. So I set an extra key (as part of the earlier testing), did a `sdw-admin --uninstall && make dev` and that key is (correctly) missing.

Issue #876, opened 2023-04-05
Simplify securedrop-keyring bootstrapping


legoktm:
We install the prod key into the main apt keyring with `apt-key` during the template creation: https://github.com/freedomofpress/qubes-template-securedrop-workstation/blob/main/securedrop-workstation/04_install_qubes_post.sh#L51

Then when we install the securedrop-keyring package we have a preinst that manually removes the key from the main apt keyring https://github.com/freedomofpress/securedrop-builder/blob/main/securedrop-keyring/debian/preinst while installing the securedrop-keyring.gpg file.

Then we have a suite of tests to make sure the prod key isn't in the main apt keyring: https://github.com/freedomofpress/securedrop-workstation/blob/main/tests/test_vms_platform.py#L140

It would be far simpler to just put the prod key where we want it in the first place, that is `/etc/apt/trusted.gpg.d/securedrop-keyring.gpg` and have the bootstrapping sources.list point to that file.

PR #1005, opened 2024-04-25
Provision deb822-style apt sources list

legoktm:
## Status

Work in progress

## Description of Changes

The new .sources file contains the content of the old sources.list files in a structure format and the PGP key, which makes it easy to ship in a templated way.

Salt will provision `/etc/apt/sources.list.d/apt_freedom_press.sources`, which will also be shipped and therefore overwritten by the securedrop-keyring package. As a consequence, the prod repo will always be enabled, even on staging/dev machines. This shouldn't have any practical implications, becuase staging/dev packages should always have higher verison numbers.

In staging/dev, salt will provision `apt-test_freedom_press.sources` with the respective components, using the apt-test key. If you then switch to a prod environment, it'll be removed.

This is a companion to
<https://github.com/freedomofpress/securedrop-client/pull/1952>, which updates the securedrop-keyring side.

* Fixes <https://github.com/freedomofpress/securedrop-workstation/issues/876>
* Refs <https://github.com/freedomofpress/securedrop/issues/6447>
* Refs <https://github.com/freedomofpress/securedrop-client/issues/1788>.

## Testing

* [ ] `make dev` plus `make test` works
* [ ] switching to `make staging` and then `make test` works
* [ ] edit `/usr/share/securedrop-workstation-dom0-config/config.json` to switch to prod, run `sdw-admin --apply` and it should remove the apt-test sources

## Deployment

Any special considerations for deployment? Only new installs are supported, as no cleanup is done of old sources.list files nor old keyring files.

## Checklist

- [ ] No docs changes need (I think)


legoktm:
The main issue I ran into is that the first provision will work, but the second will fail, because the bookworm repository does not exist yet. Here's a PR for that: https://github.com/freedomofpress/securedrop-apt-prod/pull/137.

legoktm:
I also want a test like https://github.com/freedomofpress/securedrop-client/blob/aec2057afe41953ad5a83df6409d4eace504896c/keyring/tests/test_sources.py except we don't have any setup for local tests in this repo (i.e. non-dom0) aside from the launcher tests.

PR #1007, opened 2024-04-26
Switch to poetry, split up `make install-deps`

### LINKED ISSUE #947: Migrate dev dependency management to poetry
We're still using pip-tools here, let's switch it to poetry. Also remember to enable dependabot when doing so.

```
user@dev ~/g/f/securedrop-workstation (main)> cat requirements/dev-requirements.in 
bandit
black
diffoscope
flake8
gitpython>=3.1.32 # for bandit, re CVE-2022-24439
isort
mypy
pip-tools
pip
pytest
reprotest
```

### LINKED ISSUE #992: Separate build and test dependencies in `make install-deps`
For cleanliness and slightly speedier builds, we should split `make install-deps`; currently we're installing Qt-related dependencies during RPM builds.

Follow-up to https://github.com/freedomofpress/securedrop-workstation/pull/955

legoktm:
## Status

Ready for review

## Description of Changes

Qubes 4.2's dom0 ships with Python 3.11, which is much easier to deal with than 4.1's 3.8, since it matches with Debian bookworm. As a result, we can expect most developers to be able to natively use 3.11 and not need a container wrapper.

It's not straightforward to just swap out pip for poetry here since poetry sets up a virtualenv per-user instead of globally installing.

At the same time, the `make install-deps` target has grown so that it encompases both build and test dependencies and is ripe for splitting. It also had some duplication in that build dependencies were specified in both the Makefile and specfile.

So we can address everything at once. There are now three effective sets of dependency groupings:

1) Python lint/test dependencies, specified in pyproject.toml and installed using poetry.
2) Build dependencies, specified in the `.spec` file and installed via `make build-deps`.
3) Test dependencies, specified in the Makefile and installed via `make test-deps`.

The corresponding make targets, container.sh and CI manifests have been updated to handle this. The `install-deps` target was removed so it isn't accidentally used anymore.

The launcher dependencies were identical except that virtualenv also needed PyQt5. We already had a root `make test-launcher`, so fix that up and get rid of separate launcher dependencies and Makefile.

Fixes #992.
Fixes #947.

## Testing

* [ ] visual review
* [ ] CI passes
* [ ] locally `make build-rpm` works
* [ ] same with `make rpmlint`
* [ ] and `make reprotest`
* [ ] and `make test-launcher`

## Deployment

Any special considerations for deployment? Should be all internal changes.


zenmonkeykstop:
`poetry install` works but emits:

```
Warning: The current project could not be installed: No file/folder found for package securedrop-workstation-dom0-config
If you do not want to install the current project use --no-root.
If you want to use Poetry only for dependency management but not for packaging, you can disable package mode by setting package-mode = false in your pyproject.toml file.
In a future version of Poetry this warning will become an error!
```

We probably do want to be setting package-mode=false.

legoktm:
> We probably do want to be setting package-mode=false.

Since it's just a warning I didn't want to trigger errors for people with old versions of poetry yet. I would prefer to keep ignoring the warning until they make it into a hard error and we have to migrate.

zenmonkeykstop (approved):
CI and test plan check out, LGTM

PR #1027, opened 2024-05-07
Functional RPC policy tests

### LINKED ISSUE #759: Add functional RPC policy tests
Our current [RPC policy tests](https://github.com/freedomofpress/securedrop-workstation/blob/main/tests/test_qubes_rpc.py) only test whether the policy files match the expected content, not whether they have the expected effect. We should test actual behavior, especially in light of the changes in Qubes 4.1/#751 which add significant complexity to the RPC policy behavior (two different formats, precedence rules between them).

micahflee:
## Status

Ready for review

## Description of Changes

Fixes #759.

This was simpler than I first expected it to be.

As discussed in #759, I skipped testing `qubes.USBAttach`, `qubes.USB`, `qubes.OpenInVM` rules, and also GPG-related allow rules, to allow the tests to run without user interaction.

This tests `securedrop.Log` (allow and deny), `securedrop.Proxy` (allow and deny), and the GPG-related rules (deny). It relies on there existing `sys-net` and `sys-firewall` VMs, as it uses these as example VMs where various SDW policies should be denied.

### DIFF
--- a/tests/test_qubes_rpc.py
+++ b/tests/test_qubes_rpc.py
@@ -1,9 +1,17 @@
 import os
 import unittest
+import subprocess
+
+RETURNCODE_SUCCESS = 0
+RETURNCODE_DENIED = 126
 
 
-# TODO: add actual integration testing of policies
 class SD_Qubes_Rpc_Tests(unittest.TestCase):
+    def _qrexec(self, source_vm, dest_vm, policy_name):
+        cmd = ["qvm-run", "--pass-io", source_vm, f"qrexec-client-vm {dest_vm} {policy_name}"]
+        p = subprocess.run(cmd, input="test", capture_output=True, text=True, check=False)
+        return p.returncode
+
     def test_policies_exist(self):
         """verify the policies are installed"""
         assert os.path.exists("/etc/qubes/policy.d/31-securedrop-workstation.policy")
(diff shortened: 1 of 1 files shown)


tests/test_qubes_rpc.py
```diff
@@ -1,14 +1,55 @@
...
+        return p.returncode
+
+    def test_policies_exist(self):
         """verify the policies are installed"""
         assert os.path.exists("/etc/qubes/policy.d/31-securedrop-workstation.policy")
         assert os.path.exists("/etc/qubes/policy.d/32-securedrop-workstation.policy")
 
+    # securedrop.Log from @tag:sd-workstation to sd-log should be allowed
+    def test_sdlog_from_sdw_to_sdlog_allowed(self):
+        self.assertEqual(self._qrexec("sd-app", "sd-log", "securedrop.Log"), RETURNCODE_SUCCESS)
```
zenmonkeykstop, line 22: Would it make sense to use qubesadmin here to get the sets of VMs tagged with `sd-workstation` (and not, for tests that check for blocking policies) and use those in the tests instead of hard-coding VMs?

tests/test_qubes_rpc.py
```diff
@@ -1,14 +1,55 @@
...
+
+    # securedrop.Proxy from anything else to sd-proxy should be denied
+    def test_sdproxy_from_other_to_sdproxy_denied(self):
+        self.assertEqual(self._qrexec("sys-net", "sd-proxy", "securedrop.Proxy"), RETURNCODE_DENIED)
+        self.assertEqual(
+            self._qrexec("sys-firewall", "sd-proxy", "securedrop.Proxy"), RETURNCODE_DENIED
+        )
+
+    # qubes.Gpg, qubes.GpgImportKey, and qubes.Gpg2 from anything else to sd-gpg should be denied
+    def test_qubesgpg_from_other_to_sdgpg_denied(self):
```
zenmonkeykstop, line 43: ditto on the sd-workstation tag suggestion.

zenmonkeykstop (changes requested):
Looks good, one suggestion to generalize the qrexec checks a little.

micahflee:
Now the `securedrop.Log` tests use `qubesadmin` to test each running VM with the `sd-workstation` to make sure it's allowed, and each running VM without it to make sure it's denied.

zenmonkeykstop (approved):

PR #1025, opened 2024-05-07
Stop using assert in validate_config.py

micahflee:
## Status

Ready for review

## Description of Changes

Fixes #683.

Here's how to reproduce the `TOB-SDW-024` bug:

In sd-dev switch to the `main` branch in `securedrop-workstation`. In dom0, run `make clone`.

One of the checks in `validate_config.py`  is that `self.config["vmsizes"]["sd_log"]` is an integer. So, in dom0, edit `/usr/share/securedrop-workstation-dom0-config/config.json` and make `sd_log` a string instead, like this:

```json
{
    "submission_key_fpr": "65A1B5FF195B56353CC63DFFCC40EF1228271441",
    "hidserv": {
        "hostname": "pmlfte7wii27fi5ioyjwzbgdmjk2i5nx2gd6v3rk5bym7rattnvm36ad.onion",
        "key": "NDN4MWTGFCCJMJWXISZDN2RO3DZJKMVRBGIFCNYN2RKX4VDGKJTQ"
    },
    "environment": "dev",
    "vmsizes": {
        "sd_app": 10,
        "sd_log": "5"
    }
}
```

In dom0, change to `~/securedrop-workstation` and run this to validate the config:

```sh
python3 files/sdw-admin.py --validate
```

It should fail:

```
[user@dom0 securedrop-workstation]$ python3 files/sdw-admin.py --validate
Validating...Traceback (most recent call last):
  File "/home/user/securedrop-workstation/files/sdw-admin.py", line 221, in <module>
    main()
  File "/home/user/securedrop-workstation/files/sdw-admin.py", line 172, in main
    validate_config(SCRIPTS_PATH)
  File "/home/user/securedrop-workstation/files/sdw-admin.py", line 108, in validate_config
    validator = SDWConfigValidator(path)  # noqa: F841
                ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/securedrop-workstation/files/validate_config.py", line 41, in __init__
    self.validate_existing_size()
  File "/home/user/securedrop-workstation/files/validate_config.py", line 132, in validate_existing_size
    assert isinstance(
AssertionError: Private volume size of sd-log must be an integer value.
```

Now run it with the `-OO` flag:

```sh
python3 -OO files/sdw-admin.py --validate
```

It will pass when really it should fail -- this is the bug:

```
[user@dom0 securedrop-workstation]$ python3 -OO files/sdw-admin.py --validate
Validating...OK
```

[...]

PR #1028, opened 2024-05-08
Configure ruff

### LINKED ISSUE #1022: Switch from flake8/bandit/black/isort to ruff
We've adopted ruff most everywhere else, would be good to get this repository ported over too. See https://github.com/freedomofpress/securedrop-client/pull/1885; probably worth copying that configuration to start with.

micahflee:
## Status

Ready for review

## Description of Changes

Fixes #1022.

This replaces flake8, bandit, black, and isort with ruff!

Many of the things that ruff flagged were `assert` statements in `validate_config.py`, and I already fixed those in #1025. So rather than fixing them again here, I merged that branch into this branch. So please review and merge #1025 first.

I started copying the ruff config from https://github.com/freedomofpress/securedrop-client/pull/1885, but then I paired it down to only include what's needed for this repo.

I added `S108` (probably insecure usage of /tmp) to the ignore list, as it was being triggered just on this code in `sdw_updater/Updater.py`:

```py
# We use a hardcoded temporary directory path in dom0. As dom0 is not
# a multi-user environment, we can safely assume that only the Updater is
# managing that filepath. Later on, we should consider porting the check-migration
# logic to leverage the Qubes Python API.
MIGRATION_DIR = "/tmp/sdw-migrations"  # nosec
```

I also added a per-file-ignore for `SIM115` (use context handler for opening files) for `sdw_util/Util.py` specifically for the `obtain_lock` and `can_obtain_lock` functions. This check wants you to open files using the `with` statement so they automatically close, but in this case these functions open files and then return the file handle. So I think making an exception is fine.

Though I did notice a potential small bug. In `can_obtain_lock`, the lock file is opened temporarily, just to see if it can be opened, but I don't think it's ever closed since it returns a bool instead of the file handle -- though maybe this is fine and I just don't understand how `fcntl.lockf` works:

```py
def can_obtain_lock(basename):
    """
    We temporarily obtain a shared, nonblocking lock to a lockfile to determine
    whether the associated process is currently running. Returns True if it is
    safe to continue execution (no lock conflict), False if not.

    `basename` is the basename of a lockfile situated in the LOCK_DIRECTORY.
    """
    lock_file = os.path.join(LOCK_DIRECTORY, basename)
    try:
        lh = open(lock_file)
    except FileNotFoundError:
        # Process may not have run during this session, safe to continue
        return True

    try:
        # Obtain a nonblocking, shared lock
        fcntl.lockf(lh, fcntl.LOCK_SH | fcntl.LOCK_NB)
    except OSError:
        sdlog.error(LOCK_ERROR.format(lock_file))
        return False

    return True
```


files/sdw-admin.py
```diff
@@ -70,7 +69,7 @@ def install_pvh_support():
     TODO: install this via package requirements instead if possible
     """
     try:
-        subprocess.run(["sudo", "qubes-dom0-update", "-y", "-q", "grub2-xen-pvh"])
+        subprocess.run(["sudo", "qubes-dom0-update", "-y", "-q", "grub2-xen-pvh"], check=False)
```
rocodes, line 72: wouldn't we want `check=True` here in order to trigger the exception below if the update is unsuccessful?

rocodes, line 72: We might want `check=True` or another subprocess method (`subprocess.check_call` ?) if the intention is to error if the grub2-zen-pvh installation doesn't work. (This isn't introduced by your PR, but ruff is making it noticeable.)

(Also, not really your problem since it was preexisting, but if you're in here anyway and open to random nits, I just noticed a typo in the error message on like 74: "Error installing grub2-xen-pvah" should be "Error installing grub2-xen-pvh", no "a".)

sdw_updater/Updater.py
```diff
@@ -361,12 +352,12 @@ def overall_update_status(results):
 
     if updates_failed:
         return UpdateStatus.UPDATES_FAILED
-    elif reboot_required:
+    if reboot_required:
```
rocodes, line 355: (this one's always gonna be weird for me because I feel like the `elif` is more readable and more like defensive coding even though it's not technically necessary. but okay, ruff.)

rocodes:
Thanks for preparing this, this looks great. I'm going to look into the lock file issue now. We use the lockfile only in the Updater/+Notifier, eg to make sure that we're not bugging the user to run updates while they're already running updates.

rocodes (changes requested):
Sorry to punt this back to you for just a [small fix](https://github.com/freedomofpress/securedrop-workstation/pull/1028/files#r1596837417) while our eyes are here - otherwise LGTM

files/sdw-admin.py
micahflee: Good catch. This was in my "Automatic ruff fixes" commit, so I think was automatically edited when I ran `ruff check --fix`. And sure enough, if you look at the docs for [subprocess.run()](https://docs.python.org/3/library/subprocess.html#subprocess.run), the default is `check=False`, which means it's always been false. I'll change to `subprocess.check_call()` like the other functions in this file use.

sdw_updater/Updater.py
rocodes: Thanks for pointing that out, it explains why this stuck out at me so much.

Sorry to nitpick, but for the sake of consistency, let's add back the rules that are specifically about readability, so that we have the same standards on all projects. 

So that would be, to my eye 

```
  # superflous-else- rules, find they hurt readability
    "RET505", "RET506", "RET507", "RET508",
  # Find contextlib.suppress() is harder to read
    "SIM105",
    # Find ternary statements harder to read
    "SIM108",
    # Using any()/all() can be harder to read
    "SIM110",   
```

rocodes (approved):

rocodes (approved):

Makefile
```diff
@@ -179,35 +179,19 @@ venv: ## Provision a Python 3 virtualenv for development (ensure to also install
...
-.PHONY: check-isort
-check-isort: ## Check Python import organization with isort
-	poetry run isort --check-only --diff .
-
-.PHONY: isort
-isort: ## Update Python import organization with isort
-	poetry run isort .
+.PHONY: check-ruff
+check-ruff: ## Check Python source code formatting with ruff
+	poetry run ruff check
```
legoktm, line 190: I think we also need to check `ruff format` - https://github.com/freedomofpress/securedrop-client/blob/2e2fdca7d280044e54578c5d4a6643a3207c77e5/Makefile#L34

files/destroy-vm.py
```diff
@@ -38,10 +39,11 @@ def destroy_vm(vm):
     Destroys a single VM instance. Requires arg to be
     QubesVM object.
     """
-    assert SDW_DEFAULT_TAG in vm.tags
+    if SDW_DEFAULT_TAG not in vm.tags:
+        raise Exception
```
legoktm, line 43: Can we have an error message here?

Makefile
```diff
@@ -179,35 +179,19 @@ venv: ## Provision a Python 3 virtualenv for development (ensure to also install
...
 
-.PHONY: flake8
-flake8: ## Validate PEP8 compliance for Python source files
-	poetry run flake8
+.PHONY: ruff
+ruff: ## Fix Python source code formatting with ruff
```
legoktm, line 193: Can we either call this `fix` or alias that to this? In the spirit of https://github.com/freedomofpress/securedrop-tooling/issues/11.

launcher/tests/test_updater.py
```diff
@@ -207,13 +207,14 @@ def test_write_last_updated_flags_to_disk(mocked_info, mocked_error, mocked_call
...
+        f"echo '{current_time}' > {flag_file_sd_app}",
     ]
     mocked_call.assert_called_once_with(subprocess_command)
     assert not mocked_error.called
     assert os.path.exists(flag_file_dom0)
-    contents = open(flag_file_dom0, "r").read()
-    assert contents == current_time
+    with open(flag_file_dom0) as f:
+        contents = f.read()
+        assert contents == current_time
```
legoktm, line 217: the assert can happen outside the with statement

sdw_updater/Updater.py
```diff
@@ -111,9 +108,10 @@ def apply_updates(vms=current_templates, progress_start=15, progress_end=75):
     Returns a tuple of (vm_name, percentage_progress, upgrade_results),
     for use in updating the GUI progress bar.
     """
-    sdlog.info("Applying all updates to VMs: {}".format(vms))
+    sdlog.info(f"Applying all updates to VMs: {vms}")
     # Figure out how much each completed VM should bump the progress bar.
-    assert progress_end > progress_start
+    if progress_end <= progress_start:
+        raise Exception
```
legoktm, line 114: same, can we have an error message here?

sdw_updater/Updater.py
legoktm: Can we re-run the automated fixes now that those rules are suppressed? i.e. back out these changes and elsewhere

legoktm (changes requested):
I rebased this now that the assert PR has landed and left some comments. One other thing is that the `# nosec` comments needed for bandit can now be removed.

micahflee:
The tests are failing because of this:

>   Unable to find installation candidates for pyqt5-qt5 (5.15.13)

I'm not sure what's wrong there, so any help is appreciated.

poetry.lock
```diff
@@ -421,15 +189,13 @@ PyQt5-sip = ">=12.11,<13"
 
 [[package]]
 name = "pyqt5-qt5"
-version = "5.15.2"
+version = "5.15.13"
```
legoktm, line 192: Something went wrong here, not sure why this package was upgraded. Let me back it out and see if that fixes the CI issue.

legoktm: Ah yeah, that'll do it; it tries to update everything unless you use `--no-update`

legoktm (changes requested):
Overall looks ready to land now, but could you rebase/squash this? I think some things landed in tests/test_vms_platform.py unintentionally.

legoktm (approved):
Rebased, just waiting on CI before merging

PR #1030, opened 2024-05-10
Package some dom0 files in config rpm rather than provisioning via Salt

rocodes:
## Status

Ready for review - but have not tested provisioning, so please review with extra eyes :)

## Description of Changes

Towards #274 

Changes proposed in this pull request:

Package files in rpm rather than provisioning them in Salt.  This doesn't get us all the way there, but for the "easy"/uncontroversial files it's a good start:
- RPC policy files
- Icon/enclosing directory
- Some cleanup of cleanup method (we no longer ship a qubes-templates repo file)

## Testing

- [x] Visual Review, with an eye to approving the changes in when these files are provisioned (rpm install time vs `sdw-admin --apply` time)
- [x] CI is successful
- [x] rpm building + provisioning is successful: successful SDCI test run or manual rpm + `make dev` required
- [x] Desktop icon + shortcut 

## Deployment

Any special considerations for deployment? Consider both:

1. Upgrading existing pilot instances
2. New installs

### If you have made changes to the provisioning logic

- [ ] All tests (`make test`) pass in `dom0`

### If you have added or removed files

- [ ] I have updated `MANIFEST.in` and `rpm-build/SPECS/securedrop-workstation-dom0-config.spec`

### If documentation is required

- [ ] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-workstation-docs) for these changes, or will do so later
- [ ] I would appreciate help with the documentation

Issue #274, opened 2019-06-26
Package vm-specific files 


emkll:
We use Salt to move certain configuration files to VMs when invoking Make commands in dom0. These files include files used for mime associations and default behavior for opening, but also Python scripts (in the case of sd-export):

* https://github.com/freedomofpress/securedrop-workstation/blob/master/sd-export/
* https://github.com/freedomofpress/securedrop-workstation/tree/master/sd-svs
* https://github.com/freedomofpress/securedrop-workstation/tree/master/sd-proxy

We should distribute these using .deb packages to ensure we are in a position to easily and automatically update them, without invoking any dom0 Make commands, nor requiring to update the securedrop-workstation source code.

eloquence:
Is this issue fully superseded by packaging work that's part of template consolidation (#471)?

conorsch:
To my eye, yes, it is. However, there are still a few scripts in dom0 that really should be bundled up in the RPM, to my eye, although perhaps that's best rephrased in a new issue.

zenmonkeykstop:
Ongoing and planned work on provisioning improvements will catch the stragglers here. Leaving open for now.

rocodes:
Posting some notes so we can decide next steps.
To recap, packaging files rather than updating them via salt will make it easier to reason about system status, perform migrations, depend less on Salt, etc.

<details><summary>Most files managed by Salt are in dom0 </summary>
    
    # fully managed (created, deleted) using Salt
    - /etc/yum.repos.d/securedrop-workstation-dom0.repo
    - /etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation
    - /etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation-test
    - /usr/share/securedrop/icons
    - /home/{{ gui_user }}/.config/autostart/SDWLogin.desktop
    - /usr/bin/securedrop-login
    - /etc/qubes-rpc/policy/securedrop.Log
    - /etc/qubes-rpc/policy/securedrop.Proxy
    - /home/{{ gui_user }}/Desktop/securedrop-launcher.desktop
    - /home/{{ gui_user }}/.securedrop_launcher
    - /srv/salt/qa-switch.tar.gz
    - /srv/salt/qa-switch
    - /srv/salt/consolidation-qa-switch.sh
    - /etc/qubes/policy.d/60-securedrop-workstation.policy
    - /etc/qubes/policy.d/70-securedrop-workstation.policy
     
[...]

PR #1031, opened 2024-05-13
Use conditional build in rpm to manage dev-only settings

rocodes:
## Status

Work in progress  
(needs testing, needs rpmlinting and a fixup in `build-rpm.sh` if we decide to pursue this route - this is a napkin sketch) 

## Description of Changes
Refs #274 

To further the goal of removing more logic from Salt and into packaging (easier versioning, faster dom0 provisioning therefore faster updater runs, less salt breakage) we are moving files from being provisioned during the apply run into being packaged by the dom0 config and deb vmconfig packages (see eg #1030).

This is straightforward when files are static, but is somewhat trickier when files are either dynamically generated based on runtime info, or when they are provisioned conditionally (ie depending on build flavour).

Filing this PR as a discussion starting point for handling the latter situation. It introduces a [build condition](https://rpm-software-management.github.io/rpm/manual/conditionalbuilds.html) `powersettings` , which is on by default and currently configures the logind.conf power lid settings, but if we like this approach, should be expanded to include the power management xfce logic.

The goal of this type of change is to provide a path for deprecatating all* the the `if d.environment == "staging" or d.environment == "prod"`  type dynamic checking in our salt runs. *(All, or all but the keyring/repo provisioning, which will be handled in a separate bootstrap package.)

* Replace inline configuration of logind.conf with a [drop-in file + directory](https://www.freedesktop.org/software/systemd/man/latest/systemd-system.conf.html) `/etc/systemd/login.conf.d/override_logind.conf`
* ...That is provisioned on prod and staging rpm builds (default option)
* ... And disabled on dev rpm builds (explicit build parameter passed to rpmbuild)
* And updates make targets and CI accordingly
Note: this change means that our power lid settings are applied when installing the RPM, as opposed to after successful orchestration. Per conversation with @zenmonkeykstop this seems like an acceptable decision given the amount of control we take over the SDW system, whether or not we go the conditional RPM build approach.

### More discussion
Another option could be to build fully separate rpm packages (workstation-dom0-config-staging, dom0-config-dev, etc). This could be a good idea particularly for rpm repo configuration (although in that case we might build different keyring/bootstrapping packages rather than different dom0 config packages. In any case, this solution is not mutually exclusive of that idea, but it's a little less complex.   

## Testing

- [ ] Visual Review
- [ ] CI Passes
- [ ] staging/ prod rpms are built correctly, and when installed, install a directory `/etc/systemd/login.conf.d/` with a configuration file that overrides power lid settings  
- [ ] Dev RPM is built correctly (in CI and locally) and does not include these changes

## Deployment

2. New installs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,11 @@ dev staging: assert-dom0 ## Configures and builds a dev or staging environment
 build-rpm: ## Build RPM package
 	USE_BUILD_CONTAINER=true $(CONTAINER) ./scripts/build-rpm.sh
 
+# Dev RPMs don't contain power settings adjustment (lid)
+.PHONY: build-dev-rpm
+build-dev-rpm: ## Build RPM package
+	DEV_BUILD_FLAG=true USE_BUILD_CONTAINER=true $(CONTAINER) ./scripts/build-rpm.sh
+
 # FIXME: the time variations have been temporarily removed from reprotest
 # Suspecting upstream issues in rpm land is causing issues with 1 file\'s modification time not being clamped correctly only in a reprotest environment
 .PHONY: reprotest
--- a/dom0/sd-clean-all.sls
+++ b/dom0/sd-clean-all.sls
@@ -99,12 +99,6 @@ sd-cleanup-etc-changes:
     - repl: ''
     - backup: no
 
-{% if d.environment == "prod" or d.environment == "staging" %}
-apply-systemd-changes:
-  cmd.run:
-    - name: sudo systemctl restart systemd-logind
-{% endif %}
-
 sd-cleanup-sys-firewall:
   cmd.run:
     - names:
--- a/files/logind_override.conf
+++ b/files/logind_override.conf
@@ -1,0 +1,2 @@
+[Login]
+HandleLidSwitch=poweroff
--- a/.github/workflows/nightlies.yml
+++ b/.github/workflows/nightlies.yml
@@ -30,7 +30,7 @@ jobs:
           git config --global --add safe.directory '*'
           # Version format is "${VERSION}-0.YYYYMMDDHHMMSS.fXX", which sorts lower than "${VERSION}-1"
           rpmdev-bumpspec --new="$(cat VERSION)-0.$(date +%Y%m%d%H%M%S)%{?dist}" rpm-build/SPECS/*.spec
-          make build-rpm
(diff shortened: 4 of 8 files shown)


rocodes:
Mostly filing this as a proof of concept and would love @legoktm 's opinion on this or other approaches

Makefile
```diff
@@ -32,6 +32,11 @@ dev staging: assert-dom0 ## Configures and builds a dev or staging environment
 build-rpm: ## Build RPM package
 	USE_BUILD_CONTAINER=true $(CONTAINER) ./scripts/build-rpm.sh
 
+# Dev RPMs don't contain power settings adjustment (lid)
```
zenmonkeykstop, line 35: IMO the build process should produce env-neutral packages, with env-specific stuff being set up at install- or run-time.

dom0/sd-clean-all.sls
```diff
@@ -99,12 +99,6 @@ sd-cleanup-etc-changes:
     - repl: ''
     - backup: no
 
-{% if d.environment == "prod" or d.environment == "staging" %}
```
zenmonkeykstop, line 102: Do we even need a conditional on this if we keep it? Is there a downside to the restart on dev environments?

files/logind_override.conf
```diff
@@ -0,0 +1,2 @@
+[Login]
```
zenmonkeykstop, line 1: This is definitely a better way to handle the logind config change than blockreplacing!

zenmonkeykstop (changes requested):
I'm a fan of the override as a way to reduce complexity here but I don't think dev-only RPM builds are warranted. As it stands the only difference between dev and prod RPMs are signatures and version strings - IMO this reduces the chance of untested changes getting shipped to prod.

At the same time I do support removing environment-specific config overall from the RPM - one other thing to consider here would  be splitting dom0/sd-default-config.{yml,sls} out into env-specific versions and only packaging the prod version. To do dev or staging builds devs need to clone the repo to dom0, so the non-prod versions would be available and could be subbed in after installing the RPM but before doing the sdw-admin --apply step.

rocodes:
Thanks everyone - so what I'll do is close this, submit a PR for the logind stuff, and we can prioritize systemd changes as the way to manage dev vs staging/prod power settings as @legoktm suggests.

Issue #1004, opened 2024-04-25
refactor service configuration from Salt (at provisioning) to systemd (at boot)


cfm:
* [x] I have searched for duplicates or related issues

## Description

@zenmonkeykstop asked this morning whether #1001 is sufficient for *all* VM-level configuration, not just keys and values.  I think we'll still want to use systemd units with [`ConditionHost`](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#ConditionHost=) conditions to enable individual services based on the hostname configured by Salt (and enforced by dom0 tests).

## How will this impact [SecureDrop/SecureDrop Workstation users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?

No user implications.

## How would this affect the SecureDrop Workstation [threat model](https://github.com/freedomofpress/securedrop-workstation/#threat-model)?

Along with #1001, this assumes we are comfortable with runtime (boot-time) configuration of VMs' roles and services, except for secrets.

cfm:
Switching to Qubes services makes sense, @deeplow.  Arguably it extends #1001's configuration injection to enabling services by analogy, which I like.


<https://github.com/freedomofpress/securedrop-workstation/issues/1004#issuecomment-2095970555>:

> Disadvantages:
> 
> * **salt being salt** - salt is good at setting state, but not removing it. If we ever needed to disable a qubes service, we'd need to explicitly remove it in the salt configuration

I guess I take this for granted for as long as we're using Salt in this way at all.  :-)

legoktm:
I think I lean in the direction of using ConditionHost, because I think it better fits our goal of keeping in-VM stuff in packages and using salt for dom0 things.

As a practical case is if we want to add a new "sd-log-whatever" service in a VM, we'd have to also do a corresponding workstation patch to enable the Qubes service through dom0, and gate the client release on the workstation one.z

legoktm:
Re: `immune to qube rename side-effects (if they ever happen)` - one consideration could be to set a `sd-app` Qubes service on the sd-app VM, `sd-log` service on sd-log VM, etc. and then all of the services in VM can use `ConditionPathExists=/var/run/qubes-service/<VM name>`, which does feel a little less brittle than the hostname, but also allows for a 1:1 mapping between systemd service and VM. I'm not sure that extra level indirection is worth it though.

legoktm:
> But other ideas may come up in the meeting we're having later.

[...]

PR #1038, opened 2024-05-22
Configure environment-specific services in dom0 via systemd

rocodes:
## Status

Ready for review

## Description of Changes

Refs https://github.com/freedomofpress/securedrop-workstation/issues/1004#issuecomment-2095970555
Supersedes #1031 

* Configure systemd-logind changes via logind.conf.d directory rather than editing logind.conf file directly
* Install logind powersetting changes via workstation RPM
* Disable logind powersetting changes on dev setups  [note: this means that during the time between installing the rpm and `sdw-admin --apply`, the power settings are in effect. But I think devs can handle that; usually those things follow pretty closely one after the other.]
* Write build flavour/environment flag (`dev`, `staging`, `prod`) to /var/lib/securedrop-workstation/ during Saltstack orchestration.
* use above mentioned flag file to conditionally disable logind changes in dev environments. 
* Convert xfce power settings to be handled the same way: service enabled during rpm installation, disabled conditionally at orchestration on dev machines, reverted fully if dom0 RPM is uninstalled.

## Testing

- [ ] CI passes
- [x] Uninstall SDW via `sdw-admin --uninstall` (or skip if fresh install scenario)
- [x] Create a desktop icon or shortcut on dom0 desktop.
- [x] Install rpm built from the tip of this branch: RPM installs successfully
- [x] `systemctl --user list-unit-files | grep enabled` shows `user-xfce-settings.service` and `user-xfce-icon-size.service`
- [x] `systemctl list-unit-files | grep logind` shows `logind-override-disable.service` as enabled
- [x] Reboot laptop. Upon login, desktop icon is enlarged and power settings are adjusted (no suspend/hibernate options available)
- [x] Closing laptop lid shuts off laptop
- [ ] Power on and run sdw-admin --apply [dev]. Provisioning completes successfully, and after reboot or restarting systemd (`systemctl daemon-reexec`), power setting changes are reverted
- [ ] sdw-admin --apply [staging or prod] completes successfully and power settings changes are in place after reboot
  - as discussed below, going from dev -> staging/prod does not reinstate power settings. 
- [x] Uninstalling rpm reverts icon size and power settings changes.

## Deployment

2. New installs


dom0/sd-dom0-files.sls
```diff
@@ -104,6 +104,27 @@ dom0-install-securedrop-workstation-dom0-config:
       - file: dom0-workstation-rpm-repo
 {% endif %}
 
+dom0-environment-directory:
+  file.directory:
+    - name: /var/lib/securedrop-workstation/
+    - user: {{ gui_user }}
```
legoktm, line 110: Does it need to be owned by the gui_user? I think root would be better unless there's a specific reason

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -115,6 +115,9 @@ install -m 644 files/securedrop-128x128.png %{buildroot}/usr/share/securedrop/ic
 
 install -m 755 -d %{buildroot}/opt/securedrop
 
+install -m 755 -d %{buildroot}/etc/systemd/logind.conf.d/
+install -m 644 files/10-logind_override.conf %{buildroot}/etc/systemd/logind.conf.d/
+install -m 644 files/logind-override-disable.service %{buildroot}/%{_userunitdir}/
```
legoktm, line 120: I think this would be better as a system unit and not user (not sure if a user unit can have a `Before=` a system unit)? the sdw-notify stuff is a system unit (AIUI) because it needs to use the user's GUI session.

legoktm (commented):
I like where this is going :)

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -168,6 +174,22 @@ find /srv/salt -maxdepth 1 -type f -iname '*.top' \
 # mkdir -p /tmp/sdw-migrations
 # touch /tmp/sdw-migrations/whonix-17-update
 
+# Adjust icon size (usability/accessibility)
+sudo -u $(sudo groupmems -l -g qubes) /opt/securedrop/update-xfce-settings adjust-icon-size ||:
+sudo -u $(sudo groupmems -l -g qubes) /opt/securedrop/update-xfce-settings disable-unsafe-power-management ||:
+
```
rocodes, line 180: (I'm not sure if I really like this, and I still have to test it. This would enable the power settings as soon as the rpm is installed, but then on dev machines the change wouldn't be reverted til the reboot after `sdw-admin --apply`.)

dom0/sd-clean-all.sls
```diff
@@ -117,3 +82,19 @@ disable-systemd-units:
       # Even with "runas", "systemctl --user" from root will fail unless we
       # tell it explicitly how to connect to the user systemd.
       - XDG_RUNTIME_DIR: /run/user/{{ gui_user_id }}
+
+# Reset desktop icon size to its original value
+dom0-reset-icon-size-xfce:
+  cmd.script:
+    - name: /usr/bin/securedrop/update-xfce-settings
+    - args: reset-icon-size
+    - runas: {{ gui_user }}
```
rocodes, line 91: (Eventually I want to clean this up and put it in the %preun of the rpm, but in the interests of time, it's still using Salt for the cleanup.)

dom0/sd-dom0-files.sls
```diff
@@ -104,11 +97,22 @@ dom0-install-securedrop-workstation-dom0-config:
...
+    - require:
+      - file: dom0-environment-directory
+
+dom0-write-environment-flag:
+  file.managed:
+    - name: /var/lib/securedrop-workstation/{{ d.environment }}
+    - mode: 644
+    - replace: False
+    - require:
+      - file: dom0-remove-old-environment-flag
```
rocodes, line 118: This is a flag file approach that allows us to use `ConditionPathExists` in systemd units to disable or enable services depending if we're in dev, prod, etc.  It's consistent with what we're using for vms; I have filed an issue to suggest we use .preset files instead (#1052 ), but this is an "on the way" option for now.

legoktm (changes requested):
Looks good, minor notes inline. Will test it now

legoktm:
It seems if I switch to a dev install and then back to staging/prod, rpm will not restore the deleted `10-logind_override.conf` file. (`sudo rpm -V securedrop-workstation-dom0-config` reports it as missing). https://serverfault.com/questions/1131250/can-you-force-replacement-of-files-marked-config-in-rpms says it's marked as `%config` and it's not possible to override that (like we do in Debian with our conffiles hack).

Maybe this is an OK limitation for now? that if you switch to dev you need to do a full provision from scratch to switch to staging/prod.

files/user-xfce-settings.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Enable XFCE customizations for SDW
+ConditionPathExists=|!/var/lib/securedrop-workstation/dev
```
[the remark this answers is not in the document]
rocodes: It's excessive, but what I was thinking about was failing to the behaviour we want (set the powersettings). I debated it and don't mind only doing the one check.

rocodes:
> It seems if I switch to a dev install and then back to staging/prod, rpm will not restore the deleted 10-logind_override.conf file. 

Ugh, good catch, and I don't love that. I always `sdw uninstall` if I'm switching between dev, staging, prod, just because unfortunately some of our cleanup logic is still in Salt and requires the uninstall command, but this could bite people.

Another solution could be putting a second, lower-priority configuration file that unsets this option on dev systems, instead of a service that deletes the other override file. (With the eventual goal of moving more logic out of salt and into the rpm so that we can cleanly uninstall and reinstall each time we re-provision, and so that there is symmetry in terms of where the setup/teardown actions happen, instead of them being in two places). 

Either way I'm not forcing this through if we're unhappy with it in some way - I'll make your suggested changes (thank you!) and let me know your thoughts re the way to handle the dev configuration.

files/logind-override-disable.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
```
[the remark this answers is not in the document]
rocodes: Addressed in https://github.com/freedomofpress/securedrop-workstation/pull/1038/commits/f845807e45adfb4b79ea3f7c0365ea35c9be0843

legoktm:
> Ugh, good catch, and I don't love that. I always `sdw uninstall` if I'm switching between dev, staging, prod, just because unfortunately some of our cleanup logic is still in Salt and requires the uninstall command, but this could bite people.

I think this is a reasonable limitation for now, I'm okay with documenting it in a bug as something to eventually fix. 

(Finishing up the test plan now)

files/securedrop-logind-override-disable.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Disable logind power management customization (dev systems only)
+ConditionPathExists=/var/lib/securedrop-workstation/dev
+Before=systemd-logind.service
+
+[Service]
+Type=oneshot
+ExecStart=rm -f /etc/systemd/logind.conf.d/10-logind_override.conf
```
legoktm, line 8:
```suggestion
ExecStart=rm -f /etc/systemd/logind.conf.d/10-securedrop-logind_override.conf
```

rocodes: oops. sorry

legoktm (approved):
LGTM \o/

Issue #1042, opened 2024-05-24
Refactor mime-handling configuration from Salt (at provisioning) to systemd (at boot)


deeplow:
* [x] I have searched for duplicates or related issues

## Description

Partly fixes #1004, by moving mime-time overrides onto systemd services for provisioning at boot.

## How will this impact [SecureDrop/SecureDrop Workstation users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?

No user implications.

## How would this affect the SecureDrop Workstation [threat model](https://github.com/freedomofpress/securedrop-workstation/#threat-model)?

Along with https://github.com/freedomofpress/securedrop-workstation/pull/1001, this assumes we are comfortable with runtime (boot-time) configuration of VMs' roles and services, https://github.com/freedomofpress/securedrop-workstation/issues/936#issuecomment-1930590304.

rocodes:
Yup I agree with you @deeplow, I'm musing about the future (and/or the proxy) but don't think this should be part of the immediate game plan, the changes you've proposed are solid

PR #1043, opened 2024-05-24
Removes MIME overrides out of salt and sd-proxy disposable

deeplow:
This includes several changes:
- move MIME overriding-logic into the securedrop-client repo to be implemented as systemd services for on-boot provisioning.
- makes sd-proxy into a disposable with sd-proxy-dvm as the template. This is thanks to the fact that we no longer need sd-proxy to keep any state as the only thing holding it was the need for MIME override provisioning.
- additionally this removes provisioning complexity since we can avoid salt provisioning from 3 qubes: sd-proxy, sd-devices and sd-viewer.

## Status

Ready for review.

## Description of Changes

Fixes freedomofpress/securedrop-workstation#1042

## Testing

Test plan at sister-PR https://github.com/freedomofpress/securedrop-client/pull/2033

## Deployment

2. New installs


legoktm (commented):
Overall looks good and works aside from sd-app mime handling. sd-proxy disposable is great \o/

Tests are failing, and in the other PR I've asked to rename the vm-config key for consistency, that'll need to be updated here too.

legoktm:
In my local testing there's one failure in `test_all_sd_vm_platforms` because it's no longer possible to boot the `sd-viewer` VM - see https://github.com/freedomofpress/securedrop-client/pull/2033#discussion_r1624973432. I think skipping `sd-viewer` there is reasonable for now if we decide the shutdown behavior is going to stay.

legoktm (approved):

PR #1067, opened 2024-06-05
Remove --keep-template-rpm

legoktm:
## Status

Ready for review

## Description of Changes

We no longer ship a template in an rpm package.

## Testing

* [ ] CI passes
* [ ] Visual review

## Deployment

Any special considerations for deployment? n/a, uninstall is also not an officially supported operation.


legoktm:
[Not tagging as 4.2 qa period because the flag is a no-op and uninstall is not officially supported, but that is how I noticed it]

rocodes (approved):
LGTM

PR #1068, opened 2024-06-05
Add VM test asserting no default_dispvm in sd-proxy and sd-proxy-dvm

rocodes:
## Status

Ready for review

## Description of Changes

What it says on the tin :) 

In 4.2 systems sd-proxy's default_disvpm is set to `None`. This is required for safe usage of the sd-app mimeapps.list file (aka the default file), which uses qvm-open-in-dvm for everything. Proxy should not be able to spawn any dvms, so add a test for this.

## Testing

- [x] CI passing

## Deployment

2. New installs


cfm (approved):
Looks good.  Thanks, @rocodes!

PR #1082, opened 2024-06-14
Test desired state of Linux Security Module configuration

legoktm:
## Status

Ready for review

## Description of Changes

Assert that:

* Debian VMs have AppArmor running
* Fedora VMs have SELinux running in enforce mode
* securedrop-proxy profile is in enforce mode
* sd-viewer profiles are in enforce mode

For verifying AppArmor profiles, take the same approach as the vm-config
expected keys test by defining the expected value in each VM's test and
having a base test implement the checking logic.

## Testing

* [x] CI passes

## Deployment

## Checklist

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -59,6 +59,11 @@ class SD_VM_Local_Test(unittest.TestCase):
         # expect.
         self.expected_config_keys = set()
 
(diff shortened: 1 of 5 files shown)


cfm (approved):
Nice.  I'm glad the set-based approach is proving adaptable!

PR #1084, opened 2024-06-17
Explicitly set default_dispvm on all VMs we manage

legoktm:
## Status

Ready for review

## Description of Changes

Enforce that all VMs we manage have `default_dispvm` set to None, except for sd-app and sd-devices, which are set to sd-viewer and sd-devices-dvm, respectively.

On the test side, instead of needing to specify it for each individual VM, verify `default_dispvm` is None for all of `@tag:sd-workstation`, with the above exceptions.

## Testing

* [ ] CI passes
* [ ] Open Qube Manager, visually check that all of the SDW VMs have an empty space in the "Default DispVM" column
* [ ] Or do the same with the CLI, `qvm-ls --tags sd-workstation --fields name,default_dispvm`
* [ ] Open client, verify files are still opened in a sd-viewer based dispVM

## Deployment

Any special considerations for deployment? should apply same to both

## Checklist

### DIFF
--- a/securedrop_salt/sd-app.sls
+++ b/securedrop_salt/sd-app.sls
@@ -24,6 +24,7 @@ sd-app:
     - prefs:
       - template: sd-small-{{ sdvars.distribution }}-template
       - netvm: ""
+      - default_dispvm: "sd-viewer"
     - tags:
       - add:
         - sd-client
--- a/securedrop_salt/sd-base-template.sls
+++ b/securedrop_salt/sd-base-template.sls
@@ -14,6 +14,8 @@ sd-base-template:
     - clone:
       - source: debian-12-minimal
       - label: red
+    - prefs:
+      - default_dispvm: ""
     - tags:
       - add:
         - sd-workstation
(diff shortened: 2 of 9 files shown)


rocodes:
I realized I actually led you astray here yesterday because I was conflating the dvm template vs the default_dispvm property.  sd-devices _and_ sd-devices-dvm can have default_dispvm `None`.

rocodes:
[Running through a local install and `make test` since workstation ci runner is having a rough day and will merge once done]

rocodes (approved):
Approving based on local testing and visual review. Thanks @legoktm

rocodes:
Just yanked it out of the queue because `sdw-admin --uninstall` failed for me and I want to investigate why

rocodes:
The error I saw during uninstall was in the `destroy_vm` script

`Command '['qvm-remove', '-f', 'sd-viewer-disposable']' returned non-zero exit status 2.`

Which... that's true, we don't have an "sd-viewer-disposable" vm, it's just named sd-viewer. Looking into whether this is related to this PR (i.e related to the naming of the systemwide dispvm) or not but I have one thing to finish up first.

rocodes:
Oh - it's just an issue with our test setup.

https://github.com/freedomofpress/securedrop-workstation/blob/517618391db14549305ba506285d1f341d53fa02/tests/base.py#L270-L272

rocodes:
Figured it out :) unrelated but funny, bug report coming presently

PR #1086, opened 2024-06-17
Replace sdw-login.py and .desktop file with systemd user unit

### LINKED ISSUE #1073: sdw-login should probably be started by a systemd unit instead of autostart entry

## Description

sdw-login is currently started by an autostart entry:

https://github.com/freedomofpress/securedrop-workstation/blob/95da5707b8af299e99ef1340c26155d8d07724b7/securedrop_salt/sd-dom0-files.sls#L66-L79

so it has a hardcoded sleep to make sure the desktop is set up properly:

https://github.com/freedomofpress/securedrop-workstation/blob/95da5707b8af299e99ef1340c26155d8d07724b7/files/sdw-login.py#L17-L21

probably it should be a systemd unit that triggers on login, with the necessary Requires/After stanzas instead of a sleep

rocodes:
## Status

Ready for review [will flip out of draft mode after testing]

## Description of Changes

Fixes #1073 

* Add user systemd unit to autostart sdw-updater (updater/login)
* remove sdw-login.py and autostart .desktop file

## Testing

- [ ] CI passing
- [ ] Manual testing: login/updater appear on start of user session

## Deployment

2. New installs


rocodes:
Commenting based on discussion: we may want to revisit whether autostart is a feature we want to support in any case

rocodes:
(Once https://github.com/freedomofpress/securedrop-workstation/pull/1088 lands I will update this PR to use the same approach and add this unit to the presets file)

rocodes:
If someone wants to continue with this they're welcome to take it over, but it got de-prioritized and has been sitting around for a while now.  Overall I think moving in the direction of packaging systemd files will be better than maintaining various scripts and customizations.

PR #1087, opened 2024-06-17
Actually test that logs are flowing to sd-log

legoktm:
## Status

Ready for review

## Description of Changes

Instead of checking that a folder exists, we now trigger a log entry in all VMs and then look for that log entry in the logs.
    
We also verify that sd-gpg, sd-log and sd-whonix logs are *not* flowing.

## Testing

* [ ] CI passes
* [ ] Visual review

## Deployment

## Checklist

### DIFF
--- a/tests/test_log_vm.py
+++ b/tests/test_log_vm.py
@@ -1,6 +1,9 @@
+import secrets
+import string
+import subprocess
 import unittest
 
-from base import SD_VM_Local_Test
+from base import CURRENT_DEBIAN_VERSION, SD_VM_Local_Test
 
 
 class SD_Log_Tests(SD_VM_Local_Test):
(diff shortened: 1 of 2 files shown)


tests/test_log_vm.py
```diff
@@ -29,18 +32,27 @@ def test_redis_service_running(self):
...
+            # The sudo call will make it into syslog
+            subprocess.check_call(["qvm-run", vm_name, f"sudo echo {token}"])
+
+        # Now look in the log files to find the token
+        for vm_name in ["sd-app", "sd-devices", "sd-proxy"]:
+            contents = self._get_file_contents(f"/home/user/QubesIncomingLogs/{vm_name}/syslog.log")
+            self.assertIn(token, contents)
+
+        # But nothing for sd-gpg
+        self.assertFalse(self._fileExists("/home/user/QubesIncomingLogs/sd-gpg/syslog.log"))
```
cfm, line 53: Do you see any point in parameterizing this now, such that either:
1. we have sets `logs_expected_from` and `logs_not_expected_from` that we enforce; or even
2. we have a set `log_expected_from` that we check strictly against all the VMs on the system?

(2) might well be overkill, but these are the two possibilities that occurred to me.

cfm (changes requested):

tests/test_log_vm.py
legoktm: Yeah, I think that's a good idea, let me see...

legoktm: Done! There's one todo about sd-viewer, because we can't actually run/start sd-viewer, we have to use a disposable based on it. There's a workaround for it in the sd-viewer tests, but it's not reusable yet. Once we're pytest and can have easily shared fixtures, it should be trivial.

legoktm:
The test failure looks legit...I suspect there could easily be a race condition since we run the command and then immediately look for it in the logs.

cfm:
Looks like this is now conflicted, @legoktm, but I'll be happy to re-review and merge tomorrow.

legoktm:
Rebased.

cfm (approved):
Thanks, @legoktm.

Issue #1052, opened 2024-05-30
Use preset files for conditionally enabling/disabling systemd services


rocodes:
[Preset files](https://www.freedesktop.org/software/systemd/man/latest/systemd.preset.html) allow for the %post of a package installation to simply call `systemctl preset mycoolservicename.service` no matter what, and the service will be conditionally enabled/disabled depending on the contents of the preset file (or enabled if no preset file is present).

For example, instead of creating `/var/lib/securedrop/{dev,staging,prod}` in dom0 and using ConditionPathExists to test what type of build we're in, a more systemd-ish way might be to ship `/usr/lib/systemd/user-preset/sd-config.preset` that, in dev environments, disables our custom services:

```
#/usr/lib/systemd/user-preset/10-sd-preset.conf
disable user-xfce-settings.service
```

This approach may be preferable in cases where the other option (`ConditionPathExists`) involves dropping a flag file/directory during orchestration and then checking for it at boot.

legoktm:
TIL about presets, but after reading the docs you link and poking around in `/lib/systemd/system-preset` in dom0, they seem like a good fit for our dom0 services. And they'll let us use https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_systemd in our .spec file.

rocodes:
I'm coming back to this now that we're thinking of having a separate keyring + repo bootstrap package, because it would be easy to ship a dev-only package that drops a preset file for the services we want to disable for dev machines (ensuring that services default to being enabled, which we'd want on prod).

PR #1088, opened 2024-06-18
Don't enable sdw-notify.timer unit via Salt. Use systemd-rpm-macros and user unit preset file.

### LINKED ISSUE #1006: `update-xfce-settings` is flaky in CI

## Description

In state `dom0-adjust-desktop-icon-size-xfce`, `update-xfce-settings` can flake:

```
Failed to init libxfconf: Could not connect: No such file or directory.
```

## Steps to Reproduce

https://ws-ci-runner.securedrop.org/2024-04-25-192246767193-f65d3795971be6763a6b65272f144fcc76ff9e53-Qubes_4.2_B-update_20240425044540.log.txt

## Expected Behavior

CI passes as it did on 6ce13c20fd3ded4e199006998f69658f0bed8900 in #1002.

## Actual Behavior

```
INFO:[2024-04-25-20:06:34:122376]           ID: dom0-adjust-desktop-icon-size-xfce
INFO:[2024-04-25-20:06:34:122412]     Function: cmd.script
INFO:[2024-04-25-20:06:34:122444]         Name: salt://update-xfce-settings
INFO:[2024-04-25-20:06:34:122478]       Result: False
INFO:[2024-04-25-20:06:34:122513]      Comment: Command 'salt://update-xfce-settings' run
INFO:[2024-04-25-20:06:34:122544]      Started: 20:06:25.669609
INFO:[2024-04-25-20:06:34:122579]     Duration: 146.201 ms
INFO:[2024-04-25-20:06:34:122614]      Changes:   
INFO:[2024-04-25-20:06:34:122649]               ----------
INFO:[2024-04-25-20:06:34:122680]               pid:
INFO:[2024-04-25-20:06:34:122715]                   6919
INFO:[2024-04-25-20:06:34:122746]               retcode:
INFO:[2024-04-25-20:06:34:122781]                   1
INFO:[2024-04-25-20:06:34:122816]               stderr:
INFO:[2024-04-25-20:06:34:122848]                   Failed to init libxfconf: Could not connect: No such file or directory.
INFO:[2024-04-25-20:06:34:122884]               stdout:
INFO:[2024-04-25-20:06:34:122916]                   update-xfce-settings: Adjusting icon size for user user to 64 px
```

## Comments

There aren't many citations of this exact error message versus other forms of `Failed to init libxfconf`.  The general recommendation seems to be to set `$DBUS_SESSION_BUS_ADDRESS`, which we already do:

https://github.com/freedomofpress/securedrop-workstation/blob/f65d3795971be6763a6b65272f144fcc76ff9e53/dom0/update-xfce-settings#L29-L32

rocodes:
## Status

Ready for review

## Description of Changes
Towards #1052
Fixes #1006 

- Use systemd-rpm-macros in rpm .spec and enable user units according to a preset file.  
- Manage sdw-notify.timer (and all systemd user) units via systemd macros in rpm %post (and %preun) instead of via Salt, and stop using `--global` for enabling user units.

## Testing

- [x] Visual review
- [x] CI
- [ ] Manual testing (see below)
- Start from an environment without a dom0 workstation config and without any securedrop systemd services configured in dom0. (This could be an `--uninstall`, or it could be sneakily removing the dom0 rpm and disabling then removing (deleting) the sd systemd services `securedrop-user-xfce-icon-size`, `securedrop-user-xfce-settings`, and `sdw-notify.timer` and `sdw-notify.service`. Reload the systemd user daemon (systemctl --user daemon-reload) and ensure those services are not listed.
- Build an rpm from the tip of this branch (`make build-rpm`), and copy it into dom0, then install it. Reboot the laptop.  
- [ ]  `systemctl --user status` for each of `sdw-notify.timer`, `securedrop-user-xfce-icon-size`, and `securedrop-user-xfce-settings` shows them as enabled, and `systemctl status securedrop-logind-override-disable` enabled (on dev systems the service will run; on staging and prod systems the expected behaviour is that the service will fail a ConditionPath check and be disabled.)
- [ ] When the rpm is uninstalled, securedrop-related systemd units (user and system) are all removed.
- Upgrade test plan: starting from a preinstalled securedrop-workstation-dom0-config and an existing workstation setup, confirm that desired services are loaded, as above.
- Build rpm from the tip of this branch and install over top of your existing setup.
- [x] RPM installation completes successfully
- [x] Desired services are still in place, as above. 

## Deployment

2. New installs


rocodes:
ci failure unrelated to this change - 
https://ws-ci-runner.securedrop.org/2024-11-28-190455591162-67eb6703adec3b9e60946c269ce0ae7d7c6b26a2-Qubes_4.2_A-update_20241128043845.log.txt 
`An error was encountered while installing package(s): E: Release file for tor+https://deb.debian.org/debian-security/dists/bookworm-security/InRelease is not valid yet (invalid for another 1min 10s).`

rocodes:
Hey, thanks for the review. Re: uninstall: I kept the removal logic in the salt run, so running the `uninstall` command (and rebooting or reloading systemd) should address the timer.  But I flipped back to draft mode because I think I have a cleaner way of addressing this change and the comment I made about enabling for all users too. (See https://github.com/freedomofpress/securedrop-workstation/issues/1052)

rocodes:
I have updated this PR in order to address the CI failures (which presumably could also imply real-world provisioning failures) that @deeplow and others have noticed a few different places (#1006, https://github.com/freedomofpress/securedrop-workstation/pull/1205) in a more consistent way, by avoiding running systemd user commands via salt orchestration. 

This PR does introduce an additional build dependency in our rpm .spec file (systemd-rpm-macros). These macros ship already with systemd.

rocodes:
I've stepped through a basic version of the test plan (get rid of dom0 rpm and securedrop systemd user units, install this rpm, verify units are loaded)

rocodes:
(as was rightly pointed out, now that this is coming in post 4.2 fresh install, I need to add an upgrade test plan!)

rocodes:
Notes re upgrade scenario:

The sdw-notify.service and .timer files were already managed by the workstation config rpm and installed at rpm install time, so there will be no conflict for folks upgrading. The difference is when/how the notify timer is enabled (rpm install time vs `sdw-admin --apply` time), and when/how it's disabled (rpm uninstall vs `sdw-admin --uninstall`).

This means, if folks install the workstation-dom0-config rpm and then _don't_ install SDW, they will have an edge case situation where they are notified to apply updates, but the SDW VMs haven't been provisioned yet.  That is definitely an oddity, but it's consistent with other oddities about "configure at rpm install time" vs "configure during `apply` time" that we have already previously introduced (for example, shipping the .desktop file, changing icon size, provisioning files to the /srv/salt/securedrop_salt directory, although that last one is less obviously user-facing).

My opinion is that we should change/address that behaviour all together cause it's an edge case, but that it's acceptable right now. (For example, one way to address it would be to have the rpm post immediately pop up a GUI graphical installer ;) ).  I'm documenting this so that no one is surprised, but noting that it's consistent with other changes we made as part of 4.2 compat.

cfm (approved):
I've tested the upgrade from from an up-to-date v1.0.0 and can confirm:

- [x] `systemctl --user status sdw-notify.timer`
- [x] `systemctl status securedrop-user-xfce-icon-size`
- [x] `systemctl status securedrop-user-xfce-settings` correctly on:
    https://github.com/freedomofpress/securedrop-workstation/blob/af656595a237fddbf4358d52a5f1b58d79c10fed/files/securedrop-user-xfce-settings.service#L3-L5
    —with `/var/lib/securedrop-workstation/dev` present.

PR #1089, opened 2024-06-18
Use specific unittest asserts when possible

legoktm:
## Status

Ready for review

## Description of Changes

Using assertEqual(x, y) over assertTrue(x == y) means that unittest can show us a proper diff of the two things that aren't equal, instead of just saying False isn't True.

## Testing

* [ ] CI passes

## Deployment

## Checklist

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -138,7 +138,7 @@ class SD_VM_Local_Test(unittest.TestCase):
         import difflib
 
         print("".join(difflib.unified_diff(remote_content, content)), end="")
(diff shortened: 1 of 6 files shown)


legoktm:
This will probably conflict with other stuff in flight so happy to hold it until later.

zenmonkeykstop:
Just rebased it - we can pull it in now and rebase as necessary elsewhere.

zenmonkeykstop (approved):
LGTM

PR #1093, opened 2024-06-18
Test that apt interprets our .sources files correctly

legoktm:
## Status

Ready for review

## Description of Changes

Run `apt-get indextargets` to verify that our .sources files are valid and apt is able to properly interpret them.

I didn't implement actually running apt to check that we can download files because that happens during provisioning already.

## Testing

* [x] CI passes

## Deployment

Any special considerations for deployment? n/a, test only

## Checklist

### DIFF
--- a/tests/test_vms_platform.py
+++ b/tests/test_vms_platform.py
@@ -3,7 +3,13 @@ import os
 import subprocess
 import unittest
 
-from base import CURRENT_FEDORA_TEMPLATE, SD_VMS
+from base import (
+    CURRENT_FEDORA_TEMPLATE,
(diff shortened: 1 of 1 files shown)


cfm:
This is now conflicted.  I'll be happy to re-review once that's resolved.

legoktm:
Rebased.

cfm (approved):
Thanks, @legoktm.  Since I'm largely deferring to @deeplow's previous review, I hadn't noticed earlier your anti-superstition change here: if what we're testing is a property of templates, _just test the templates_, not every VM that necesarily derives that property from its template.  Nice.

Issue #1091, opened 2024-06-18
A VM test adds VMs to system, breaking sdw-admin --uninstall


rocodes:
## Description

Noticed during review for #1084, where I provisioned sdw, ran `make test` then `sdw-admin --uninstall`

We have a test that creates VMs based on a dvm template:

https://github.com/freedomofpress/securedrop-workstation/blob/517618391db14549305ba506285d1f341d53fa02/tests/base.py#L267-L280

However, this breaks sdw-admin --uninstall, which expects named dispvms but not vms named xx-xx-disposable (see eg https://github.com/freedomofpress/securedrop-workstation/pull/1084#issuecomment-2176926227 )

## Steps to Reproduce

`make test` in dom0 then `sdw-admin --uninstall` 

## Expected Behavior

uninstall completes successfully; running `make test` does not modify system state

## Actual Behavior

as linked 

## Comments

- Let's clean up any VMs created during testing, and/or discuss whether we want to create vms during testing

rocodes:
Fixed by https://github.com/freedomofpress/securedrop-workstation/pull/1095

PR #1095, opened 2024-06-24
Ensure outdated templates are removed

deeplow:
## Status

Ready for review

## Description of Changes

Inverts the logic of test_expected from checking that the VMs we wanted do exist...
https://github.com/freedomofpress/securedrop-workstation/blob/23529537b484d801ad921e08cc43ee867fce5549/tests/test_vms_exist.py#L22-L25

...to one where we account for all the VMs tagged with `sd-workstation` or any other VM that we may have missed.

## Testing

This is a test already.

## Deployment

No deployment considerations.

Docs not required.

### DIFF
--- a/tests/test_vms_exist.py
+++ b/tests/test_vms_exist.py
@@ -6,6 +6,8 @@ from base import WANTED_VMS
 from qubesadmin import Qubes
 
 DEBIAN_VERSION = "bookworm"
+SD_TEMPLATE_LARGE = f"sd-large-{DEBIAN_VERSION}-template"
+SD_TEMPLATE_SMALL = f"sd-small-{DEBIAN_VERSION}-template"
 
 
 class SD_VM_Tests(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


cfm:
(Assigning per Slack discussion just for visibility; feel free to reassign as needed.)

tests/test_vms_exist.py
```diff
@@ -20,17 +25,17 @@ def tearDown(self):
         pass
 
     def test_expected(self):
-        vm_set = set(self.app.domains)
-        for test_vm in WANTED_VMS:
-            self.assertIn(test_vm, vm_set)
+        sdw_tagged_vm_names = [vm.name for vm in self.sdw_tagged_vms]
+        expected_vms = set(SD_VMS + SD_DVM_TEMPLATES + SD_TEMPLATES)
+        self.assertEqual(set(sdw_tagged_vm_names), set(expected_vms))
```
rocodes, line 30: Just noting for our future selves: Sometimes the set of sd-tagged vms isn't exactly equal to the set of VMs expressed here, but they're all edge cases and shouldn't interfere:
- devs may tag their sd-dev vm @tag:sd-workstation if they follow our (old?) developer guidelines. This shouldn't be a problem though because those environments aren't where we should be running `make test` since they're not indicative of real conditions
- prod instances may(?) configure additonal VMs and tag them with @sd-workstation to take advantage of our strict rpc policies. I don't think this is happening, but in any case, prod setups shouldn't be running dom0 tests.

rocodes (approved):
Visual review LGTM, CI LGTM, laughed at 

> ## Testing
> This is a test already.

:) 

Thanks deeplow. I think this fixes #1091 too.

Issue #1097, opened 2024-06-24
sd-viewer and sd-app not marked as internal


deeplow:
* [x] I have searched for duplicates or related issues

## Description

sd-viewer is not marked as internal on a prod system, even though it supposedly [was set to be internal](https://github.com/freedomofpress/securedrop-workstation/pull/857/files#diff-e602638d95f8e7e1bf97f45e706acb538ce0d3f7dd0a7a59114c29ef3792b4e1).

## Steps to Reproduce

Setup a prod environment

## Expected Behavior
`sd-viewer` is internal
`sd-app` is internal

## Actual Behavior

`sd-viewer` is not internal
`sd-app` is not internal

## Comments

This is due a duplicated yaml entry.

PR #1098, opened 2024-06-24
Make sd-viewer and sd-app internal (again)

deeplow:
## Status

Ready for review

## Description of Changes

Fixes https://github.com/freedomofpress/securedrop-workstation/issues/1097

## Testing

n/a (test included)

## Deployment

Any special considerations for deployment? no

Not required.

### DIFF
--- a/tests/test_vms_exist.py
+++ b/tests/test_vms_exist.py
@@ -12,6 +12,9 @@ from base import (
 )
 from qubesadmin import Qubes
 
+with open("config.json") as f:
+    CONFIG = json.load(f)
+
 
 class SD_VM_Tests(unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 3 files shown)


tests/test_vms_exist.py
```diff
@@ -7,6 +7,9 @@
 
 DEBIAN_VERSION = "bookworm"
 
+with open("config.json") as f:
```
legoktm, line 10: Note to self: this is a little duplicative of the `self.config` but otherwise it can't be used in the skipIf().

tests/test_vms_exist.py
```diff
@@ -24,6 +27,16 @@ def test_expected(self):
         for test_vm in WANTED_VMS:
             self.assertIn(test_vm, vm_set)
 
+    @unittest.skipIf(CONFIG["environment"] != "prod", "Skipping on non-prod system")
```
legoktm, line 30: As a future improvement, instead of skipping entirely non non-prod systems, you could have those tests assert that features.get("internal") is != 1.

legoktm (approved):
Rebased, LGTM. I have suggested an improvement to the test inline, but that can be addressed in a follow-up.

legoktm (approved):
Rebased once more...

PR #1128, opened 2024-07-08
Fix Updater only restarting sys-* qubes

### LINKED ISSUE #1127: Investigate updater not restarting some qubes
## Description

The 4.2 updater integration does restart some qubes, but not all. This is undesirable behavior since we want to enforce all qubes to restart.

## Steps to Reproduce

SDW 1.0.0rc2

## Expected Behavior

All running qubes based on templates that were updated should have been restarted.

## Actual Behavior

Only a hand-full of qubes actually restarted.

| qubes which got restarted | qubes (still) running after updater finished with pending updates | 
|---|---|
| which-restarted | which-restarted2 |

## Comments

- when I encountered this I confirmed that `sd-{small,large}` did in fact have updates pending (via `apt list --upgradable`)
- If this ends up being an upstream bug, for the time being we should just force-restart all workstation-based app qubes.

deeplow:
After [an update to qubes-vm-update](https://github.com/QubesOS/qubes-core-admin-linux/commit/93078d0b) (which the SDW updater uses under the hood) `--restart` was changed to only restart `sys-*` qubes. The new qubes-vm-update flag --apply-to-all has the behavior we intend.

## Status

Ready for review

## Description of Changes

Fixes #1127 by replacing `qubes-vm-update`'s flags from `--restart`  to `--apply-to-all`

## Testing

Run the updater and ensure `sd-app`, `sd-log`, `sd-gpg`, etc. (all app qubes) are restarted after a successful update process.

## Deployment

no special considerations.

  - :warning: not on a prod

### DIFF
--- a/sdw_updater/Updater.py
+++ b/sdw_updater/Updater.py
@@ -159,7 +159,7 @@ def apply_updates_templates(templates=current_templates, progress_callback=None)
 def _start_qubes_updater_proc(templates):
     update_cmd = [
         "qubes-vm-update",
-        "--restart",  # Enforce app qube restarts
+        "--apply-to-all",  # Enforce app qube restarts
         "--show-output",  # Update transaction details (goes to stdout)
         "--just-print-progress",  # Progress reporting (goes to stderr)
         "--targets",


legoktm:
I manually applied this change to a 1.0.0rc2 system and then:
* Downloaded an old securedrop-client deb, installed it into the small template so there would be 1 package needing an upgrade.
* Deleted dom0's `~/.securedrop_updater`
* Rebooted Qubes
* Updater notification showed up, proceeded. sd-gpg, sd-proxy and sd-log VMs were running at this point
* sd-small-bookworm-template upgraded and shutdown
* sd-gpg, sd-log, sd-proxy shutdown
* sd-app starts, which triggers start of sd-log
* Updater concludes, lets me "Continue" to start client (no dom0 updates)

Is that the expected behavior, that it shuts down all the VMs instead of restarting them? And then the updater manually starts sd-app?

legoktm:
> Is that the expected behavior, that it shuts down all the VMs instead of restarting them?

Based on https://github.com/QubesOS/qubes-core-admin-linux/blob/main/vmupdate/vmupdate.py#L384-L387, I believe so. Only `servicevm`s are restarted, the others are just shutdown.

legoktm:
(SDW CI failure is https://github.com/freedomofpress/securedrop-workstation/issues/1006 - OK to ignore.)

legoktm (approved):

Issue #1145, opened 2024-07-10
Mark sd-proxy as a service VM so the updater will restart it


legoktm:
* [x] I have searched for duplicates or related issues

## Description

The updater shuts down all VMs with template updates, but only restarts the ones marked as `servicevm` (usually `sys-*`), which sd-whonix is.

We should do the same for sd-proxy, which effectively is also a stateless service VM that just provides networking for the client. 

## Steps to Reproduce

Run the updater, which shuts down sd-proxy. Open the client and log in.

## Expected Behavior

Log in request is sent immediately and you log in.

## Actual Behavior

Need to wait for sd-proxy to start so the RPC call can fire off the request. If this takes too long (plus tor weather), it's possible the TOTP will expire before the request makes it to the server.

## Comments

I assume we can trivially set this via salt.

PR #1146, opened 2024-07-11
Set named dispvm sd-proxy as a servicevm

rocodes:
## Status

Ready for review

## Description of Changes

Fixes https://github.com/freedomofpress/securedrop-workstation/issues/1145

Set named qubes feature `servicevm` for named dispvm sd-proxy during Salt run.

Note: `servicevm` is approximately taken to mean "provides network" (https://github.com/QubesOS/qubes-desktop-linux-menu/blob/1ee90aa09903dbd470f540b22e48d0a500feff6e/qubes_menu/vm_manager.py#L122-L125), so this will show that sd-proxy "provides a network" in the appmenu (the network-connected icon shows up beside the qube). However, unlike qubes that really _do_ provide a network, we don't set the qvm-pref `provides_network` to `True` for sd-proxy, so it won't show up as a networking option when new qubes are created.  (This is good; we wouldn't want that.)

So we may be acting like outliers here by setting the qvm-feature (for menu + restart behaviour) but not the qvm-pref (for what capabilities it advertises to other Qubes).

## Testing

- [ ] CI
- [ ] Visual review
- [ ] Feature name is correct: verify with eg https://github.com/QubesOS/qubes-desktop-linux-menu/blob/1ee90aa09903dbd470f540b22e48d0a500feff6e/qubes_menu/vm_manager.py#L50
- [x] sd-proxy shows in Service section of menu with this feature set (tested via `qvm-features sd-proxy servicevm 1`)

## Deployment

2. New installs

### DIFF
--- a/securedrop_salt/sd-proxy.sls
+++ b/securedrop_salt/sd-proxy.sls
@@ -54,6 +54,7 @@ sd-proxy-create-named-dispvm:
         - service.securedrop-mime-handling
       - set:
           - vm-config.SD_MIME_HANDLING: default
+          - servicevm: 1
       {% if d.environment == "prod" %}
           - internal: 1
       {% endif %}


legoktm:
> Note: servicevm is approximately taken to mean "provides network" (https://github.com/QubesOS/qubes-desktop-linux-menu/blob/1ee90aa09903dbd470f540b22e48d0a500feff6e/qubes_menu/vm_manager.py#L122-L125), so this will show that sd-proxy "provides a network" in the appmenu (the network-connected icon shows up beside the qube).

TIL, thanks for finding that.

> So we may be acting like outliers here by setting the qvm-feature (for menu + restart behaviour) but not the qvm-pref (for what capabilities it advertises to other Qubes).

Ack, we should probably check in with the Qubes team that this won't have any problematic consequences then.

rocodes:
The property being reset if "provides network" is toggled (by a user manually?) is good to know as we move towards updater changes, but since the [dom0 highstate](https://github.com/freedomofpress/securedrop-workstation/blob/b274744ee27c37c1f7af0cbeb6e686551743d530/sdw_updater/UpdaterApp.py#L257) still runs right now and would enforce that property, I think we're okay until we make any significant updater changes.

PR #1159, opened 2024-08-09
Move provision-all to sdw-admin.py

### LINKED ISSUE #539: Call qubesctl from securedrop-admin Python code

deeplow:
## Status

**Work in progress** - this is just a literal conversion. I will review to ensure none of the semantics were lost and pythonize some of the functions, where appropriate.

## Description of Changes

Fixes #539

## Testing

Successful case:
- [ ] run `make dev` and compare the output to `make dev` from main. Only debugging information should differ (something similar to what I did [here](https://github.com/freedomofpress/securedrop-workstation/pull/1159#issuecomment-2311041468))
- [ ] run `make test`

Failure injection:
- force some salt commands and confirm the script reported the appropriate failure 

## Deployment

Any special considerations for deployment?

Not that I can think of. If anything, it should improve deployment in case of failures. I didn't find any references to `provision-all` other than the ones modified, not even in the docs. But it is always possible some developer's workflows will break if they used `provision-all`.


files/sdw-admin.py
```diff
@@ -82,14 +86,167 @@ def copy_config():
...
     try:
-        subprocess.check_call([os.path.join(SCRIPTS_PATH, "scripts/provision-all")])
+        subprocess.check_call(args)
     except subprocess.CalledProcessError:
-        raise SDWAdminException("Error during provision-all")
+        raise SDWAdminException(f"Error while running {' '.join(args)}")
 
-    print("Provisioning complete. Please reboot to complete the installation.")
+
+def qubesctl_call(step_description: str, args: List[str]):
```
rocodes, line 111: (You might want a skip_dom0 parameter set to True, that can optionally be overridden, but that fails "safe" and defaults to skipping dom0)

rocodes:
Thanks @deeplow  - took a quick look through, have not tested it out since you mention WIP + draft mode, but very happy to do so if and when.

A couple thoughts: 
* we only have a few different Salt-related steps we run through (run the highstate with or without a list of targets, run a saltstate, basically), so you might want to have those be the methods that you create, rather than having separate methods that each pass a similar list to `qubesctl_call`. That way looking at your `provision_all()` it will be clear what's happening:

```
SD_STATE_PREFIX="securedrop_salt."
def provision_all():
    apply_state({SD_STATE_PREFIX}sls_name1, args...)
    apply_state({SD_STATE_PREFIX}sls_name2, args...)
    apply_highstate(vmname, args...)
    # etc
```

* Same thought for the rebooting vms or syncing appmenus - we may be able to parameterize those methods further 
* I wonder if you can directly use the qubesadmin api for the `run_cmd` commands?
* We'll probably want to figure out error handling (one nice thing about your changes is how easy it will be to tell which step was an error :) )
* We might want check_output instead of check_call; we might want to be able to configurably see that output, for cases where we're trying to debug something, or to print the output of a failed state. I haven't looked into this though because it's quite a lot of output and could (potentially?) block the subprocess pipe - maybe you have thoughts here   
 
In case it's helpful: I also have some related work in progress from the retreat ([my branch - very much WIP/probably not runnable yet](https://github.com/freedomofpress/securedrop-workstation/commit/39bab1ab98d02c3dc0a8992522af1371d0656bf4#diff-87b136d519f9106582830b3b69ef3bbfe62788e8dced9a92241eac4a82662210)) - there, I took the approach of "every qubesctl call is a subprocess call" just in the interests of focusing on the gui installer part, but it would be quite easy to integrate your changes there if and when. :)

legoktm:
> I did briefly try that, but unfortunately QubesVM.shutdown() is missing a `wait` parameter, but I'll explore other ways of doing that.

Feel free to copy what Kev+I did in https://github.com/freedomofpress/securedrop-workstation-ci/blob/218b571d3047eeee53e391245e4077b82bb6790c/dom0/runner.py#L71

files/sdw-admin.py
```diff
@@ -80,16 +84,124 @@ def copy_config():
...
 
+def provision_and_configure():
+    """
+    Applies the salt state.highstate on dom0 and all VMs
+    """
+    provision("Provisioning Fedora-based system VMs", "securedrop_salt.sd-sys-vms")
+    provision("Provisioning base template", "securedrop_salt.sd-base-template")
+    configure("Configuring base template", ["sd-base-bookworm-template"])
+    provision_all()
+    configure("Configuring logging VMs early", ["sd-small-bookworm-template"], restart=["sd-log"])
```
cfm, line 95: Nit: suggest clarifying, something like:

```suggestion
    configure("Configuring template for log sink before anything else", ["sd-small-bookworm-template"], restart=["sd-log"])
```

But feel free to ignore this if you want to keep this pull request as a strict port.

cfm (approved):
Very nice.  `make dev` succeeds and `make test` passes as expected.

I'm approving this with one nit inline.  If you agree, I'll re-approve and merge; otherwise feel free to merge this right away.

cfm:
(It also occurs to me that provisioning versus configuration is emerging as a clear architectural principle we should document, including in how it's implemented in this script. :-)

cfm (approved):
Thanks, @deeplow.  Approving after 507b903f9da2ce412a8c763c37a0be29773c324e based on <https://github.com/freedomofpress/securedrop-workstation/pull/1159#pullrequestreview-2298876865>.

PR #1165, opened 2024-08-27
Replace salt updater in sd-sys-vms

### LINKED ISSUE #1166: Newly installed Fedora template is not getting updates right after install

## Description

Due to [the wrong variable](https://github.com/freedomofpress/securedrop-workstation/blob/main/securedrop_salt/sd-sys-vms.sls#L35) `fedora-40` was being updated instead of the newly fetched `fedora-40-xfce`. `fedora-40` does not exist in default systems and the salt updater was silently failing.

This should not affect current installations, since in 4.2.2 already came with fedora-40-xfce and thus no actual template upgrades have occurred.

## Steps to Reproduce

No need to reproduce. Visual inspection suffices.

## Expected Behavior

Downloaded template is updated immediately

## Actual Behavior

Silent failure.

deeplow:
Replace salt updater with qubes-vm-update in sd-sys-vms

## Status

Work in progress

## Description of Changes

Fixes #1166.

## Testing

Run `make dev` and ensure when running sd-sys-vms step the qube started and fully updated.

## Deployment

2. New installs

### DIFF
--- a/securedrop_salt/sd-sys-vms.sls
+++ b/securedrop_salt/sd-sys-vms.sls
@@ -32,7 +32,7 @@ set-fedora-template-as-default-mgmt-dvm:
 # If the VM has just been installed via package manager, update it immediately
 update-fedora-template-if-new:
   cmd.wait:
-    - name: qubes-vm-update --quiet --force-update --targets {{ sd_supported_fedora_version }}
+    - name: qubes-vm-update --quiet --force-update --targets {{ sd_fedora_base_template }}
     - require:
       - cmd: dom0-install-fedora-template
       # Update the mgmt-dvm setting first, to avoid problems during first update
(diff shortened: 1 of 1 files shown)


rocodes:
Hey @deeplow, thanks for preparing this, couple thoughts:
- I agree with you that it's too verbose - 8800+ lines of console output when I tried this. I'd remove the `--just-print-progress` flag (I think that's the super noisy one? we'll use it maybe with a gui integration but not here?).
- Haven't looked closely but if we could deprecate the ``{{sd_supported_fedora_version}}`` variable, or at mimimum add a comment (or replace it with a version-only variable to reduce the risk of confusion) that would be great
- I suggest using the  `--force-update` flag (force checking for new updates even if there aren't any). I think if we have just downloaded the template, the qubes-update-check may not have run, so for provisioning we may want to force a check,  let me know what you think.

rocodes (changes requested):
(per https://github.com/freedomofpress/securedrop-workstation/pull/1165#issuecomment-2318373617)

rocodes (approved):
In my head this was a different PR, must've been confusing when I said I needed a backup reviewer at standup :) LGTM.

PR #1183, opened 2024-09-24
Verify PGP key embedded in .sources files are correct

legoktm:
## Status

Ready for review

## Description of Changes

It's kind of hard to manually test the PGP key inside apt's .sources files are the ones we intended, so write a test to do it. This is copied from the one I wrote in securedrop-client, and adds coverage for the apt-test sources file.

This isn't a launcher test, but it needs to have external dependencies installed, so this is the best place for it now until we have unit tests for the rest of the provisioning code.
## Testing

* [ ] CI passes
* [ ] Visual review (key fingerprint in test is correct)

## Deployment


zenmonkeykstop (approved):
LGTM

PR #1188, opened 2024-10-02
Automatically log into the client application

deeplow:
## Status

Ready for review

## Description of Changes

Most of the time we're developing the client, we are not testing the login screen, and just want to quickly log in. With a combination of xdotool and oathtool, we can script typing the correct values and login.

It's a bit fragile for two reasons. First, we're just blindly typing, so if some other window takes focus or switch, it'll be typed into that window. Second, we're kicking off a background process that sleeps, launching the window in the foreground, and then the background process will do the typing into the foreground.

Adapted from work by @legoktm: https://github.com/freedomofpress/securedrop-client/pull/2237/

## Testing

- [ ] visual review
- [ ] run `make run-client`
- [ ] confirm that it asks you to install dependencies in dom0
- [ ] install dependencies (`xdotool` and `oathtool`)
- [ ] run again `make run-client`

## Deployment

dev-facing - no deployment considerations 

- [x] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-workstation-docs) for these changes, or will do so later

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,11 +7,6 @@ CONTAINER := $(if $(shell grep "Thirty Seven" /etc/fedora-release),,./scripts/co
 
 HOST=$(shell hostname)
 
-# Client autologin variables
-CLIENT_TOTP=$(shell oathtool --totp --base32 JHCOGO7VCER3EJ4L --start-time "$(shell date -d "+5 seconds" +"%Y-%m-%d %H:%M:%S")")
-XDOTOOL_PATH=$(shell command -v xdotool)
-OATHTOOL_PATH=$(shell command -v oathtool)
-
 assert-dom0: ## Confirms command is being run under dom0
 ifneq ($(HOST),dom0)
 	@echo "     ------ Some targets of securedrop-workstation's makefile must be used only on dom0! ------"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -155,6 +160,27 @@ test-whonix: test-prereqs ## Runs tests for SD Whonix VM
...
+ifeq ($(XDOTOOL_PATH),)
+	@echo $(XDOTOOL_PATH)
+	@echo 'please install xdotool with "sudo qubes-dom0-update xdotool"'
+	@false
+endif
+ifeq ($(OATHTOOL_PATH),)
+	@echo 'please install oathtool with "sudo qubes-dom0-update oathtool"'
+	@false
+endif
+	sdw-updater
```
legoktm, line 174: Maybe `qvm-run sd-app securedrop-client` so it doesn't accidentally trigger the updater and type into it?

Makefile
```diff
@@ -7,6 +7,11 @@ CONTAINER := $(if $(shell grep "Thirty Seven" /etc/fedora-release),,./scripts/co
 
 HOST=$(shell hostname)
 
+# Client autologin variables
+CLIENT_TOTP=$(shell oathtool --totp --base32 JHCOGO7VCER3EJ4L --start-time "$(shell date -d "+5 seconds" +"%Y-%m-%d %H:%M:%S")")
```
legoktm, line 11: I would just inline this below instead of putting it at the top

legoktm (commented):
This is awesome \o/ left a suggestion!

legoktm (approved):
Thanks!

Issue #945, opened 2024-02-09
Avoid manual downloading of dom0 rpm during bootstrapping


legoktm:
I want to split out the first step from #942 as its own separate issue:
>  The rough flow is as follows:
> 1. configure a network-attached VM to download the dom0 rpm, copy and install it

https://workstation.securedrop.org/en/stable/admin/install.html#download-and-install-securedrop-workstation describes a lot of steps related to verifying the key, manually writing a .repo file, verifying and moving the RPM around and then finally installing it.

My proposed solution would be to have a `securedrop-keyring` RPM that is shipped in the Qubes repositories, so you can simply run `sudo dnf install securedrop-keyring` from dom0 and it provisions our public signing key in `/etc/pki/rpm-gpg/` (or elsewhere). Then the user creates the .repo file, and can run `sudo dnf install securedrop-workstation-dom0-config` and we're all set. Instead of directly trusting the SD release key, we would be trusting that the Qubes-signed package provided us with the correct key (certainly if Qubes is compromised, the whole workstation is compromised anyways).

This would largely require the Qubes team being willing to ship our key, figuring out details around update cadence, timelines, etc.

rocodes:
@legoktm I think this is a good idea whether or not we end up working towards an ISO, although I think it's currently semi blocked on the last sentence you mentioned (bit more of an internal conversation about update cadence/making sure our upstream ask isn't too onerous).

What I'm wondering if we do this is why we wouldn't also ship the .repo file and have it be autoconfigured in the postinnst. I think we can find some way of visually confirming to the user before they proceed with the installation, and/or we can still support the legacy method of checking and importing the key manually for users who want to go the extra mile.

rocodes:
Bumping this for a couple reasons, besides the initial benefits discussed above:

- https://github.com/freedomofpress/securedrop-workstation/issues/953 (/earlier https://github.com/freedomofpress/securedrop-workstation/issues/423) would be well-served by a package postinst. We should resolve this before our next key expiry this summer
- rather than configuring .repo files dynamically with Salt, it would be nice to be able to define a static .repo file(s). It would also lend more confidence about system state this way (eg: verifying that `securedrop-yum-keyring` and not `securedrop-yum-keyring-test` package is installed, vs manually searching for all the places the key could be)
[...]

PR #1210, opened 2024-12-03
WIP: Configure dom0 rpm repo settings via boostrap rpm package.

rocodes:
## Status

Work in progress, pushing in draft mode to discuss approach 
Blocked on release of securedrop-workstation-keyring boostrap rpm (to fpf repos at minimum)

## Description of Changes

Refs #945
Description TK
rpm version check compat todo (for details on rpm constraints, see https://rpm-software-management.github.io/rpm/manual/boolean_dependencies.html )

## Testing

## Deployment

2. New installs

### DIFF
--- a/securedrop_salt/sd-dom0-files.sls
+++ b/securedrop_salt/sd-dom0-files.sls
@@ -6,47 +6,6 @@
 # over time. These scripts should be ported to an RPM package.
 ##
 
-# Imports "sdvars" for environment config
-{% from 'securedrop_salt/sd-default-config.sls' import sdvars with context %}
-
-dom0-rpm-test-key:
-  file.managed:
-    # We write the pubkey to the repos config location, because the repos
-    # config location is automatically sent to dom0's UpdateVM. Otherwise,
-    # we must place the GPG key inside the fedora TemplateVM, then
-    # restart sys-firewall.
-    - name: /etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation
-    - source: "salt://securedrop_salt/{{ sdvars.signing_key_filename }}"
-    - user: root
-    - group: root
-    - mode: 644
-
-dom0-rpm-test-key-import:
-  cmd.run:
-    - name: sudo rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation
-    - require:
-      - file: dom0-rpm-test-key
-
-dom0-workstation-rpm-repo:
-  # We use file.managed rather than pkgrepo.managed, because Qubes dom0
-  # settings write new repos to /etc/yum.real.repos.d/, but only /etc/yum.repos.d/
-  # is copied to the UpdateVM for fetching dom0 packages.
-  file.managed:
-    - name: /etc/yum.repos.d/securedrop-workstation-dom0.repo
-    - user: root
-    - group: root
-    - mode: 644
-    - contents: |
-        [securedrop-workstation-dom0]
-        gpgcheck=1
-        skip_if_unavailable=False
-        gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation
-        enabled=1
-        baseurl={{ sdvars.dom0_yum_repo_url }}
-        name=SecureDrop Workstation Qubes dom0 repo
-    - require:
-      - file: dom0-rpm-test-key
-
 # Ensure debian-12-minimal is present for use as base template
 dom0-install-debian-minimal-template:
   cmd.run:
--- a/rpm-build/SPECS/securedrop-workstation-dom0-config.spec
+++ b/rpm-build/SPECS/securedrop-workstation-dom0-config.spec
@@ -41,6 +41,12 @@ BuildRequires:	systemd-rpm-macros
 Requires:		qubes-mgmt-salt-dom0-virtual-machines
 Requires:		python3-qt5
 
+# Require keyring boostrap package.
+# To ensure that production systems can only be provisioned with prod repos,
+# require any non-prod repo configuration to depend on a dev-only virtual package that
+# is built locally
+Requires:       (securedrop-workstation-keyring or ((securedrop-workstation-keyring-staging or securedrop-workstation-keyring-dev or securedrop-workstation-keyring-qa) and sdw-dev))
+
 %description
 This package contains VM configuration files for the Qubes-based
 SecureDrop Workstation project. The package should be installed
--- a/securedrop_salt/sd-default-config.sls
+++ b/securedrop_salt/sd-default-config.sls
@@ -4,6 +4,9 @@
 ##
 # Handles loading of config variables, via environment-specific
 # setting in the config file.
+# These settings apply only to apt repo components used in
+# individual VMs. dom0 rpm settings are managed by the
+# boostrap rpm package; see rpm .spec file.
 
 # Load YAML vars file
 {% load_yaml as sdvars_defaults %}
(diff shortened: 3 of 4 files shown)


Makefile
```diff
@@ -25,10 +25,23 @@ all: assert-dom0
 
 dev staging: assert-dom0 ## Configures and builds a dev or staging environment
 	./scripts/configure-environment.py --env $@
+	$(MAKE) assert-keyring-%
```
rocodes, line 28: for now, this is just a check. Once the related work (#1393) is merged, then configuration can be automated, and `make dev` or `make staging` can include the entire bootstrap and provision process.

rocodes:
(It's that the dev keyring is being installed instead of the prod one; just moved it back to in progress to fix up the test that's checking)

rocodes:
@deeplow : the existing openqa `install_dev` target installs a prod config.json:

https://github.com/deeplow/openqa-tests-qubesos/blob/8ee37b89929459f0099d9f48461d6468ea0c53d5/tests/securedrop/install_workstation.pm#L103-L105

I'm going to adjust this in a branch and I'll file a ticket here to track.

securedrop_salt/sd-dom0-files.sls
```diff
@@ -86,17 +45,6 @@ dom0-securedrop-launcher-desktop-shortcut:
     - mode: 755
 
 {% import_json "securedrop_salt/config.json" as d %}
-{% if d.environment != "dev" %}
-# In the dev environment, we've already installed the rpm from
-# local sources, so don't also pull in from the yum-test repo.
-dom0-install-securedrop-workstation-dom0-config:
```
rocodes, line 92: This is one change that isn't a strict keyring refactor, that I've called out in the issue body.  I'm proposing we deprecate reinstalling the rpm here:
- we no longer support "rpmless" install (ie just copying and running salt file) even for dev machines
- the user just installed the rpm they want (if dev/staging), or if the user is installing a prod rpm, they will run the updater and get the newest version after that
- installing a new rpm can lead to QA bugs/confusion (ie "I edited /etc/yum.repos.d to point to yum-qa, why does it now point to yum again?")

tests/test_dom0_rpm_repo.py
```diff
@@ -1,55 +1,87 @@
 import json
+import subprocess
 import unittest
+from string import Template
 
-DEBIAN_VERSION = "bookworm"
 FEDORA_VERSION = "f37"
```
rocodes, line 6: fyi: the next PR adds onto these tests, see https://github.com/freedomofpress/securedrop-workstation/pull/1392/files#diff-40301369f612ab6d4d2ac9bc02d6c72aa1f8b816198d25349332f1a246fa4d2d

Makefile
```diff
@@ -25,10 +25,23 @@ all: assert-dom0
...
 	sdw-admin --apply
 
+.PHONY: assert-keyring-%
+assert-keyring-%: ## Correct keyring pkg installed
+	@rpm -q securedrop-workstation-keyring-$* >/dev/null 2>&1 || { \
+		echo "Error: Install securedrop-workstation-keyring-$*" >&2; exit 1; \
+	}
+	@if [ "$*" = "staging" ]; then \
+		if rpm -q securedrop-workstation-keyring-dev >/dev/null 2>&1; then \
+			echo "Error: Uninstall securedrop-workstation-keyring-dev" >&2; \
```
legoktm: nit: some more clear wording

```suggestion
			echo "Error: Uninstall securedrop-workstation-keyring-dev to use staging" >&2; \
```

tests/test_dom0_rpm_repo.py
```diff
@@ -1,55 +1,87 @@
...
-                self.yum_repo_url = self.yum_repo_url_test
+            # Fall back to checking environment via keyring package
+            # (will eventually replace config.json environment)
+            if not self.env:
+                self.env = self._get_env_by_package()
 
-    def test_rpm_repo_public_key(self):
-        with open(self.pubkey_actual) as f:
-            pubkey_actual_contents = f.readlines()
+        with open("tests/files/repo_config.json") as repo_config:
```
legoktm: Is there an advantage to having this be a separate file versus just a Python dict?

tests/test_dom0_rpm_repo.py
```diff
@@ -1,55 +1,87 @@
...
+                ["rpm", "-q", pkg], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL
+            )
+
+            return True
+        except subprocess.CalledProcessError:
+            return False
+
+    def test_dom0_has_keyring_package(self):
+        # Prod keyring is always installed
+        assert self._is_installed("securedrop-workstation-keyring")
```
legoktm, line 94: This is understandably (I think) failing in OpenQA since we don't have the prod keyring set up yet...

Should we publish it to yum-test as well? (We already have the client's securedrop-keyring published in apt-test)

rocodes: Yeah, I think if we're doing that for apt, it's ok to do it here too.

tests/test_dom0_rpm_repo.py
rocodes: It could be in a Python dict too; I was trying to make clearly findable/readable/static (ie the only change is in 2027 or if the key fingerprints change).

tests/test_dom0_rpm_repo.py
rocodes: https://github.com/freedomofpress/securedrop-yum-test/pull/72

Makefile
rocodes: addressed in [892ec66](https://github.com/freedomofpress/securedrop-workstation/pull/1210/commits/892ec663faf636b1778bf4d5b6e579bcd6ee31cf)

rocodes:
Added a commit to address review feedback, but didn't re-trigger openqa; I'll wait til the package is on yum-test then amend commits + force-push so just the last one has the openqa trigger.

tests/test_dom0_rpm_repo.py
legoktm: I would personally prefer a Python dict, just so file reading is one less thing that can go wrong or cause side effects in tests but I won't block over it.

tests/test_dom0_rpm_repo.py
legoktm: @rocodes Merged, though now that I think of it, it's not clear to me what will force the installation of the prod keyring until https://github.com/freedomofpress/securedrop-workstation/pull/1409. Am I missing something? Should we pull that change into this PR, or move this test into that one?

rocodes: ~In openqa, I can go back to setting the environment to "prod" in config.json, and that should pull in the prod keyring. It's a bit weird, but I think allows us to test all the changes separately.~
Edit: this is true, but I just cherrypicked the commit in from #1409 and that way we can close that PR when this passes openqa.

deeplow: Not sure if it helps, but we can pass a variable specifying the target test environment and then in OpenQA we fork behavior based on that.

rocodes: no need, we published the package to yum-test nightlies: https://github.com/freedomofpress/securedrop-yum-test/commit/b70cd4282f7b75d33900499343d0677250b1ec31 

I think in general we should try to use one branch on our openqa fork for consistency, and just submit an upstream PR when we're happy with the changes

rocodes:
Pulling in the dependency on the prod keyring early made things 5% more complicated, I'm putting this briefly back in progress.

rocodes:
We introduced a problem when we introduced the DIST substitution in the keyring. I've submitted https://github.com/freedomofpress/securedrop-workstation-keyring/pull/30 to correct it; once it's merged I'll rebuild the keyring package for yum-test and yum-test nightlies, then we should be on track.

PR #1213, opened 2024-12-05
Update github actions to use ubuntu 24

enpaul:


legoktm:
Closing since all the GHA jobs passed; only the SDW CI runner failed, which doesn't use GHA.

PR #1216, opened 2024-12-09
Do not mark sd-app/sd-gpg as internal

### LINKED ISSUE #1214: sd-app as internal vm hides it from backup targets

## Description

Setting sd-app to be an internal VM hides it from the list of potential backup targets in the Qubes backup tool. 

## Steps to Reproduce

Prod config + try to back up sd-app.

## Expected Behavior

sd-app visible in backup targets via Qubes backup tool. This is consistent with our backup / restore docs.

## Actual Behavior

sd-app is missing

## Comments

sd-app internal pref should be removed for usability

rocodes:
## Status

Ready for review

## Description of Changes

Fixes #1214 - internal VMs are not visible as backup targets, so revert setting this flag for sd-app and sd-gpg. For the time being, I think it's still reasonable to mark the proxy and viewer VMs as internal, but it's also worth revisiting. Our rationale was that the Qubes Applications menu was overwhelming for users and we wanted to avoid unnecessary menu entries (or possibilities to, for example, launch the client without running the updater, or make configuration changes), but I think easy backup options takes priority. 

## Testing

## Deployment

No considerations: when this package is released, existing instances will be automatically upgraded (dom0 salt states).


legoktm (approved):
I think this is a reasonable temporary fix because we want these VMs to be backed up, but I think long-term we especially want sd-gpg and sd-app to be considered internal as those are the most sensitive VMs?

legoktm:
Oops - I will put in a fix!

legoktm:
Please see #1219 for the follow-up.

PR #1219, opened 2024-12-11
Always set a VM's internalness

legoktm:
## Status

Ready for review

## Description of Changes

To ensure we correctly flip the status of a VM's internal feature or not, ensure the property is always being set, but conditional the value on the environment.

And then explicitly set sd-app and sd-gpg as non-internal, so that salt will update them from when they used to be internal (see e754b40e487).

Refs #1214

## Testing

* [ ] provision a prod workstation using 1.0.0 or anything before e754b40e487; sd-app, sd-gpg, sd-proxy are internal  
* [ ] re-provision that prod workstation using this PR; sd-app and sd-gpg should be non-internal, while sd-proxy stays internal.
* [ ] re-provision as a dev workstation using this PR; sd-proxy should switch to being non-internal

## Deployment

Any special considerations for deployment? This is to implement handling for upgrades.

## Checklist


legoktm:
I don't have my normal SDW testing machine with me so I haven't been able to test this at all.

securedrop_salt/sd-proxy.sls
```diff
@@ -25,11 +25,9 @@ sd-proxy-dvm:
       - netvm: sd-whonix
       - template_for_dispvms: True
       - default_dispvm: ""
-  {% if d.environment == "prod" %}
     - features:
       - set:
-        - internal: 1
-  {% endif %}
+        - internal: {{ int(d.environment == "prod") }}
```
legoktm, line 30: salt/jinja isn't happy with this syntax so I'm just going to change it to a normal if/else clause.

Issue #1225, opened 2025-01-02
Discussion: can we deprecate or simplify `securedrop-handle-upgrade`? 


rocodes:
* [x] I have searched for duplicates or related issues

## Description

We have a script, `securedrop-handle-upgrade`, that is invoked via Salt in `sd-upgrade-templates.sls`. Its main purpose is to make sure that base templates for vms (debian, whonix) are correctly up to date for a given VM. (Its secondary purpose, via `remove`, is to clean up old sd templates after an upgrade).

As the script mentions, to update an appvm's template, the vm must be a) powered off, and b) not be appvm template for a disposable VM that is either the default disposable or the system disposable.
  
The downsides of this approach are:
- it's convoluted (Salt requsite -> salt state that invokes a script with a defined arg -> script shells out to "qvm" commands, but we're already in the middle of a salt run...)
- it is applied both in the dom0 top file, but also (inconsistently) as a requisite in some but not all appvms (eg: it's a requisite in sd-gpg.sls but not sd-app.sls or sd-viewer.sls). Note that if it were applied to sd-viewer _after_ the sd-viewer vm were created and set as the system default dispVM and there was a template upgrade needed, it would break, since it would be trying to remove a vm that is a system default disposable.

It might be better to move this out of a script and into the individual vm configuration. Afaict we can use `qvm.vm`, which I don't think existed when this script was created, to ensure that the VM's state is "powered off" (goal 1), and use "prefs" for the VM to ensure that it has the correct template (goal 2).

The original goal of the script was that if a user missed a dom0 update where the template migration took place, they could still successfully upgrade (particularly during template consolidation). But I think we can deprecate this, at least at the individual vm level.  Would like a second opinion though (@deeplow @legoktm maybe ?)

## How will this impact [SecureDrop/SecureDrop Workstation users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?
Should be no-op; potentially faster provisioning

## How would this affect the SecureDrop Workstation [threat model](https://github.com/freedomofpress/securedrop-workstation/#threat-model)?
Should be no-op

## User Stories
(n/a - developer quality of life/ease of maintenance)

legoktm:
It's been a while since I looked at that script, I have no attachment to it and would go on my normal spiel about how that complex logic doesn't belong in bash scripts, etc. :), so I'm down to move it to salt.

deeplow:
Look at this now in the context of upgrading to debian 13 (#1719)

[...]

Issue #1060, opened 2024-06-04
refactor `sd-whonix` provisioning


cfm:
We've gone back and forth on our whonix provisioning approach in sdw: we previously used `qvm.anon-whonix`, which are upstream tested Salt states that configure the latest whonix vms and set up an up to date disposable anon-whonix vm. Now we invoke the template name directly, needing a little more attention when upgrading, to make sure we upgrade everywhere consistently. (It also doesn't change the user's Tor VMs, which is good or bad depending on how heavy-handed we think we should be).

I kind of think we should:
* run the upstream `qvm.anon-whonix` state again as part of our use of whonix (they describe it as a mandatory step - it configures [update policy information](https://www.whonix.org/wiki/Dev/Qubes#salt), in addition to setting up vms).
* Then clone the whonix gateway and name it whatever we want, and apply the custom configuration to that.

(I think what I'm suggesting can/should be a separate PR)

_Originally posted by @rocodes in https://github.com/freedomofpress/securedrop-workstation/pull/1051#discussion_r1624796502_

rocodes:
I plan to put up a draft PR to include the builtin `qvm.anon-whonix` as a requisite to our whonix config, which we historically included [until the transition to Qubes 4.1](https://github.com/freedomofpress/securedrop-workstation/pull/751/files#diff-6ca804bebe065769f7bc34a932c9bd73f48a7ee02468f7168d164139f9f05a8aL9-L14). Using a more standardized whonix setup may help with the failures we are seeing in https://github.com/freedomofpress/securedrop-workstation-ci/issues/68. 

That can happen independently of whether we decide to clone the whonix-gateway template for our own customization or not.

PR #1226, opened 2025-01-02
Set vm template consistently across Salt files; cleanup unused requisites

rocodes:
## Status

Ready for review

## Description of Changes

Refactor-only, this should be effectively a no-op.

- Set template consistently across all vm salt files. Once a VM has been created, `prefs` _must_ be used to update the template's attributes (`present` checks only the vm name and has no other effect).  However, after looking into it a bit I think it's still a good idea to set the preference in `present`, since it's not clear if the vm is briefly created with the default appvm template otherwise.
- Remove unused `include` of `securedrop_salt.sd-upgrade-templates` in VMs where the requisite has been removed. This was in place [from before template consolidation](https://github.com/freedomofpress/securedrop-workstation/pull/619/files#diff-3fc9c3822ab07f6bbf0b11a85d1ff73c27768b70dfa0e0601fcfc300ed615b81), when these files contained not just the VM setup but also the per-VM templates, where `sd-upgrade-templates` was required. 
  - Note: there are still a few files where the requisite _is_ used (sd-gpg, sd-whonix). Since this PR is meant to be a no-op, I have not made changes to those files. I think that should be addressed separately, via #1225 and/or also via https://github.com/freedomofpress/securedrop-workstation/issues/1060 in the case of sd-whonix, since ([per comment](https://github.com/freedomofpress/securedrop-workstation-ci/issues/68#issuecomment-2568141007)) we may want to use the whonix-recommended update strategy, which is not just `apt update`.  
- Include `sd-viewer.sls` as an explicit requisite of sd-app, since sd-app sets sd-viewer as its dispvm 
- Correct whonix version number in test docs (whonix 16 -> whonix 17)
 
## Testing

- [x] Visual review
- [ ] CI

## Deployment

2. New installs

### DIFF
--- a/securedrop_salt/sd-app.sls
+++ b/securedrop_salt/sd-app.sls
@@ -14,13 +14,18 @@
 
 include:
   - securedrop_salt.sd-workstation-template
-  - securedrop_salt.sd-upgrade-templates
+  - securedrop_salt.sd-viewer
 
 sd-app:
   qvm.vm:
     - name: sd-app
     - present:
+      # Sets attributes if creating VM for the first time,
+      # otherwise `prefs` must be used.
+      # Label color is set during initial configuration but
+      # not enforced on every Salt run, in case of user customization.
       - label: yellow
+      - template: sd-small-{{ sdvars.distribution }}-template
     - prefs:
       - template: sd-small-{{ sdvars.distribution }}-template
       - netvm: ""
@@ -38,6 +43,7 @@ sd-app:
         - service.securedrop-mime-handling
     - require:
       - qvm: sd-small-{{ sdvars.distribution }}-template
+      - sls: securedrop_salt.sd-viewer
 
 sd-app-config:
   qvm.features:
--- a/securedrop_salt/sd-devices.sls
+++ b/securedrop_salt/sd-devices.sls
@@ -11,14 +11,17 @@
 
 include:
   - securedrop_salt.sd-workstation-template
-  - securedrop_salt.sd-upgrade-templates
 
 sd-devices-dvm:
   qvm.vm:
     - name: sd-devices-dvm
     - present:
-      - template: sd-large-{{ sdvars.distribution }}-template
+      # Sets attributes if creating VM for the first time,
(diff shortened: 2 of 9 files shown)


securedrop_salt/sd-app.sls
```diff
@@ -14,13 +14,16 @@
...
-  - securedrop_salt.sd-upgrade-templates
+  - securedrop_salt.sd-viewer
 
 sd-app:
   qvm.vm:
     - name: sd-app
     - present:
+      # Sets attributes if creating VM for the first time,
+      # otherwise `prefs` must be used.
       - label: yellow
```
legoktm, line 25: Should this also be duplicated under `prefs`? Or did you not because we assume it's static (which is reasonable IMO)

legoktm (commented):
Should we have a test that verifies that all of the qvm.vm stanzas have a `template` entry under both present and prefs? And that they're the same?

securedrop_salt/sd-app.sls
rocodes: The colour label you mean? yeah I guess technically someone could change the label colour. (If they do, do we _want_ to forcibly change it back?)

legoktm: Yeah, I was just curious why it was the only key that wasn't duplicated. I think it's a reasonable thing to only set once and let the user override if they want.

But let's explicitly document it, and maybe a test can assert the color is only set in present and not under prefs?

rocodes: Have added explanatory comments (squashed) in https://github.com/freedomofpress/securedrop-workstation/pull/1226/commits/b70de6c7ca9ce1dfeba081f9ddd794bf0819b9c2

rocodes:
(added comments, rebased on main/squashed a couple commits)

rocodes:
I saw your suggestion, and I'd like to keep it in `present` for now even though it's a bit redundant, so that the vm consistently uses our configured templates at all times.

My preference would be to address https://github.com/freedomofpress/securedrop-workstation/issues/1236 in a followup PR ?

Issue #998, opened 2024-04-24
Use pytest for dom0 tests


legoktm:
* [x] I have searched for duplicates or related issues

## Description

People have already wanted pytest for the `make test` dom0 tests, but noted that it's not available in dom0

```
    def test_sd_proxy_template(self):
        """
        Asserts that the 'sd-proxy' VM is using a supported base OS.
        """
        # This test is a single example of the method for testing: it would
        # be ideal to use a loop construct (such as pytest.mark.parametrize),
        # but doing so would introduce additional dependencies to dom0.
        vm = self.app.domains["sd-proxy"]
        self._validate_vm_platform(vm)

    def test_all_sd_vm_platforms(self):
        """
        Test all VM platforms iteratively.

        Due to for-loop implementation, the first failure will stop the test.
        Therefore, even if multiple VMs are NOT running a supported platform,
        only a single failure will be reported.
        """
        # Would prefer to use a feature like pytest.mark.parametrize
        # for better error output here, but not available in dom0.
        for vm_name in WANTED_VMS:
            vm = self.app.domains[vm_name]
            self._validate_vm_platform(vm)
```

Fedora 37 has pytest 7.1.3, we could `sudo qubes-dom0-update python3-pytest` to get it into dom0. This would be similar to how running the testinfra tests on an admin workstation require you to install extra testing dependencies.

legoktm:
A few things I've stumbled through along the way:
* In qubes-dom0-update (because of DNF I think?), `-y` is not short for `--yes`, it's just `-y`.
* `python -m pytest` and `pytest` are different, the former is what we need for our `from base import ...` to work, until we can move that to fixtures.
* Tests that read config.json except to be run from the root instead of tests/.

Initially I was thinking we switch to pytest as the runner so future tests can use pytest, but it might just be easier to port all the tests over and get rid of `import base`

legoktm:
(not actively working on this right now)

PR #1234, opened 2025-01-16
(WIP) Add dom0 systemd tests

rocodes:
## Status

Work in progress

## Description of Changes

Now that we have more things being configured via systemd, we should test in dom0 that the behaviour is as we want.

These tests sketch out how we can
- check that we have the number of services we think we do*
- and that they are well-formed (via systemd-analyze)
- and that they are present and enabled
- and that their last exit status was successful

These tests will not catch systemd service/timer files that are not named with `securedrop` somewhere in the name, which has so far been our pattern but is worth noting.

They also assume that we _want_ services enabled (they don't require `preset` enabled though), so if our logic gets more complex down the line, we may need to change that. 

## Testing

## Deployment

2. New installs


tests/test_dom0_systemd.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env python3
+import subprocess
+import unittest
+
+# .spec file shows which units are shipped with rpm
+SYSTEM_UNITS = ["securedrop-logind-override-disable.service"]
+
+USER_UNITS = [
```
legoktm, line 8: It would be nice if we could just glob to find these so we don't need to remember to manually add new services/timers to this list.

tests/test_dom0_systemd.py
```diff
@@ -0,0 +1,106 @@
...
+    "securedrop-user-xfce-icon-size.service",
+    "securedrop-user-xfce-settings.service",
+    "sdw-notify.timer",
+]
+
+# list all system unit files (append "--user" for user files)
+# that match our naming convention
+SYSTEMD_LIST_UNITS = ["systemctl", "--legend=false", "list-unit-files", "--all", "*securedrop*"]
```
legoktm, line 16: This approach is great too, I wonder if we should be renaming the `sdw-notify.timer` unit as out of compliance with this then.

tests/test_dom0_systemd.py
```diff
@@ -0,0 +1,106 @@
...
+
+# from systemctl status format
+SYSTEMD_EXIT_SUCCESS = "status=0/SUCCESS"
+
+
+class Dom0_Systemd(unittest.TestCase):
```
legoktm, line 22: I would personally suggest not wrapping this into a class unless there's a strong reason to do so, and just do pytest style `test_*` functions with helpers as normal functions. To make it easier whenever we get around to https://github.com/freedomofpress/securedrop-workstation/issues/998

legoktm (commented):
This is really great, thanks for starting this. We could probably use most of this as-is in the server too. Few suggestions inline.

tests/test_dom0_systemd.py
rocodes: Yeah, I wondered about that, but also I wonder if there is a benefit to being explicit about an exact match of the right number of services.  Fancy us would be able to grab from the spec file and/or the presets files (or, if we had a better directory structure in the workstation repo, we could have `files` have a subdirectory juts for systemd files...), and test that against what's on dom0, instead of globbing twice on dom0. Will think about it

eloquence:
@rocodes Also possibly a next sprint candidate?

PR #1237, opened 2025-01-16
Explicitly unset internalness of sd-app and sd-proxy

deeplow:
With #1219 sd-app and sd-proxy were explicitly as "internal: 0". However, this does not work, as the developer docs warn [1]:

> Some extensions interpret the values as boolean. In this case,
> the empty string means False and non-empty string (commonly '1')
> means True.

To fix this, the internalness of a qube is explicitly set to "".

[1]: https://dev.qubes-os.org/projects/core-admin-client/en/latest/manpages/qvm-features.html#description

## Status

Ready for review

## Description of Changes

Refs #1219

## Testing

- deploy version from main (may be developer environment)
- restart the Applications Menu: `pkill qubes-app-menu && qubes-app-menu)
- ensure `sd-app` and `sd-gpg` is not present (this is the bug)
- deploy this branch via `make dev`
- restart the Applications Menu: `pkill qubes-app-menu && qubes-app-menu)

## Deployment

2. New installs

### DIFF
--- a/securedrop_salt/sd-app.sls
+++ b/securedrop_salt/sd-app.sls
@@ -32,7 +32,7 @@ sd-app:
     - features:
       - set:
         - vm-config.SD_MIME_HANDLING: sd-app
-        - internal: 0
+        - internal: ""
       - enable:
         - service.paxctld
         - service.securedrop-mime-handling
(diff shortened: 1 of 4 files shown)


rocodes (approved):
Ah, thanks for the catch- lgtm

PR #1253, opened 2025-02-20
Configure sd-log at the same time as other VMs

### LINKED ISSUE #1247: Sometimes sd-log shutdown error makes provisioning fail
Noticed eg in https://ws-ci-runner.securedrop.org/2025-02-13-222929262445-46d18ffdae2edf06935352b67dc690db2a2625d8-Qubes_4.2_D-update_20250213043413.log.txt although I've seen it on and off:  `qvm-shutdown --wait` on sd-log doesn't always succeed, so sometimes (CI) provisioining fails (but presumably would also be true of real world systems).

Could be because logs are being streamed in via qrexec - could be that we need to force-shut down instead of just shutting down, or that there's a more graceful way to handle all of it (a pre-shutdown signal to stop the qrexec logging service and allow clean shutdown, if that is at all the issue?).

Merging https://github.com/freedomofpress/securedrop-workstation/pull/1246 (where this most recently occurred) and tracking this as a separate issue.

deeplow:
Historically sd-log (and its template) was being configured earlier than the remaining qubes so that logs would be captured early in the install process. However, is some cases restarting sd-log hangs due the already-configured templates trying to send logs to it, which starts sd-log, while it is getting shut down (#1247).

The current solution to avoid this race-condition and adhering more to the salt convention of not trying to force order outside of salt, sd-log will no longer be setup early.

After a full system restart sd-log should be capturing logs as it normally would.

## Status

Ready for review

## Description of Changes

Should fix #1247

## Testing

- [ ] Visual review
- [ ] CI (clean install) completes without error 
- [ ] Upgrade (manual testing): upgrade on existing SDW install completes without error 

## Deployment

2. New installs

### DIFF
--- a/files/sdw-admin.py
+++ b/files/sdw-admin.py
@@ -91,11 +91,6 @@ def provision_and_configure():
     provision("Provisioning base template", "securedrop_salt.sd-base-template")
     configure("Configuring base template", ["sd-base-bookworm-template"])
     provision_all()
-    configure(
-        "Configuring template for log sink before anything else",
-        ["sd-small-bookworm-template"],
-        restart=["sd-log"],
-    )
     configure("Enabling Whonix customizations", ["whonix-gateway-17"])
     configure(
         "Configure all SecureDrop Workstation VMs with service-specific configs",


rocodes (approved):
- [x] Visual review
- [x] CI (clean install) completes without error 
- [x] Upgrade (manual testing): upgrade on existing SDW install without error 

We have discussed that this means that sd-log won't contain logs from inside of vms during the installation/provisioning process or til the sd-log VM is rebooted. However, that's reasonable given that we don't make use of those logs til vms reboot, and we instruct users to reboot after provisioning. 

Will add to QA plan and docs that testers must also reboot (at minimum VMs, but really, the whole system) after provisioning, in order to reduce manual testing error. 

LGTM

PR #1260, opened 2025-02-27
fix(`update-fedora-template-if-new`): run as GUI user

cfm:
## Status

Ready for review (but drafted without testing)

## Description of Changes

Fixes the case where a new template's update log in `/var/log/qubes/update-${TEMPLATE}.log` is owned by (Salt running as) `root:qubes`, breaking subsequent updates that expect to be able to write to it as `user:qubes`.

## Testing

```sh-session
$ sudo qubesctl state.sls securedrop_salt.sd-sys-vms
$ qubes-vm-update --apply-to-all --force-update --show-output --just-print-progress --targets fedora-41-xfce
$ ls -l /var/log/qubes/update-fedora-41-xfce.log
-rw-r--r-- 1 user qubes 44456 Feb 27 10:37 /var/log/qubes/update-fedora-41-xfce.log
             ^^^^^^^^^^
```

- [ ] Succeeds with `gui_user == "user"`
- [ ] Succeeds with `gui_user != "user"`

## Deployment

- Clean installation on Qubes 4.2.4 (`fedora-41-xfce` already present)
- Upgrade: `fedora-41-xfce` already installed via `qvm-template`, and then `securedrop-workstation-dom0-config` 1.1.1 is installed
- Upgrade: `securedrop-workstation-dom0-config` 1.0.2→1.1.1 installs `fedora-41-xfce`

### DIFF
--- a/securedrop_salt/sd-sys-vms.sls
+++ b/securedrop_salt/sd-sys-vms.sls
@@ -14,6 +14,8 @@ include:
 {% set sd_supported_fedora_version = 'fedora-41' %}
 {% set sd_fedora_base_template = sd_supported_fedora_version + '-xfce' %}
 
+{% set gui_user = salt['cmd.shell']('groupmems -l -g qubes') %}
+
 # Install latest templates required for SDW VMs.
 dom0-install-fedora-template:
   cmd.run:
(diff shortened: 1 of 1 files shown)


rocodes (approved):
- [x] Visual review
- [x] Test plan against a non-default username; log file permissions are set correctly

LGTM.

cfm:
Why have we not added a regression test for this fix?  @rocodes, @deeplow, and I discussed this and concluded that we can't write a meaningful pytest case for post-update states, only for post-provisioning states.  This is rather a candidate for OpenQA: that an updater run is successful; or, better, yet, that two consecutive updater runs are successful.

PR #1295, opened 2025-04-29
Remove remaining hardcoded references to Whonix version

deeplow:
## Status

Ready for review

## Description of Changes

Removes hardcoded whonix version from remaining code base.

Refs #1060 (partial fix -- still missing Trixie debs being available for SecureDrop packages)

### Simulate a Whonix 18 upgrade
Apply the following patches:
`/srv/salt/securedrop_salt/sd-sys-whonix-vms.sls`:

   ```diff
   whonix-{{ component }}-{{ whonix.whonix_version }}-installed:
   +  qvm.vm:
   +    - name: whonix-{{component}}-{{ whonix.whonix_version }}
   +    - present:
   +      - label: black
   -  qvm.template_installed:
   -    - name: whonix-{{ component }}-{{ whonix.whonix_version }}
   -    - fromrepo: {{ whonix.whonix_repo }}
   -    - require:
   -      - sls: qvm.template-whonix-{{ component }}
   ```
`/srv/formulas/base/virtual-machines-formula/qvm/template-whonix-workstation.sls`
   ```diff
   template-whonix-gateway-{{ whonix.whonix_version }}:
   -  qvm.template_installed:
   -    - name:     whonix-gateway-{{ whonix.whonix_version_override_template }}
   -    - fromrepo: {{ whonix.whonix_repo }}
   +  qvm.vm:
   +    - name: whonix-gateway-{{ whonix.whonix_version }}
   +    - present:
   +      - label: black
   ```

### Actual test

- [ ] `make test` passes in dom0 (except for `test_sd_whonix_config`)
- [ ] run updater and confirm in logs that whonix-gateway-18 was updated

### Undo changes

- [ ] Don't forget to undo the changes in `/srv/formulas`. The easiest way is to `sudo qubes-dom0-update --action=reinstall qubes-mgmt-salt-dom0-virtual-machines`.

## Deployment

2. New installs

### DIFF
--- a/sdw_util/Util.py
+++ b/sdw_util/Util.py
@@ -174,10 +174,8 @@ def get_whonix_version():
             universal_newlines=True,
         )
         whonix_version = int(output.strip())
-    except subprocess.CalledProcessError:
-        return None
-    except (AttributeError, ValueError):
-        return None
+    except (AttributeError, ValueError, subprocess.CalledProcessError):
+        raise RuntimeError("Whonix version could not be obtained")
 
     return whonix_version
 
--- a/sdw_util/Util.py
+++ b/sdw_util/Util.py
@@ -158,6 +158,8 @@ def get_whonix_version():
         {{ whonix.whonix_version }}
 
     This function extracts this value so that it can be on this script.
+
+    NOTE: function takes a few seconds to run
     """
 
     try:
--- a/sdw_util/Util.py
+++ b/sdw_util/Util.py
@@ -148,6 +148,40 @@ def get_qubes_version():
     return version
 
 
+def get_whonix_version():
+    """
+    Obtain Whonix version from /srv/formulas/base/virtual-machine-formula/qvm/whonix.jinja
+
+    In '.sls' files the whonix version is obtained with
+
+        {% import "qvm/whonix.jinja" as whonix %}
+        {{ whonix.whonix_version }}
+
+    This function extracts this value so that it can be on this script.
+    """
+
+    try:
+        output = subprocess.check_output(
+            [
+                "sudo",
+                "qubesctl",
+                "jinja.load_map",
+                "qvm/whonix.jinja",
+                "whonix_version",
+                "--out",
+                "newline_values_only",
+            ],  # avoid "local:" prefix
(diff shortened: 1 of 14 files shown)


sdw_util/Util.py
```diff
@@ -148,6 +150,45 @@ def get_qubes_version():
...
+    In '.sls' files the whonix version is obtained with
+
+        {% import "qvm/whonix.jinja" as whonix %}
+        {{ whonix.whonix_version }}
+
+    This function extracts this value so that it can be on this script.
+    """
+
+    # "caching" because qubesctl takes several seconds
+    global whonix_version  # noqa: PLW0603
```
legoktm, line 166: If we need caching, I would suggest using [`functools.cache`](https://docs.python.org/3/library/functools.html#functools.cache), instead of a global variable.

sdw_util/Util.py
```diff
@@ -148,6 +150,45 @@ def get_qubes_version():
...
+                "jinja.load_map",
+                "qvm/whonix.jinja",
+                "whonix_version",
+                "--out",
+                "newline_values_only",
+            ],  # avoid "local:" prefix
+            universal_newlines=True,
+        )
+        whonix_version = int(output.strip())
+    except subprocess.CalledProcessError:
```
legoktm, line 184: Is this the right behavior? If we can't get the whonix version, shouldn't we hard fail? (None of the callers have any error handling to check for a `None` value, we're just going to end up with operations on `whonix-gateway-None`.

legoktm (changes requested):
Just looked at the Python so far, will test it now.

legoktm:
> Some back and forth to fix CI. But this is annoying

You might need to run `poetry install --sync` to pick up the latest version of ruff.

> Something is up with the CI. It's been running for too long.

If it's stuck in the "queued" state, that means it crashed before the job even started running (which does happen sometimes). The solution is to amend and repush it. But...

legoktm (approved):
Let's do it.

legoktm:
Let's have a follow-up PR for that after we get rc1 out? At least it's a clear enough failure that we can verify the intent of the test worked.

PR #1329, opened 2025-05-28
Initial pytest dom0 adaptation

### LINKED ISSUE #1337: Initial pytest support for dom0 tests
See parent issue for rationale (https://github.com/freedomofpress/securedrop-workstation/issues/1320). Minimal changes necessary to fully run pytest tests in dom0, migrated from unittest, while taking advantage of the pytest-unittest backwards compatibility.

deeplow:
## Status

Ready for review

## Description of Changes

Fixes #1337. We need pytest in dom0 in order to generate JunitXML reports cleanly. Even through pytest is somewhat backwards compatible, in practice the base class tests where failing with the following issue:

```
self = <tests.base.SD_VM_Local_Test testMethod=test_enforced_apparmor_profiles>

    def setUp(self):
        self.app = Qubes()
>       self.vm = self.app.domains[self.vm_name]
E       AttributeError: 'SD_VM_Local_Test' object has no attribute 'vm_name'
````

So it includes some refactors to get this at least working. 

## Testing

- [ ] visual review
    (**NOTE:** commit by commit may be an easier review, possibly with `--ignore-space-change`)
- [ ] SDW CI passes (blocked on https://github.com/freedomofpress/securedrop-workstation-ci/pull/82)
- [ ] Number of tests run doesn't change

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,21 +135,12 @@ clean: assert-dom0 prep-dev ## Destroys all SD VMs
 # if clean has already been run.
 	./scripts/sdw-admin.py --uninstall --force
 
-PYTEST_PATH=$(shell command -v pytest)
-COVERAGE_PATH=$(shell command -v coverage) # instead of pytest-cov, check for its dep.
 .PHONY: test-prereqs
 test-prereqs: assert-dom0 ## Checks that test prerequisites are satisfied
 	@echo "Checking prerequisites before running test suite..."
-	test -e config.json || exit 1
-	test -e sd-journalist.sec || exit 1
-ifeq ($(PYTEST_PATH),)
-	@echo 'please install pytest with "sudo qubes-dom0-update python3-pytest"'
-	@false
-endif
-ifeq ($(COVERAGE_PATH),)
-	@echo 'please install pytest with "sudo qubes-dom0-update python3-pytest-cov"'
-	@false
-endif
+	test -e config.json || (echo "Ensure config.json is in this directory" && exit 1)
+	test -e sd-journalist.sec || (echo "Ensure sd-journalist.sec is in this directory" && exit 1)
+	which pytest coverage || (echo 'please install test dependencies with "sudo qubes-dom0-update python3-pytest python3-pytest-cov"' && exit 1)
 
 test: test-prereqs ## Runs all application tests (no integration tests yet)
 	pytest -v tests

### CODE REVIEW

File: tests/conftest.py [hunk 1]
```diff
@@ -0,0 +1,13 @@
...
+
+import pytest
+
+PROJ_ROOT = Path(__file__).parent.parent
+
+
+@pytest.fixture
+def dom0_config():
+    """Make the dom0 "config.json" available to tests."""
+    with open(PROJ_ROOT / "config.json") as config_file:
```
rocodes, line 12: This might be a bit brittle - do we know for sure that config.json is in that location? Also, maybe we want the config that was used for provisioning and not a repo copy (so that for users who are running manual workstation tests on their own machines, they aren't bitten by stale config file lying around)?

To support both CI and manual test scenarios, wdyt about looking in /usr/share/securedrop-workstation-dom0-config/ for the real config file that the system was provisioned against, and make that available, since these tests only work after a system provisioning?

tests/test_proxy_vm.py
```diff
@@ -1,60 +1,64 @@
...
+        enforced_apparmor_profiles={"/usr/bin/securedrop-proxy"},
+    )
+
+
+def test_do_not_open_here(qube):
+    """
+    The do-not-open here script has been removed from sd-proxy.
+    All VMs now default to using open-in-dvm.
+    """
+    assert not qube.fileExists("/usr/bin/do-not-open-here")
```
rocodes, line 23: (We can probably remove this test since do-not-open-here hasn't been present since long before 4.2 and this will always pass, but no strong feelings about removing now or later)

rocodes (commented):
Thank you @deeplow! Took a pass through, left a couple of minor comments. 

Overall lgtm, pretty straightforward without a lot of changes besides the necessaries. As followup work, we may consider making things a little more DRY, as you talked about when figuring out how to port this over to pytest - perhaps by having one single place for configuration information that feeds into both the test suite and the vm configs themselves, so that we can just write one generic test for each behaviour, and feed it a vm and an expected result. But that's a more involved effort so it's more future speculation than an ask for right now.

Thanks for preparing this!  I'll run through a local test in a bit.

tests/conftest.py
deeplow: I agree with you, but I didn't want to make "editorial decisions" and wanted to keep it as close as possible to the original, but I coudn't resist.

The [original code](https://github.com/freedomofpress/securedrop-workstation/pull/1329/files#diff-bdbf5290b28c74db41a6af29864daeabbed3fcc92051c15111aebaf5ba5ba703L64-L66) as shown bellow assumed the user was running from a directory where `config.json` was, but I found even more brittle.

```python
        # Make the dom0 "config.json" available to tests.
        with open("config.json") as config_file:
            self.dom0_config = json.load(config_file)
```

I'd honestly prefer to address these in a separate PR and keep it simple on this one, if OK with you. With that said, would you prefer I keep it close to the original or with the `PROJ_ROOT`, as I have implemented?

tests/test_proxy_vm.py
deeplow: I'd prefer leaving as-is and addressing later. I did remove some deprecated test code, but that was commented out, I think, so it's more clear-cut on why it's gone in a "minimal conversion" PR.

But I have documented this in https://github.com/freedomofpress/securedrop-workstation/issues/1339 so we can keep track of it.

tests/conftest.py
rocodes: My opinion is that this isn't editorializing, it's a very minor change and/or a bugfix to load the config from the correct file location, but I'm okay with it being addressed in a followup if you prefer, particularly if the current code is compatible with the workstation-CI runner.

Makefile
```diff
@@ -135,29 +135,39 @@ clean: assert-dom0 prep-dev ## Destroys all SD VMs
...
 .PHONY: test-prereqs
 test-prereqs: assert-dom0 ## Checks that test prerequisites are satisfied
 	@echo "Checking prerequisites before running test suite..."
 	test -e config.json || exit 1
 	test -e sd-journalist.sec || exit 1
+ifeq ($(PYTEST_PATH),)
+	@echo 'please install pytest with "sudo qubes-dom0-update python3-pytest"'
+	@false
+endif
+ifeq ($(COVERAGE_PATH),)
```
rocodes: I ran through this without pytest and pytest-cov installed in dom0 to test out these prompts. If `config.json` or `sd-journalist.sec` are not in place, the user doesn't see these prompts til they resolve that issue (sidenote: we could improve this either here or in a followup with as little as `(echo "Ensure config.json is in this directory" && exit 1)`).

Once config files are present, running the make target does correctly show the "please install pytest" prompt, but the pytest-cov prompt does not display -- with only pytest installed, after running `make test`, the user sees 

```
Checking prerequisites before running test suite...
test -e config.json || exit 1
test -e sd-journalist.sec || exit 1
pytest -v tests
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir]
#... etc
make: *** [Makefile:155: test] Error 4
```

I'd suggest showing a single message instructing users about both dependencies, and if you want to be really friendly, it can be before the `test -e ...` stuff, but that's not important.

deeplow: Let me know if 18090e5 is what you were looking for.

Makefile
```diff
@@ -138,26 +138,27 @@ clean: assert-dom0 prep-dev ## Destroys all SD VMs
 .PHONY: test-prereqs
 test-prereqs: assert-dom0 ## Checks that test prerequisites are satisfied
 	@echo "Checking prerequisites before running test suite..."
-	test -e config.json || exit 1
-	test -e sd-journalist.sec || exit 1
+	test -e config.json || (echo "Ensure config.json is in this directory" && exit 1)
+	test -e sd-journalist.sec || (echo "Ensure sd-journalist.sec is in this directory" && exit 1)
+	which pytest coverage || (echo 'please install test dependencies with "sudo python3-pytest python3-pytest-cov"' && exit 1)
```
rocodes: There's a slight mistake in the text here: `sudo python3-pytest python3-pytest-cov` should be `sudo qubes-dom0-update python3-pytest python3-pytest-cov`

rocodes (changes requested):

Makefile
deeplow: Force pushed with d95c049

PR #1343, opened 2025-06-05
Trigger OpenQA from GitHub Actions

### LINKED ISSUE #1319: GitHub - OpenQA integration: triggering test and getting status
GitHub PR integration testing SecureDrop Workstation repos on OpenQA.

## Requirements

**Must:**
- Trigger test for single repo [workstation repo](https://github.com/freedomofpress/securedrop-workstation) or the [client repo](https://github.com/freedomofpress/securedrop-client) OR both!
- Reporting test result back to GitHub

**Should:**
- **Reporting result** on GitHub and linking to OpenQA test result ([example](https://openqa.qubes-os.org/tests/134747))
- **Easily rerun test** (in the current CI, we have to amend the last commit and force-push, which is workable but not ideal) (**Note:** this can be already done in the OpenQA CI, so we may not need this).

**Could:**
- **Test from multiple repos**: [workstation](https://github.com/freedomofpress/securedrop-workstation) PR with [client PR](https://github.com/freedomofpress/securedrop-client). — Not critical for real parity, but good to consider when choosing approach.
- **Allow permissions delegation to start PRs** (e.g. PR contributor, for their own PR?) (see [this comment](https://github.com/freedomofpress/securedrop-workstation/issues/1319#issuecomment-2902559150))
- **Avoid wasteful runs**: start test on every push is probably unnecessary — it could just have been a lint fix.

## Background

OpenQA offers a [REST API](https://open.qa/docs/#_use_of_the_rest_api) as well as a CLI client. For practial examples see this [README](https://openqa.qubes-os.org/tests/138878#step/TC_00_Direct_debian-12-xfce/1) from the Qubes Team's openQA tests repo.

### Starting a Test on OpenQA via CLI
 We can start tests with a command like this:

```
$ openqa-cli api --host openqa.qubes-os.org isos -X post \
    VERSION=4.2 \
    BUILD=$(date +%Y%m%d%H)-4.2 \
    DISTRI=qubesos ARCH=x86_64 \
    FLAVOR=securedrop \
    CASEDIR="https://github.com/deeplow/openqa-tests-qubesos.git#securedrop-basic-functionality"\
    MAX_JOB_TIME=10800
{"count":2,"failed":[],"ids":[138956,138957],"scheduled_product_id":2785}
```

As you can see, the returned value shows two tests: [138956](https://openqa.qubes-os.org/tests/138956) and [138957](https://openqa.qubes-os.org/tests/138957). There's probably a an easy way to check in on the status through OpenQA.

### Current Workstation CI GH Action

[TKTKT @freedomofpress/fpf-infra]

Limitations:
- no way to restart job

# Alternative Solutions

## [Alternative] Github Action calling OpenQA directly 

1. GitHub Action starts an OpenQA job, passing along a parameter with the current PR URL
2. OpenQA build dom0 packages in `sd-dev`
3. runs `make test`
4. Uploads artifacs to OpenQA
5. GitHub Action detects success/failure on openQA and reports status, blocking merging while running or failed.

### [Alternative] Use Qubes - OpenQA integration

The Qubes Team has built their own integration with OpenQA. What we have bellow is @deeplow's understanding of the CI and moving pieces. If we want to consider this approach we should talk to Marek as we may be able to use their infra, if we choose to.

Here's a diagram of the system I put together:

Image

<details><summary>How it works</summary>

Constraints:
- Features may be spread around multiple repos and need to be tested together

#### Developer perspective
Let's see a practical example: a developer wants to test a feature

1. Developer chooses which PRs need to be tested together. This is a _manual_ process. It's basically clicking "new pipeline" at https://gitlab.com/QubesOS/qubes-continuous-integration/-/pipelines, you can adjust some settings there, for example all PRs with the <kbd>openqa-pending</kbd> label. Here's what the pipeline creation looks like.

    Image

    But this can, of course, be scriptable with the GitLab API.

2.  _In the background_: the packages start being built and OpenQA runs the test with those new packages.

3. GitHub bot replies with the test results. These include:
   - link to OpenQA test
   - list of PRs tested together
   - list of test results
   - list of unstable test results (have been inconsistent in last runs)

Image

<p align="center">[...]</p>

Image

<p align="center">[...]</p>

Image

4. If the developer wants to re-run the test they can go to the test in OpenQA and restart it there:
     ! [Image](https://github.com/user-attachments/assets/80222214-df8f-4ff6-b15e-77f1cead84b4)

   > **Note:** to avoid visual clutter, the test results are shown as an edit on the previous test results

5. In parallel to all of this, a GitHub action shows the status of package build on GitLab (example)

Image

Image

#### How it Works under the hood
The infrastructure is quite complex. I'll attempt to summarize the relevant bits:
- **GitHub**: where code lives. A relatively simple GitHub action reports the final [CI status](https://github.com/QubesOS/qubes-builderv2-github/blob/main/github-action.py)
- **GitLab**: where most of CI infra is run. The code is mirrored from GitHub. (They [moved from Travis](https://github.com/QubesOS/qubes-posts/blob/57e9bceef341a2dc29b323cb99c78ec215dc6670/2021-02-28-improvements-in-testing-and-building.md#gitlab-ci-migration) probably before GitHub Actions became a usable thing)
- [Qubes-BuilderV2](https://github.com/QubesOS/qubes-builderv2) tool used to build everything in Qubes OS. Every package built and tested by the infrastructure must have build files that it accepts ([see example](https://github.com/QubesOS-contrib/qubes-skeleton/)). @deeplow's read is that this would be necessary for the code to be tested with the Qubes team's infra.
- **Qubes Build Infrastructure** is self-hosted 
- OpenQA is hosted at openqa.qubes-os.org and this is the integration testing platform
- (not super relevant for CI) but they [have a page](https://qubesos.gitlab.io/qubes-g2g-report/) where the build and reproducibility status for every package is visible. 

1. After a new GitLab pipeline has been created, it builds the packages.

2. The server endpoint (running [gitlab_api.py](https://github.com/QubesOS/openqa-tests-qubesos/blob/main/api/gitlab_api.py)) looks for `publish:repo` jobs in the pipeline and imports the packages to later serve in the OpenQA test and triggers the correct openQA run using its API.

    > it should be possible to let openQA add relevant repositories pointing at gitlab.com directly (our "update2.pm" module configures extra repositories to fetch packages for testing), but that doesn't work with Whonix tests, as gitlab.com refuses connections via Tor... So we copy the files around to host them on openqa host itself which doesn't do this stupidity.
 _Originally posted by @marmarek in [#1319](https://github.com/freedomofpress/securedrop-workstation/issues/1319#issuecomment-2901450804)_

3. OpenQA does its thing running the tests and assets are uploaded and made available directly on OpenQA

4. **Displaying the results:** OpenQA results are monitored using something different: 

    > [AMQP events](https://github.com/QubesOS/openqa-tests-qubesos/blob/main/utils/amqp_consume.py) emitted by the openQA are watched, and in reaction to `qubes.openqa.job.done` event, when all jobs in a given group are done [`github_reporting.py`](https://github.com/QubesOS/openqa-tests-qubesos/blob/main/utils/github_reporting.py) is called. This collects all the relevant results via openQA API and then post a comment via github API. That last one has the logic to [decide where to post the comment(s)](https://github.com/QubesOS/openqa-tests-qubesos/blob/main/utils/lib/openqa_api.py#L294-L307). Currently we don't set PR status (or rather commit status)  based on that, as we have too many flaky tests, but if desirable, that would be the place. We can easily extend it to report to other places too (for example whonix tests results are already reported into github.com/Whonix). And we can also extend it to set PR status for SD tests, if you wish (currently the script does nothing for jobs with `FLAVOR=securedrop`).

</details>

legoktm:
## Status

Ready for review; depends on #1329.

## Description of Changes

Trigger OpenQA from GitHub Actions
    
A GHA job will trigger our jobs in OpenQA and then monitor them until
they finish.

The OpenQA job will only be started if it is a direct push that is to a
branch that isn't `main` nor part of the GH merge queue. This is to
avoid duplicate jobs while it's still quite slow; eventually this should
be a pre-merge check.

Because this is a slow job running on Qubes hardware, it's trivial to
skip running it if the latest commit message mentions the string "skip"
and "openqa" separated by a single space, case insensitive. This is
what the `check-openqa` job does and is based off the "skip" "ci"
keywords that GHA itself supports.

When the GHA job is cancelled, we'll also send off requests to cancel
the openQA jobs in a best-effort manner.

Finally we include links to the openQA logs at the bottom just to make
it easy for humans to hunt down failures.

This work was initially based off of
https://github.com/os-autoinst/os-autoinst-distri-example/blob/main/.github/workflows/openqa.yml
and deeplow's example in #1319.

Note that the `openqa-cli monitor` command is only available in trixie.

Fixes #1319

## Testing

* [ ] CI correctly executes the openQA jobs as expected
* [ ] using the `skip openqa` magic string in a commit works (no tests are run)
* [ ] cancelling a workflow also stops the jobs in openQA.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,15 @@
 ---
 name: ci
 
-on: [push, pull_request, merge_group]
+on:
+  - push
+  - pull_request
+  - merge_group
+
+# Only build for latest push/PR unless it's main or release/
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' && !startsWith( github.ref, 'refs/heads/release/' ) && !startsWith( github.ref, 'refs/heads/gh-readonly-queue/' ) }}
 
 jobs:
   lint:
@@ -59,3 +67,93 @@ jobs:
       - name: Run launcher tests
         run: |
           make test-launcher
+  # If the most recent commit message contains "skip openqa" (case insensitive), skip the openqa job
+  check-openqa:
+    runs-on: ubuntu-latest
+    outputs:
+      should-skip: ${{ steps.check.outputs.should-skip }}
+    env:
+      COMMIT_MSG: ${{ github.event.head_commit.message }}
+    steps:
+      - name: Check commit message
+        id: check
+        run: |
+          if echo "${COMMIT_MSG}" | grep -iq "skip openqa"; then
+            echo "skipping OpenQA job"
+            echo "should-skip=true" >> $GITHUB_OUTPUT
+          else
+            echo "running OpenQA job"
+            echo "should-skip=false" >> $GITHUB_OUTPUT
+          fi
+  openqa:
+    runs-on: ubuntu-latest
+    # Do not start until all other checks have passed
+    needs:
+      - lint
+      - build-rpm
+      - launcher-tests
+      - check-openqa
+    # Run only if:
+    # 1. This is a push (not pull_request)
+    # 2. We're not on main
+    # 3. We're not on a merge queue branch
+    # 4. The "skip openqa" string isn't present
+    if: |
+      github.event_name == 'push' && github.ref != 'refs/heads/main' && !startsWith( github.ref, 'refs/heads/gh-readonly-queue/' )
+      && needs.check-openqa.outputs.should-skip != 'true'
+    env:
+      OPENQA_HOST: "https://openqa.qubes-os.org/"
+      OPENQA_KEY: ${{ secrets.OPENQA_KEY }}
+      OPENQA_SECRET: ${{ secrets.OPENQA_SECRET }}
+      GIT_REF: ${{ github.event.head_commit.id }}
+    container:
+      image: debian:trixie
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - name: Install dependencies
+        run: |
+          apt-get update
+          apt-get install --yes openqa-client jq
+      - name: Start openQA jobs
+        run: |
+          echo "Queuing openQA jobs for commit ${GIT_REF}"
+          openqa-cli api \
+            --host "$OPENQA_HOST" \
+            --apikey "$OPENQA_KEY" \
+            --apisecret "$OPENQA_SECRET" \
+            isos -X post \
+            VERSION=4.2 \
+            BUILD=$(date +%Y%m%d%H%M)-4.2 \
+            DISTRI=qubesos ARCH=x86_64 \
+            GIT_REF="$GIT_REF" \
+            FLAVOR=securedrop \
+            CASEDIR="https://github.com/deeplow/openqa-tests-qubesos.git#securedrop-basic-functionality"\
+            MAX_JOB_TIME=10800 | tee openqa.json
+      - name: Wait for openQA
+        run: |
+          openqa-cli monitor \
+            --host "$OPENQA_HOST" \
+            --apikey "$OPENQA_KEY" \
+            --apisecret "$OPENQA_SECRET" \
+            $(jq -r '.ids[]' openqa.json)
+      - name: Cancel openQA test (if needed)
+        if: cancelled()
+        run: |
+          # Cancel all the jobs we queued in a best-effort manner
+          for id in $(jq -r '.ids[]' openqa.json); do
+            openqa-cli api \
+              --host "$OPENQA_HOST" \
+              --apikey "$OPENQA_KEY" \
+              --apisecret "$OPENQA_SECRET" \
+              -X POST "jobs/${id}/cancel" ||:
+          done
+      - name: View logs
+        if: always()
+        run: |
+          echo "View the logs for this run at:"
+
+          for id in $(jq -r '.ids[]' openqa.json); do
+            echo "${OPENQA_HOST}tests/${id}"
+          done
--- a/Makefile
+++ b/Makefile
@@ -135,29 +135,39 @@ clean: assert-dom0 prep-dev ## Destroys all SD VMs
 # if clean has already been run.
 	./scripts/sdw-admin.py --uninstall --force
 
+PYTEST_PATH=$(shell command -v pytest)
+COVERAGE_PATH=$(shell command -v coverage) # instead of pytest-cov, check for its dep.
 .PHONY: test-prereqs
 test-prereqs: assert-dom0 ## Checks that test prerequisites are satisfied
 	@echo "Checking prerequisites before running test suite..."
 	test -e config.json || exit 1
 	test -e sd-journalist.sec || exit 1
+ifeq ($(PYTEST_PATH),)
+	@echo 'please install pytest with "sudo qubes-dom0-update python3-pytest"'
+	@false
+endif
+ifeq ($(COVERAGE_PATH),)
+	@echo 'please install pytest with "sudo qubes-dom0-update python3-pytest-cov"'
+	@false
+endif
 
 test: test-prereqs ## Runs all application tests (no integration tests yet)
-	python3 -m unittest discover -v tests
+	pytest -v tests
 
 test-base: test-prereqs ## Runs tests for VMs layout
-	python3 -m unittest discover -v tests -p test_vms_exist.py
+	pytest -v tests/test_vms_exist.py
 
 test-app: test-prereqs ## Runs tests for SD APP VM config
-	python3 -m unittest discover -v tests -p test_app.py
+	pytest -v tests/test_app.py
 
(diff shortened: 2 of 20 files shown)


legoktm:
Still in draft mode since it depends on the pytest PR. Also I haven't fully tested cancellation yet.

legoktm:
> Also I haven't fully tested cancellation yet.

I've tested this now and it works as expected, see https://github.com/freedomofpress/securedrop-workstation/issues/1319#issuecomment-2960404197.

.github/workflows/ci.yml
```diff
@@ -1,7 +1,15 @@
 ---
 name: ci
 
-on: [push, pull_request, merge_group]
+on:
+  - push
+  - pull_request
+  - merge_group
+
+# Only build for latest push/PR unless it's main or release/
```
legoktm, line 9: This is copied from securedrop-client. So if you push to a branch, it'll cancel any running jobs and start a new one.

.github/workflows/ci.yml
```diff
@@ -59,3 +67,93 @@ jobs:
       - name: Run launcher tests
         run: |
           make test-launcher
+  # If the most recent commit message contains "skip openqa" (case insensitive), skip the openqa job
+  check-openqa:
```
legoktm, line 71: Not sure if this job name is misleading or if there's a better option

.github/workflows/ci.yml
```diff
@@ -59,3 +67,93 @@ jobs:
...
+            --host "$OPENQA_HOST" \
+            --apikey "$OPENQA_KEY" \
+            --apisecret "$OPENQA_SECRET" \
+            isos -X post \
+            VERSION=4.2 \
+            BUILD=$(date +%Y%m%d%H%M)-4.2 \
+            DISTRI=qubesos ARCH=x86_64 \
+            GIT_REF="$GIT_REF" \
+            FLAVOR=securedrop \
+            CASEDIR="https://github.com/deeplow/openqa-tests-qubesos.git#securedrop-basic-functionality"\
```
legoktm, line 133: This will eventually point at the upstream QubesOS repo, but also can be modified to test changes to the openqa config.

.github/workflows/ci.yml
```diff
@@ -1,7 +1,15 @@
 ---
 name: ci
 
-on: [push, pull_request, merge_group]
+on:
+  - push
```
deeplow, line 5: Could this trigger an OpenQA job in the group `ci-<COMMIT_HASH>` instead of `ci-<BRANCH_NAME>?

.github/workflows/ci.yml
```diff
@@ -59,3 +67,94 @@ jobs:
...
+    runs-on: ubuntu-latest
+    outputs:
+      should-skip: ${{ steps.check.outputs.should-skip }}
+    env:
+      COMMIT_MSG: ${{ github.event.head_commit.message }}
+    steps:
+      - name: Check commit message
+        id: check
+        run: |
+          if echo "${COMMIT_MSG}" | grep -iq "skip openqa"; then
```
deeplow: We may end up being a bit annoying in terms of editing commit messages, but I guess it'll work for now.

But one thing I'd suggest is starting white-list first and then switching back. Specifically, I'd suggest that we mark instead with "test openqa" just so if there is a bug we don't accidentally trigger too many jobs. Once we're confident, we flip the switch.

.github/workflows/ci.yml
deeplow: I've also left a comment on push behavior in https://github.com/freedomofpress/securedrop-workstation/issues/1319#issuecomment-2968024116

legoktm: The intent is that when you push (or force push) to update a branch/PR, it cancels any existing jobs and queues up a new one. Using a group based on the commit hash wouldn't do that, because pushing a different commit changes the hash.

Or did I misunderstand what you meant?

.github/workflows/ci.yml
legoktm: Ack, that makes sense, will do.

.github/workflows/ci.yml
deeplow: > Could this trigger an OpenQA job in the group ci-<COMMIT_HASH> instead of `ci-<BRANCH_NAME>?

Sorry. I meant to say: is there there a risk this [Could this trigger an OpenQA job in the group ci-<COMMIT_HASH> instead of `ci-<BRANCH_NAME>?]

legoktm: Ah, no. https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context it's documented as always being the branch or tag, but we don't trigger on tags, so it's always going to be a branch name :)

deeplow: Perfect. I'll close this now, then.

Issue #1338, opened 2025-06-05
Full pytest conversion for dom0 tests


deeplow:
In #1337 the goal was to have initial support for pytest in dom0. However, this leaves many tests in need of conversion (search for `unittest` and you'll still find many references missing conversion).

While doing this migration we could use the opportunity to take advantage of fixtures and other pytest features that allow for less repetitive and implicit testing.

Examples of things to refactor:
  - <strike>replace all [sdw_tag-finding functions](https://github.com/freedomofpress/securedrop-workstation/blob/e653eb3/tests/test_vms_exist.py#L29) and have instead a shared fixture.</strike> (done in https://github.com/freedomofpress/securedrop-workstation/pull/1487)
  - Remove all `unittest` references and replace with pytest ones (e.g. [skipif](https://github.com/freedomofpress/securedrop-workstation/pull/1487#discussion_r2539418618))
  - Use `pytest.fail` instead of `self.fail`
  - Remove all `unittest.TestCase` classes

PR #1346, opened 2025-06-10
tests: use policy graph in qubes rpc tests

### LINKED ISSUE #1331: Replace RPC tests with `qrexec-policy-graph`

## Description

[`test_qubes_rpc.py](https://github.com/freedomofpress/securedrop-workstation/blob/1.2.0/tests/test_qubes_rpc.py) is highly inefficient and only cares about [currently running VMs](https://github.com/freedomofpress/securedrop-workstation/blob/1.2.0/tests/test_qubes_rpc.py#L28) (for some reason).

The suggestion here would be to replace this test entirely with a dedicated qubes tool (`qrexec-policy-graph`) to check current policies without needing to actually run them. This would would ensure proper coverage (no just the running qubes), running instantly as opposed to taking many seconds and not generate the `securedrop.Log` failed notifications on the screen.

Example:
```
[user@dom0 ~]$ qrexec-policy-graph --target sd-log --service securedrop.Log
WARNING:root:warning: !compat-4.0 directive in file /etc/qubes/policy.d/35-compat.policy line 16 is transitional and will be deprecated
digraph g {
  "sd-app" -> "sd-log" [label="securedrop.Log" color=red];
  "sd-base-bookworm-template" -> "sd-log" [label="securedrop.Log" color=red];
  "sd-devices" -> "sd-log" [label="securedrop.Log" color=red];
  "sd-devices-dvm" -> "sd-log" [label="securedrop.Log" color=red];
  "sd-large-bookworm-template" -> "sd-log" [label="securedrop.Log" color=red];
  "sd-proxy" -> "sd-log" [label="securedrop.Log" color=red];
  "sd-proxy-dvm" -> "sd-log" [label="securedrop.Log" color=red];
  "sd-small-bookworm-template" -> "sd-log" [label="securedrop.Log" color=red];
  "sd-viewer" -> "sd-log" [label="securedrop.Log" color=red];
  "sd-whonix" -> "sd-log" [label="securedrop.Log" color=red];
}
```

vickiniu:
## Status

Ready for review

## Description of Changes

Fixes #1331 

Changes proposed in this pull request: uses `qrexec-policy-graph` tool for qubes rpc tests to check policies rather than running the VMs + checking return codes.

### DIFF
--- a/tests/test_qubes_rpc.py
+++ b/tests/test_qubes_rpc.py
@@ -1,72 +1,62 @@
-import os
+import functools
 import subprocess
 import unittest
 
 from qubesadmin import Qubes
 
-RETURNCODE_SUCCESS = 0
-RETURNCODE_ERROR = 1
-RETURNCODE_DENIED = 126
-
 
 class SD_Qubes_Rpc_Tests(unittest.TestCase):
-    def _qrexec(self, source_vm, dest_vm, policy_name):
-        cmd = [
-            "qvm-run",
-            "--pass-io",
-            source_vm,
-            f"qrexec-client-vm {dest_vm} {policy_name}",
-        ]
-        p = subprocess.run(cmd, input="test", capture_output=True, text=True, check=False)
-        return p.returncode
+    @classmethod
+    def setUpClass(cls):
+        cls.all_vms = set()
+        cls.vms_by_tag = {}
 
-    def _get_running_vms_with_and_without_tag(self, tag):
-        vms_with_tag = []
-        vms_without_tag = []
         app = Qubes()
-        for vm in app.domains:
-            if vm.name != "dom0" and vm.is_running():
-                if tag in list(vm.tags):
-                    vms_with_tag.append(vm.name)
-                else:
-                    vms_without_tag.append(vm.name)
-        return vms_with_tag, vms_without_tag
+        cls.all_vms = {vm for vm in app.domains if vm.name != "dom0"}
+        cls.sdw_tagged_vms = {vm for vm in app.domains if "sd-workstation" in vm.tags}
 
-    def test_policies_exist(self):
-        """verify the policies are installed"""
-        assert os.path.exists("/etc/qubes/policy.d/31-securedrop-workstation.policy")
-        assert os.path.exists("/etc/qubes/policy.d/32-securedrop-workstation.policy")
+    @functools.cache
+    def _qrexec_policy_graph(self, service):
+        cmd = ["qrexec-policy-graph", "--service", service]
+        p = subprocess.run(cmd, capture_output=True, text=True, check=False)
+        return p.stdout
+
+    def _policy_exists(self, source, target, service):
+        service_policy_graph = self._qrexec_policy_graph(service)
+        policy_str = f'"{source}" -> "{target}" [label="{service}"'
+        return policy_str in service_policy_graph
 
     # securedrop.Log from @tag:sd-workstation to sd-log should be allowed
     def test_sdlog_from_sdw_to_sdlog_allowed(self):
-        vms_with_tag, _ = self._get_running_vms_with_and_without_tag("sd-workstation")
-        for vm in vms_with_tag:
+        for vm in self.sdw_tagged_vms:
             if vm != "sd-log":
-                assert self._qrexec(vm, "sd-log", "securedrop.Log") == RETURNCODE_SUCCESS
+                assert self._policy_exists(vm, "sd-log", "securedrop.Log")
 
     # securedrop.Log from anything else to sd-log should be denied
     def test_sdlog_from_other_to_sdlog_denied(self):
-        _, vms_without_tag = self._get_running_vms_with_and_without_tag("sd-workstation")
-        for vm in vms_without_tag:
+        non_sd_workstation_vms = self.all_vms.difference(self.sdw_tagged_vms)
+        for vm in non_sd_workstation_vms:
             if vm != "sd-log":
-                assert self._qrexec(vm, "sd-log", "securedrop.Log") == RETURNCODE_DENIED
+                assert not self._policy_exists(vm, "sd-log", "securedrop.Log")
 
     # securedrop.Proxy from sd-app to sd-proxy should be allowed
     def test_sdproxy_from_sdapp_to_sdproxy_allowed(self):
-        # proxy RPC returns an error due to malformed input, but it still goes through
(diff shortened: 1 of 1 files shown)


tests/test_qubes_rpc.py
```diff
@@ -4,33 +4,33 @@
...
+                    cls.vms_by_tag[tag].append(vm.name)
                 else:
-                    vms_without_tag.append(vm.name)
-        return vms_with_tag, vms_without_tag
+                    cls.vms_by_tag[tag] = [vm.name]
+
+    def _qrexec_policy_graph(self, source, target, service):
+        cmd = ["qrexec-policy-graph", "--source", source, "--target", target, "--service", service]
+        p = subprocess.run(cmd, capture_output=True, text=True, check=False)
+        return p.stdout
```
deeplow: I think we can improve this by another order of magnitude if we call the policy graph less times. One way to do this would be to change the tests semantics so they adapt to qrexec_policy_graph (I don't think it's worth it). The other way is to make more generic calls to qrexec-policy graph (and cache results). So this would be my suggestion.

Here's how this could look like (without necessary imports nor calling function adaptations)

```suggestion
    @functools.cache  # needs importing
    def _qrexec_policy_graph(self, service):
        cmd = ["qrexec-policy-graph", "--service", service]
        p = subprocess.run(cmd, capture_output=True, text=True, check=False)
        return p.stdout
```

What do you think?

tests/test_qubes_rpc.py
```diff
@@ -4,33 +4,33 @@
...
+
+    def _policy_graph_rule_exists(self, source, target, service):
```
deeplow: Shall we make this a bit shorter to make test calls easier to read? It think it's fine if the fact that we're using the policy graph stays as an internal detail.

```suggestion
    def policy_exists(self, source, target, service):
```

tests/test_qubes_rpc.py
```diff
@@ -39,43 +39,75 @@ def test_policies_exist(self):
...
-        for vm in vms_without_tag:
+        non_sd_workstation_vms = self.all_vms.difference(self.vms_by_tag["sd-workstation"])
+        for vm in non_sd_workstation_vms:
             if vm != "sd-log":
-                self.assertEqual(self._qrexec(vm, "sd-log", "securedrop.Log"), RETURNCODE_DENIED)
+                self.assertFalse(
+                    self._policy_graph_rule_exists(
+                        source=vm, target="sd-log", service="securedrop.Log"
+                    )
+                )
```
deeplow: Small nitpick, do you think we could make this into a one-line call? Perhaps removing the arguments? I think after dealing with qrexec policies it becomes a bit intuitive which is which. But if you find the keyword args helpful, then I'd say we keep them for clarity.

In my mind, once we move this to pytest, it could as clear as:

```
assert not policy_exists(vm, "sd-log", "securedrop.Log")
```

tests/test_qubes_rpc.py
deeplow: Or if you'd like to keep it generic, we could keep all the arguments, but construct cmd from the parameters which are not `None` and then call it just the service:

```suggestion
    @functools.cache
    def _qrexec_policy_graph(self, source, target, service):
        cmd = ["qrexec-policy-graph"]
        if source:
          cmd += ["--source", source]
         if target:
           ["--target", target]
         if service:
           cmd += ["--service", service]
        p = subprocess.run(cmd, capture_output=True, text=True, check=False)
        return p.stdout
```

vickiniu: great suggestion, thank you!! just updated to call `qrexec-policy-graph` once per-service. now the tests run in 2.6s on my machine :)

deeplow: Perfect. That's matches with my own testing.

tests/test_qubes_rpc.py
```diff
@@ -1,36 +1,38 @@
...
+                continue
+            cls.all_vms.add(vm.name)
+            for tag in vm.tags:
+                if tag in cls.vms_by_tag:
+                    cls.vms_by_tag[tag].append(vm.name)
                 else:
-                    vms_without_tag.append(vm.name)
-        return vms_with_tag, vms_without_tag
+                    cls.vms_by_tag[tag] = [vm.name]
+
```
deeplow: I missed this during my first review. I see that you generalized the tags searching and the original function already did some of that, but we mostly care about <kbd>sd-workstation</kbd>-tagged VMs. We have [this](https://github.com/freedomofpress/securedrop-workstation/blob/e653eb3/tests/test_vms_exist.py#L29) in other places as well.

So, I'd suggest something as simple as:

```suggestion
        app = Qubes()
        cls.all_vms = [vm for vm in app.domains if vm.name != "dom0"]
        cls.sdw_tagged_vms = [vm for vm in app.domains if "sd-workstation" in vm.tags]
```

(with the necessary adaptations in the remaining code-base)

Later we'll want to "DRY" them all in https://github.com/freedomofpress/securedrop-workstation/issues/1338, but let's leave that for later. Repetition is fine for now.

tests/test_qubes_rpc.py
```diff
@@ -39,34 +41,36 @@ def test_policies_exist(self):
...
+        assert not self._policy_exists("sys-net", "sd-gpg", "qubes.Gpg")
+        assert not self._policy_exists("sys-firewall", "sd-gpg", "qubes.Gpg")
+        assert not self._policy_exists("sys-net", "sd-gpg", "qubes.GpgImportKey")
+        assert not self._policy_exists("sys-firewall", "sd-gpg", "qubes.GpgImportKey")
+        assert not self._policy_exists("sys-net", "sd-gpg", "qubes.Gpg2")
+        assert not self._policy_exists("sys-firewall", "sd-gpg", "qubes.Gpg2")
+
+
+def load_tests(loader, tests, pattern):
+    return unittest.TestLoader().loadTestsFromTestCase(SD_Qubes_Rpc_Tests)
```
deeplow, line 62: Let's also remove now this is "pytest" code. We'll later do a pass on all tests to ensure full pytest migration (https://github.com/freedomofpress/securedrop-workstation/issues/1338), but that's not necessary for now. 

But because this is now removed in main, I think we shouldn't add it back.

tests/test_qubes_rpc.py
vickiniu: ah, makes sense!! much simpler now with just the `sd-workstation` tagged vms :)

PR #1349, opened 2025-06-17
sdw-admin: add --configure to import submission key + workstation config

### LINKED ISSUE #942: Add --configure to sdw-admin script, to replace manual key and config.json configuration steps
(I think there was talk of GUI installers and such before, but can't see any relevant issues. This is a suggestion for a more lightweight approach.)

Currently prod SDW  configuration and installation  involves manual attachments of USB keys along with "qvm-run --pass-io" commands to copy files to dom0. The rough flow is as follows:

1. configure a network-attached VM to download the dom0 rpm, copy and install it
2. attach the SVS to a non-networked VM, copy the privkey to dom0
3. attach the JW/AW, copy the JI service details to dom0
4. (optionally copy the JW/AW keepassxc vault to a vault VM)
5. copy the key into place and retrieve its fingerprint
6. create config.json and add in the fingerprint and service details via a text editor
7 run sdw-admin --validate, then sdw-admin--apply

There is plenty of opportunity for error and frustration in the above, and it should be scriptable. I'd propose something more like:

1. configure a network-attached VM to download the dom0 rpm, copy and install it (this could be simplified a little but some variation is probably unavoidable as a bootstrapping step)
2. Run `sdw-admin --configure` which will:
  1. check for the existence of a valid privkey, and either prompt the user to change it if found or prompt the user to add a new key (guiding through device attachment and validating that the VM chosen is non-networked) directly under /usr/share/...
  2. read config.json if it exists and prompt the user to confirm the values found, or prompt the user to provide new ones (guiding attachment and then reading from the expected location directly) 
  3. validate the new values and give the user the option to save and optionally apply them

Desired behaviours (attaching sticks to non-networked VMs etc) could be checked and enforced by the script.

vickiniu:
## Status

Ready for review

## Description of Changes

Fixes #942 

Adds `--configure` option to `sdw-admin` which (1) checks for submission key in `dom0` and, if not found, prompts the user to insert submission key drive + handles import, and (2) checks for valid `config.json` and, if not found, prompts user to insert workstation drive and imports Journalist Interface config.

## Testing

### DIFF
--- a/files/sdw-admin.py
+++ b/files/sdw-admin.py
@@ -295,7 +295,6 @@ def extract_secret_key_fingerprints(gpg_output):
     lines = gpg_output.strip().split("\n")
     fingerprints = []
 
-    primary_key = True
     for idx, line in enumerate(lines):
         if not line.strip():
             continue


files/sdw-admin.py
```diff
@@ -270,6 +288,185 @@ def perform_uninstall():
...
+        sd_log_gb = input(
+            f"Enter desired size for sd-log private volume in GiB (default: {DEFAULT_SD_LOG_GB}GiB)"
+        )
+        if not sd_log_gb:
+            sd_log_gb = DEFAULT_SD_LOG_GB
+
+        config = {
+            "submission_key_fpr": submission_key_fingerprint,
+            "hidserv": {
+                "hostname": ji_addr,
```
zenmonkeykstop: ji_addr is the key part only of the onion -  it's missing the `.onion` part.

files/sdw-admin.py
```diff
@@ -270,6 +288,185 @@ def perform_uninstall():
...
+    if not os.path.exists(SCRIPTS_PATH + SUBMISSION_KEY):
+        return None
+    gpg_output = subprocess.check_output(
+        ["gpg", "--show-keys", "--with-fingerprint", "--with-colon", SCRIPTS_PATH + SUBMISSION_KEY],
+        text=True,
+    )
+    return extract_fingerprint(gpg_output)
+
+
+def import_submission_key():
```
zenmonkeykstop, line 363: This assumes there's a single secret key in the keyring - there may be multiple ones if the submission key has been rotated previously. For now we just support using one key during provisioning, so if there are multiple keys there are a couple of possible approaches:
- (preferred) list the secret key user ids and fingerprints found, prompt the user to choose one (either via a numerical list or by typing the full fingerprint)
- exit with an error, saying multiple submission keys are not supported at installation
- something else?

files/sdw-admin.py
```diff
@@ -270,6 +288,185 @@ def perform_uninstall():
...
+                "key": ji_auth_token,
+            },
+            "environment": "prod",
+            "vmsizes": {"sd_app": sd_app_gb, "sd_log": sd_log_gb},
+        }
+        temp_file = "/tmp/config.json"
+        with open(temp_file, "w") as f:
+            json.dump(config, f, indent=2)
+        subprocess.check_call(["sudo", "cp", temp_file, SCRIPTS_PATH])
+        print(
```
zenmonkeykstop, line 516: Confirming the values found here would be nice but is optional.

files/sdw-admin.py
```diff
@@ -270,6 +288,185 @@ def perform_uninstall():
...
+        )
+    else:
+        print("Found submission key file, proceeding")
+
+    try:
+        validate_config(SCRIPTS_PATH)
+        print("Valid configuration found, configuration complete")
+    except SDWAdminException:
+        subprocess.Popen(
+            ["qvm-start", "vault"], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL
```
zenmonkeykstop, line 465: One other option here could be to spin up a non-networked dispvm instead, but it might be overkill.

zenmonkeykstop (changes requested):
This is already looking like a great improvement over the manual flow - passing it back with one bug and a couple of suggestions for improvement.

files/sdw-admin.py
vickiniu: Great callout! Changed to list the secret key fingerprints + prompt the user to choose if there are multiple keys in the keyring

zenmonkeykstop (approved):
LGTM, confirmed that the new flow works smoothly.

zenmonkeykstop:
All good!

Issue #1333, opened 2025-06-02
dom0 tests FAIL: `test_sd_vm_apt_sources` (`apt-test` instead of expected `apt.freedom.press`)


deeplow:
## Description

CI failed recently, I believe on main as well ([see the log](https://ws-ci-runner.securedrop.org/2025-05-30-063003879727-22933834c7592f0bb0966ce404232ec0ad1880f0-Qubes_4.2_D-update_20250530043423.log.txt)).

The failure is as follows:

`AssertionError: 'Description: https://apt.freedom.press bookworm/main amd64 Packages\n' not found in <'apt-get indextargets' output>`

<details><summary>Full CI failure</summary>

```
INFO:[2025-05-30-07:31:09:305975] ======================================================================
INFO:[2025-05-30-07:31:09:306010] FAIL: test_sd_vm_apt_sources (test_vms_platform.SD_VM_Platform_Tests.test_sd_vm_apt_sources)
INFO:[2025-05-30-07:31:09:306050] Test that the three templates we install our apt sources into are correct
INFO:[2025-05-30-07:31:09:306083] ----------------------------------------------------------------------
INFO:[2025-05-30-07:31:09:306139] Traceback (most recent call last):
INFO:[2025-05-30-07:31:09:306180]   File "/home/user/securedrop-workstation/tests/test_vms_platform.py", line 197, in test_sd_vm_apt_sources
INFO:[2025-05-30-07:31:09:306217]     self.assertIn(
[...]

PR #1356, opened 2025-07-02
CI: Run OpenQA tests from main upstream branch

deeplow:
## Description of Changes

With the merging of the base workstation tests [1], we can now point to main.

[1]: https://github.com/QubesOS/openqa-tests-qubesos/pull/35

## Testing

OpenQA tests are passing.

### DIFF
--- a/tests/test_vms_platform.py
+++ b/tests/test_vms_platform.py
@@ -194,6 +194,7 @@ class SD_VM_Platform_Tests(unittest.TestCase):
             vm = self.app.domains[vm_name]
             # First verify it looks like what we provisioned
             self._validate_apt_sources(vm)
+
             stdout, stderr = vm.run("apt-get indextargets")
             contents = stdout.decode().strip()
             assert (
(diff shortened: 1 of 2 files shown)


tests/test_vms_platform.py
```diff
@@ -194,6 +194,10 @@ def test_sd_vm_apt_sources(self):
             vm = self.app.domains[vm_name]
             # First verify it looks like what we provisioned
             self._validate_apt_sources(vm)
+
+            # Refresh sources so test runs in CI, where updater has not yet run
+            vm.run("sudo apt-get update")
```
eloquence: Is this intended to be part of this change?

deeplow: Thanks for catching this! I'll fix it. I generally never do these mistakes because I use `git --add patch` to select the chunks I want.

But it turns out in this case I had already staged that change (for https://github.com/freedomofpress/securedrop-workstation/issues/1333), then removed it from the code (I preferred to [on the CI instead](https://github.com/freedomofpress/securedrop-workstation-ci/pull/84)) and `git --add patch` was asking me if I wanted to approve the removal from staging... I interpreted it as adding instead and rejected. This was the leftover from a commit I was about to make for , but then realized I wanted to do it . So I already had staged changes and `git add --patch` was suggesting the removal of that, which is what I wanted.  :face_exhaling:

deeplow: Fix force-pushed as f282f73

PR #1362, opened 2025-07-08
Run launcher tests dom0

### LINKED ISSUE #1334: Running launcher tests in dom0

## Description

Now that we'll have pytest in dom0 (https://github.com/freedomofpress/securedrop-workstation/issues/1320), we have the ability to run launcher tests in dom0 as well. These already run in CI (GH actions), but some tests were not written because we were missing Qubes integration.

deeplow:
## Description of Changes

Fixes #1334 and also fixes a test that was actually not functionally correct.

## Testing

- make sure OpenQA is running also the launcher tests
- visual check
- make test runs launcher tests as well
- `make test-launcher` works in dom0 as well as `sd-dev`

## Deployment

(dev-facing. No deployment considerations)

### DIFF
--- a/launcher/tests/test_signing_key.py
+++ b/launcher/tests/test_signing_key.py
@@ -1,21 +1,17 @@
 import socket
-
-import pytest
-
-if socket.gethostname() == "dom0":
-    pytest.skip(
-        reason="not running due to unavailable dom0 python modules", allow_module_level=True
-    )
-
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
 
-import pysequoia
-from debian import deb822
+if socket.gethostname() != "dom0":
+    import pysequoia
+    from debian import deb822
+
+from conftest import skip_in_dom0
 
 ROOT = Path(__name__).parent.parent
 
 
+@skip_in_dom0
 def test_apt_sources():
     """Verify the key in the sources file is our prod signing key"""
     path = ROOT / "securedrop_salt/apt_freedom_press.sources.j2"
@@ -24,6 +20,7 @@ def test_apt_sources():
     assert_key(sources["Signed-By"].encode())
 
 
+@skip_in_dom0
 def test_dom0_key():
     path = ROOT / "securedrop_salt/securedrop-release-signing-pubkey-2021.asc"
     assert_key(path.read_bytes())
--- a/launcher/tests/test_signing_key.py
+++ b/launcher/tests/test_signing_key.py
@@ -1,3 +1,12 @@
+import socket
+
+import pytest
+
+if socket.gethostname() == "dom0":
+    pytest.skip(
+        reason="not running due to unavailable dom0 python modules", allow_module_level=True
+    )
+
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
 
--- a/launcher/tests/test_signing_key.py
+++ b/launcher/tests/test_signing_key.py
@@ -8,6 +8,10 @@ if socket.gethostname() != "dom0":
 
 from conftest import skip_in_dom0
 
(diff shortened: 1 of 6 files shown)


rocodes:
@deeplow Can you remind me - which openqa jobs we would expect to see succeeding here? I see that the `wait for openqa` action didn't complete, should we run it again?  

https://github.com/freedomofpress/securedrop-workstation/actions/runs/16168135641/job/45635001332?pr=1362

adaFPF (changes requested):
retest

rocodes:
Nice, thank you! Do you know why it's still showing up as a failure in github ci?

launcher/tests/test_signing_key.py
```diff
@@ -1,12 +1,17 @@
...
+if socket.gethostname() != "dom0":
+    import pysequoia
+    from debian import deb822
+
+from conftest import skip_in_dom0
 
 ROOT = Path(__name__).parent.parent
 
 
+@skip_in_dom0
```
rocodes, line 18: One question: why do we skip these in dom0?  With the keyring changes, they will eventually be changed (eg to something like `assert_keyring_installed()`), but that should run in the dom0 environment nonetheless.

launcher/tests/test_sources.py
```diff
@@ -5,14 +5,19 @@
...
+if socket.gethostname() != "dom0":
+    import pysequoia
+    from debian import deb822
+
+from conftest import skip_in_dom0
 
 SECUREDROP_SALT = Path(__file__).parent.parent.parent / "securedrop_salt"
 
 
+@skip_in_dom0
```
rocodes, line 25: (same question as above re: why we skip these)

launcher/tests/test_signing_key.py
deeplow: A couple of things that may make this worth its own follow-up PR:
  - The python modules used for checking aren't available in dom0. Would you prefer a direct comparison with the string we expect?
  - The actual deployed file is conditional on whether this is `prod` or `dev`. Since we only have the `dev` flavor, we'd be missing the other variant. But I guess it can be argued that skipping is a worse alternative.
  - with all the keyring stuff in flux, whatever we put here is likely to change in the near future.

launcher/tests/test_sources.py
deeplow: (replied in the other thread)

launcher/tests/test_signing_key.py
legoktm: These tests were kind of a hack, I dropped them in the launcher/ folder because these tests weren't run in dom0 so I could rely on the poetry-installed pysequoia package (they have nothing to do with the launcher). Plus if we do switch to having dynamically generated `.sources` files, then we'd get rid of the signing key in this repo and no longer need to verify it isn't expired :)

Now that we're running these in both dom0 and CI, I think `@skip_in_dom0` is a reasonable workaround for now, but I would want us to add a comment explaining why like:

```
# Skip in dom0 since this relies on poetry-installed dependencies.
# This test may go away later once the .sources files are dynamically generated by the keyring package.
```

rocodes: Thanks, this all makes sense.  Here's what I propose - leave them as skipped, once this is merged I'll adjust them to make them keyring-compatible.

rocodes (approved):
Thank you @deeplow - LGTM. As mentioned above, I'll add a followup commit (probably in https://github.com/freedomofpress/securedrop-workstation/pull/1210) to refactor the signing key checks that are in the launcher right now so that they're keyring-compatible and can be run in dom0.

Issue #1245, opened 2025-02-12
Ensure Support and Transition to Qubes 4.3


deeplow:
* [x] I have searched for duplicates or related issues

## Description
Ensure support and timely transition to Qubes 4.3. 

## How will this impact [SecureDrop/SecureDrop Workstation users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?
There will be an in-place upgrade. It may require some admin follow-up.

## How would this affect the SecureDrop Workstation [threat model](https://github.com/freedomofpress/securedrop-workstation/#threat-model)?
It shouldn't unless some fundamental Qubes behavior changes in a way that breaks some of our assumptions.

legoktm:
IIRC `sudo dnf install` in dom0 does not resolve dependencies (per https://github.com/freedomofpress/securedrop-workstation/issues/1063).

Presumably those were always pre-installed in dom0 so it happened to work, if you `sudo qubes-dom0-update -y python3-qt5` what happens?

Also note that the default Python version on Fedora 41 is actually 3.13; we should just install `python3`. Fedora provides all the different versions of Python to make testing easier, but per https://developer.fedoraproject.org/tech/languages/python/multiple-pythons.html:
> Warning: For production purposes you should use the python3 package only. Other CPython versions might be unstable or even dangerous (either because they are extremely old or quite the contrary alpha/beta quality) and are intended solely for development.

deeplow:
I have updated the original post to reflect the current status.

deeplow:
Closing this one since the upgrade is well under way. We can separate follow-up issues if something comes up, but the core needs have been met.

The things left to tackle are mostly taking advantage of some new features, but they require boarder architectural discussion and can be done at a later point.

deeplow:
Extra follow up: https://github.com/freedomofpress/securedrop-workstation/issues/1718

Issue #1266, opened 2025-03-12
Add 4.3 to Current CI setup


deeplow:
## Description

Install 4.3 on the [current CI](https://github.com/freedomofpress/securedrop-workstation-ci/) to help flag the 4.3 compatibility status.

deeplow:
This is being worked on under https://github.com/freedomofpress/securedrop-workstation/pull/1373

PR #1373, opened 2025-07-15
Qubes 4.3 base compatibility

deeplow:
## Description of Changes

Towards #1245. Base Qubes 4.3 compatibility. Opening so we can iterate on OpenQA.

NOTES:
- Keeps Debian 12 templates -- even though debian 13 is officially installed in 4.3, [debian 12 will still be available](https://github.com/freedomofpress/securedrop-workstation/issues/1396#issuecomment-3455848215).
- Inplace upgrade from Qubes 4.2 is out of scope for this PR

## Testing

Requirements:
- Debian **13** dev env is needed after 5ce36cb (or equivalent after rebase)
- Machine running Qubes 4.3 (but feel free to test on 4.2 first)

Preparation:
1. Follow tests in 4.2 machine
2. Install Qubes 4.3 and create workstation build environment
3. ~~**Temporary:** until the fix https://github.com/QubesOS/qubes-issues/issues/10598 is in the stable repos, you need to enable `debian-12-minimal`'s testing repos:~~
   - [ ] download `debian-12-minimal` template manually
   - [ ] in `dom0`: `qvm-run -u root debian-12-minimal xterm` 
      - [ ] [enable testing repos](https://doc.qubes-os.org/en/latest/user/how-to-guides/how-to-install-software.html#debian) 
      - [ ] `sudo apt update && sudo apt upgrade`
      - [ ] shut down template 
4. `sd-dev`: 
   - [ ] Set correct poetry env: `poetry env use python3.13`
   - [ ] Checkout out this branch
5. `dom0`:
   - [x] Download and install dom0 keyring RPM (e.g. from 4.2 repos — it is not yet available in 4.3 repos)
   - [x] `make clone`
   - [x] `make dev`

Testing:
- [ ] Qubes 4.2: smoke testing workstation
- [x] Qubes 4.3
  - [x] `make dev` succeeds
  - [x] Opening document in disposable makes use of preloaded disposable (document opens within ~5 seconds as opposed to the usual longer delay )

## Deployment

2. New installs

### DIFF
--- a/scripts/prep-dev
+++ b/scripts/prep-dev
@@ -12,7 +12,6 @@ dom0_dev_dir="$HOME/securedrop-workstation"
 
 function find_latest_rpm() {
     # Look up which version of dom0 we're using.
-    # Qubes 4.2 is fedora-37, Qubes 4.3 fedora-41.
     fedora_version="$(rpm --eval '%{fedora}')"
     find "${dom0_dev_dir}/rpm-build/RPMS/" -type f -iname "*fc${fedora_version}.noarch.rpm" -print0 | xargs -r -0 ls -t | head -n 1
 }


legoktm:
Note that I've pulled out some of this into #1485.

.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     container:
-      image: quay.io/fedora/fedora:37
+      image: quay.io/fedora/fedora:41
```
deeplow: Do we even need to run "lint" in a container environment? If so, does it make sense to bump this to the dom0's Fedora version and add "matrix strategy"?

.github/workflows/openqa.yml
deeplow, line 1: I'm not sure how to handle OpenQA here. I've started a discussion in https://github.com/freedomofpress/securedrop-workstation/issues/1266#issuecomment-3570334067

deeplow:
(temporarily changing base to #1495, since that one is just in need of merging and can effectively be assumed as `main`)

deeplow:
**Update:** out standing tasks that need discussion on:
- We need to figure out the keyring logistics #1508 (for now I'm simply commenting this part out)
- How do we want to handle OpenQA? (see [discussion](https://github.com/freedomofpress/securedrop-workstation/pull/1373#discussion_r2555951194))

Without these we can't really merge against main without breaking the main branch 4.2.

deeplow:
The NEEDLES_DIR removal may actually be needed here. It's failing to find [an `x11` needle](https://openqa.qubes-os.org/tests/162695#).

tests/test_dom0_rpm_repo.py
deeplow, line 1: **NOTE:** This should be removed after https://github.com/freedomofpress/securedrop-workstation/pull/1533

deeplow:
The previous test was failing on [the keyring](https://openqa.qubes-os.org/tests/165460). It should go away once I rebase on top of https://github.com/freedomofpress/securedrop-workstation/pull/1533.

deeplow:
I'm rebasing and testing on 4.3 locally. I'll force-push once that's one.

files/sdw-admin.py
```diff
@@ -273,6 +277,17 @@ def perform_uninstall():
     )
 
 
+def is_managed(qube_name: str) -> bool:
+    """
+    Help assess if qube is to be managed directly
+
+    Currently excluded qubes:
+    - preloaded qubes: they are restarted when changes are
+    applied to tempaltes and do no need explicit management.
```
legoktm:
```suggestion
    applied to templates and do no need explicit management.
```

scripts/prep-dev
```diff
@@ -12,7 +12,7 @@ dom0_dev_dir="$HOME/securedrop-workstation"
 
 function find_latest_rpm() {
     # Look up which version of dom0 we're using.
-    # Qubes 4.1 is fedora-32, Qubes 4.2 fedora-37.
+    # Qubes 4.2 is fedora-37, Qubes 4.3 fedora-41.
```
legoktm: I would be okay dropping this comment, doesn't really serve a purpose other than being something that can get outdated

deeplow: Done in 75df1a7.

files/sdw-admin.py
deeplow: Done in 8627a29. This typo was also in `sdw-admin.py`.

PR #1370, opened 2025-07-14
Add workaround for SELinux fedora breakage

deeplow:
## Description of Changes
Fedora update was failing in certain templates. Out of abundance of caution, we're disabling SELinux (an acceptable risk) on a temporary basis. This allows the templates to successfully update until the original issue is fixed. Otherwise fedora-based VMs won't boot.

Fixes #1369

## Testing

Preparation
- [x] rename your fedora-41-xfce to something else
- [x] install breakage-prone fedora-41-xfce from provided backup (ask @deeplow)
- [x] shutdown and set `sys-firewall`, `sys-net` and `sys-usb` to `fedora-41-xfce` (the provided one)
- [x] confirm that SELinux is enabled for `fedora-41-xfce` via `qvm-features fedora-41-xfce selinux` (should print `1`) 

Testing:
- [ ] run `sdw-updater`
- [ ] Ensure runs successfully
- [x] `qvm-features fedora-41-xfce selinux` (should return nothing)
- [ ] Open client application and ensure it still connects to internet

(just to be extra sure):
- restart machine
- system starts just fine
- client connects to network

## Deployment

Any special considerations for deployment? Consider both:

1. Upgrading existing pilot instances
2. New installs (will be tested with CI)

### If you have made changes to the provisioning logic

- [ ] All tests (`make test`) pass in `dom0`

### If you have added or removed files

- [ ] I have updated `MANIFEST.in` and `rpm-build/SPECS/securedrop-workstation-dom0-config.spec`

### If documentation is required

- [ ] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-workstation-docs) for these changes, or will do so later
- [ ] I would appreciate help with the documentation

PR #1375, opened 2025-07-17
Reenable SELinux fedora (breakage fixed upstream)

deeplow:
**NOTE**: this should be rebased after https://github.com/freedomofpress/securedrop-workstation/pull/1372 is merged.

## Description of Changes
Reverts a workaround added in commit 9e088d for the [hotfix release 1.2.1](https://github.com/freedomofpress/securedrop-workstation/pull/1370). The original issue does not appear to have affected the workstation but this was information we did not have at the time of the release.

## Testing

Be on system with release 1.2.1  (dev or prod).

- confirm `qvm-features fedora-41-xfce selinux` should be <strike>`0`</strike> (**Update**: should be empty)
- from this branch run `make dev`
- confirm `qvm-features fedora-41-xfce selinux` should be <strike>`0`</strike> (**update** should be `1`)
- Dom0 tests pass (and especially `test_lsm_enabled` for `sys-usb` is NOT skipped)

## Deployment

2. New installs

### DIFF
--- a/rpm-build/SPECS/securedrop-workstation-dom0-config.spec
+++ b/rpm-build/SPECS/securedrop-workstation-dom0-config.spec
@@ -154,8 +154,8 @@ qubesctl saltutil.sync_all refresh=true -l quiet --out quiet > /dev/null || true
 qubesctl top.enable securedrop_salt.sd-workstation > /dev/null ||:
 
 # Force full run of all Salt states - uncomment in release branch
-mkdir -p /tmp/sdw-migrations
-touch /tmp/sdw-migrations/undo-fedora-selinux-fix
+# mkdir -p /tmp/sdw-migrations
+# touch /tmp/sdw-migrations/undo-fedora-selinux-fix
 
 # Enable service that conditionally removes our systemd-logind customizations
 # on dev machines only.


rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -154,8 +154,8 @@ qubesctl saltutil.sync_all refresh=true -l quiet --out quiet > /dev/null || true
 qubesctl top.enable securedrop_salt.sd-workstation > /dev/null ||:
 
 # Force full run of all Salt states - uncomment in release branch
-# mkdir -p /tmp/sdw-migrations
-# touch /tmp/sdw-migrations/fedora-selinux-fix
+mkdir -p /tmp/sdw-migrations
+touch /tmp/sdw-migrations/undo-fedora-selinux-fix
```
eloquence: With the benefit of a bit more time for this revert: Do we in fact need to perform a full `sdw-admin --apply` run to apply this change? The updater does apply the `dom0` state in every run:
https://github.com/freedomofpress/securedrop-workstation/blob/34663f87f1b498932db3b957aa77384e293203c1/sdw_updater/Updater.py#L437

And the `sd-sys-vms.sls` is included in the `dom0` states in `sd-workstation.top`:
https://github.com/freedomofpress/securedrop-workstation/blob/34663f87f1b498932db3b957aa77384e293203c1/securedrop_salt/sd-workstation.top#L8

deeplow: You're totally right. I keep on forgetting that the dom0 state is applied on updater runs. This was actually missed [during last release](https://github.com/freedomofpress/securedrop-workstation/commit/9e088dba5f4b30c5225b4b59877e8f12eeef461e), so we may have triggered an unnecessary migration.

Should be fixed in 2bbb9d4.

eloquence:
Awesome, thanks! I want to give this a full spin with an updater run, should be able to do so on Monday.

eloquence (changes requested):
Unfortunately, it is not yet working for me. I built an RPM from this branch, installed it, and forced an updater run. The updater run failed.

I've pasted the detailed logs in [this gist](https://gist.github.com/eloquence/76efb6b706ae2bcdfc2130570426d4e3), but the gist of the gist is that the change to Fedora 41's config triggers an update run, which itself errors out with `retcode` 26.

Attempting to run `sudo qubes-vm-update --force-update --targets fedora-41-xfce` in `dom0` also failed, and led to "failed to start" pop-ups from Qubes. Looking at the `fedora-41-xfce` logs I see .. lots of SELinux errors, e.g.:

```
[    2.203510] systemd[1]: Failed to set SELinux security context system_u:object_r:bpf_t:s0 for /sys/fs/bpf: Permission denied [2025-07-28 15:21:34] [.[0;1;31m!!!!!!.[0m] Failed to mount API filesystems.  [2025-07-28 15:21:34] [    2.262038] systemd[1]: Freezing execution. 
```

[Here's a gist](https://gist.github.com/eloquence/0032438b561b25cb4cae65054c01a838) with the full log.

After disabling SELinux again, the template boots up just fine.

Just re-enabling may not be enough to get things back into a working state. We may have to do a re-labeling operation, or apply other permission changes. Looking in the template. it looks like `/.autorelabel` exists, but if I understand how that works correctly, we never even get to the point where `systemd` could run it.

I can reproduce the issue quickly by just disabling and re-enabling SELinux on the affacted template without any of the code here.

(CC @marmarek in case he has any advice here.)

eloquence:
Here's the full log after re-enabling just now: https://gist.github.com/eloquence/dc6189768413f341fc53bbc892fb8243

marmarek:
So, there are a couple of things:
1. Looks like some important files got labels wrong (*). Maybe `/` directory (should have label `system_u:object_r:root_t:s0`, you can check with `ls -lZd /`)? I'm not sure how that happened by just running with SELinux disabled for a relatively short time,  but it isn't important right now.
2. SELinux policy gets loaded with enforcing enabled - that prevents booting such broken system to the point to attempt relabel. This can be worked around by adding `enforcing=0` to kernelopts.
3. With `enforcing=0`, the boot proceeds further and reaches `selinux-relabel.service`. But that's not the one, it needs to be `qubes-relabel-root.service`. And the latter one doesn't get started (at least for me). This seems to be related to `/.qubes-relabeled` being present. This behavior is a bug, but for now you can workaround it by removing that file.
4. When `qubes-relabel-root.service` actually starts, it barely fits (for me) in the default startup timeout (relabel takes about 50s). If it doesn't fit, retrying a few times should work eventually (checking if file has correct label is faster than setting it), but better increase the timeout for this operation.

So, all in all, it looks like re-enabling SELinux in this state is quite complicated...


(*) I'm not sure what specifically went wrong in your case, but I emulated similar behavior by removing _all_ labels...

eloquence:
Thanks much for the deep-dive, Marek. I haven't touched this particular machine much since the [release](https://securedrop.org/news/advisory-securedrop-workstation-1_2_1-released/) which disabled SELinux for the template. I have run the updater once or twice on that template (at least once today), which itself would have modified a few system files of course.

That's what I'd expect to be true for a lot of the SecureDrop Workstations in production use as well, so this may take a little bit to untangle ... I'll see if I can come up with a minimal sequence to unbreak the template tomorrow, but additional suggestions are very welcome!

eloquence:
> but given the good timing of the Fedora 42 release it could make sense to do a clean install of the new template.

This makes sense to me as a pragmatic path forward.

> At the same time, I'd argue that SELinux to be re-enabled in the old template

I don't really agree with that; why should we knowingly break the old template?

legoktm:
Agreed on just moving to Fedora 42 with SELinux enabled.

PR #1377, opened 2025-07-24
Simplify pull request template

eloquence:
Same principles as https://github.com/freedomofpress/securedrop/blob/develop/.github/PULL_REQUEST_TEMPLATE.md

- Use project board to track state
- "Description" is implied
- When a checklist is not completed, that should always convey useful information

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,35 +1,13 @@
-## Status
+Fixes #
 
-Ready for review / Work in progress
-
-## Description of Changes
-
-Fixes #.
-
-Changes proposed in this pull request:
-
-## Testing
-
-If you made non-trivial code changes, include a test plan and validated it for this PR.
-
-## Deployment
-
-Any special considerations for deployment? Consider both:
-
-1. Upgrading existing pilot instances
-2. New installs
+## Test plan
+<!-- Delete this section if not applicable (e.g., some docs-only changes) -->
 
 ## Checklist
 
-### If you have made changes to the provisioning logic
+<!-- If you leave any box below unchecked, please clarify where you may need support.
+     If you're unsure, that's fine — a reviewer can help you out. -->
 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,35 +1,13 @@
...
-1. Upgrading existing pilot instances
-2. New installs
+## Test plan
+<!-- Delete this section if not applicable (e.g., some docs-only changes) -->
 
 ## Checklist
 
-### If you have made changes to the provisioning logic
-
-- [ ] All tests (`make test`) pass in `dom0`
```
nathandyer, line 26: By removing this checkbox for tests, I assume the implication is that if all tests aren't passing then the PR wouldn't be ready for review in the first place - is that correct?

I think removing this is the right move, I just want to make sure we're not overlooking it.

nathandyer (approved):
Thanks @eloquence, this LGTM! Left one comment for clarification just to make sure we're not overlooking tests, but I think this is good to merge as-is.

Marking an approval, but not a maintainer so someone else will need to do the final OK and add to merge queue.

.github/PULL_REQUEST_TEMPLATE.md
eloquence: Yes; one related principle that Kunal suggested was that checklist items shouldn't duplicate stuff that CI takes care of for us. IIRC this checklist item pre-dates us having automated tests for `dom0`; now that we do, it seems reasonable to rely on them.

eloquence: (Of course, `dom0` tests are still a bit flaky in CI as the test run on this PR demonstrates, so this does assume that we'll need to be diligent when we ignore them vs. re-run them.)

nathandyer: Excellent, that principle makes my minimalist heart happy :)

zenmonkeykstop (approved):

Issue #1212, opened 2024-12-04
Proposal: build variant managed by boostrap package 


rocodes:
## Proposal:
Proposal with more detail on two separate rpm boostrap packages, securedrop-workstation-keyring{,staging} and how it will help us with cleaner encapsulation and separation of prod from non-prod content/logic.

After https://github.com/freedomofpress/securedrop-workstation/issues/1058#issuecomment-2515713456 I realized I needed to write stuff up more formally and it probably belongs in a separate issue/proposal instead of a ticket comment.  

### Affected components
SecureDrop Workstation

### People and roles
sd maintainers 

### Problem Statement
A few related problems:
- Workstation rpm ships with some non-prod artifacts (test key, yaml that includes prod and staging repos), as well as non prod logic (salt logic that includes conditional provisioning depending on dev, staging, prod) #1058  #945 
- Branching salt logic is error prone/hard to read/slows down provisioning  (various)
- Conditional logic in systemd services relying on `ConditionPathExists` and flag files/directories (/var/lib/securedrop-workstation/{dev,prod,staging}) is a bit messy and potentially brittle #1052  

### Solution impact
- don't ship non-prod stuff in prod packages
- easier to reason about your system 
- reduce time on salt runs

### Requirements and constraints
- Previous requirement/constraint: "don't want to maintain two different dom0 config RPMs" (this is reasonable, this was a comment from Kev when I suggested having a separate dev vs prod rpm for similar reasons). Constraint: securedrop-workstation-dom0-config is the same rpm across all build variants.
- Prod by default: constraint should be things should "fail closed/fail to prod configuration" (don't have a situation where a prod user could accidentally not have a service or system enabled; eg logind overrides). 
 
### Initial proposal
My proposal is as follows:

- securedrop-workstation-dom0-config depends on a boostrap package that provides repo config + signing key config. (So this would include apt-sources, not just yum.)
- securedrop-workstation-dom0-config stops shipping any non-prod logic. (More on this later - it's readily doable).
- securedrop-workstation-keyring, prod package, contains the .repo file, signing key, and apt-sources file(!) needed for a prod configuration.
- systemd services are configured via a preset file that is shipped in the dom0 rpm. (Random minor detail? totally. but relevant in a minute #1052 ). 
[...]

PR #1390, opened 2025-08-05
Remove individual VM provisioning targets from Makefile

### LINKED ISSUE #1378: Remove make targets that call qubesctl

## Description

There are a lot of make targets that we don't use anymore, and it's probably better to stick to a standardized provisioning strategy (ie "install rpm, run `apply`"), particularly since some configuration also happens in rpm  scriptlets.

Proposing we remove the per-VM make targets and standardize around above-mentioned install strategy.

## How will this impact SecureDrop/SecureDrop Workstation users
no user-facing changes

n/a

## User Stories
n/a -- developer-facing

rocodes:
Fixes #1378

- Remove `make $VM` targets from makefile - people should use standard provisioning (sdw-admin --apply)
- Remove duplicate validation step in prep-dev - prep-dev was also calling "validate", so we were validating twice. 
- Don't require prep-dev in the uninstall target, since it reinstalls an rpm. Just try to run the uninstall script. 
- Deprecate `validate` target in Makefile and reference the validation script directly in the dev/staging target, to avoid diverging validation logic.

Notes and future work:
- If/when we include apt repo info in the boostrap rpm, we can deprecate make dev/staging altogether, because the install workflow will always be the same - install bootstrap, install dom0 rpm, apply. See for reference https://github.com/freedomofpress/securedrop-workstation/issues/1212#issuecomment-3114807015 

## Test plan
- [ ] Visual review
- [ ] CI passes 
- [ ] make dev/staging passes (begins apply stage is OK)

## Checklist

This change accounts for:


rocodes:
Passing openQA, but failing sdci runner due to whonix clock sync (unrelated) - will mark as ready for review

Makefile
```diff
@@ -67,70 +67,10 @@ clone: assert-dom0 ## Builds rpm && pulls the latest repo from work VM to dom0
...
-	@./files/destroy-vm.py sd-devices-dvm
-
-remove-sd-log: assert-dom0 ## Destroys SD logging VM
-	@./files/destroy-vm.py sd-log
-
-clean: assert-dom0 prep-dev ## Destroys all SD VMs
+clean: assert-dom0 ## Destroys all SD VMs
 # Use the local script path, since system PATH location will be absent
 # if clean has already been run.
 	./scripts/sdw-admin.py --uninstall --force
```
deeplow, line 73: I was trying to do some git archaeology to understand why prep-dev was introduced in the 1st place (to know if safe to remove). It was added in commit https://github.com/freedomofpress/securedrop-workstation/commit/e540ea6c10ee9471692b0a0de4b8141c43d2a945, but no particular reason is given.

My best bet is that this would have helped in the development of uninstall scripts. Otherwise one has to do `make clone` (as usual), install the RPM and only then `make clean`. On the other hand with `make clean` one usually expects it to do nothing but to clean, which makes your approach feel more normal.

No strong feelings. Just wanted to add a note on this.

Makefile
```diff
@@ -24,8 +24,8 @@ all: assert-dom0
 
 dev staging: assert-dom0 ## Configures and builds a dev or staging environment
 	./scripts/configure-environment.py --env $@
-	$(MAKE) validate
-	$(MAKE) prep-dev
+	@./scripts/prep-dev
```
deeplow, line 27: I don't see any other reference to `prep-dev` now. Shall we nuke it as well? (unless anyone is using `make prep-dev` directly). TBH I like the direction you're going with all of these removals, which just made the tooling more complicated to use. So I'd be in favor of removing as well.

Makefile
rocodes: My guess is it was to ensure that the newest salt files were copied into place (prep dev would do that) because the uninstall logic relied on Salt / qubesctl commands.

We now use a python script and we also don't officially support uninstall this way (whereas before we considered doing so). Arguably one could still have the problem of not having the newest version of the python script in place before running the command, but I think we should probably maintain less and less of the dev-facing custom tooling and move towards predictable prodlike patterns, and I think it's a bit surprising when an _uninstall_ script installs a package :upside_down_face:

Makefile
rocodes: We can't nuke it here yet - it installs the rpm.  I do want to go in that direction though!  I think eventually we will be able to just have one simple install workflow, where this entire target disappears, and users either:
- install a keyring, `qubes-dom0-update` and grab the config rpm, then `apply` (nightly, staging, prod), OR
- install a keyring, build dom0 config rpm locally, install it (what prep-dev does), then run `apply` (local dev build)

rocodes: Yep - tis gone.  As an aside: we need to keep the logic here for now, but eventually I think we can deprecate this whole dev/staging target too, and just have two possible install paths:

- path 1: install a keyring, `sudo qubes-dom0-update securedrop-workstation-dom0-config`, `apply`: works for all sdw except local builds (nightlies, staging, prod), depending on the keyring 
- path 2: install a keyring, `sudo dnf install my-locally-build-dom0config-rpm`, `apply` (which is what `prep-dev` does now): works for a local dev sdw build

Makefile
deeplow: Yeah. No path is optimal, but your proposed one breaks less expectations. If someone wants to test that component, they'll soon enough figure out that things are not refreshing on `make clean` and look under the hood.

Let's close this one, then.

Makefile
deeplow: Thanks for the clarification. I'd be in for keeping the targets, actually. if anything because I think getting and installing the keyring is too long of a command to properly memorize. But let's weight in the options when we get to it.

Issue #1053, opened 2024-05-30
Use dev/staging pillars to remove branching logic in provisioning


rocodes:
* [x] I have searched for duplicates or related issues

## Description

Filing for discussion

Right now our provisioning logic copies everything needed for provisioning into one common Salt directory, and uses conditionals in .sls files to choose the right path(s) for provisioning, leading to lots of branching logic `{% if d.environment ==  "staging" %} ... {% elif d.environment == "prod" %}`. Part of why the provsioning is complex is because it's trying to deal with different build flavours/environments, and can't tell which one to use until orchestration/apply time. 

I think we could use saltenv/pillarenv and simplify our lives:
* define separate pillars (data stores) for staging, dev, or prod data, but keep salt states shared across all environments so no Salt duplication
* at provisioning time, specifying the environment when invoking salt command  (`qubesctl state.apply pillarenv=dev`)
** (use the `pillarenv_from_saltenv` [configuration option](https://docs.saltproject.io/en/latest/ref/configuration/minion.html#pillarenv-from-saltenv) for sanity)
* In each of dev, staging, or prod, we'd have the pillars (data) specific to the build environment, namely (`securedrop-workstation.repo`, `sd-release-key.asc`), meaning sd-default-config{.sls,.yml} wouldn't be needed.
* We could also simplify our conditional provisioning of systemd services (wip #1038) and be conditionally provisioned via the use of preset files that vary depending on build envrionment. (This could happen anyway: https://github.com/freedomofpress/securedrop-workstation/issues/1052)
* I think this structure will translate even when we move away from Salt. (ansible environments etc)
 
## How will this impact [SecureDrop/SecureDrop Workstation users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?
* The more logic we remove from dom0 salt states, the faster provisioning is, but more importantly, the faster every updater run is

## How would this affect the SecureDrop Workstation [threat model](https://github.com/freedomofpress/securedrop-workstation/#threat-model)?
* More reliance on systemd
* Maybe easier to reason about system state

cfm:
I support this approach wholeheartedly.  As you point out, @rocodes, it has precedent in how we parameterize `securedrop`'s Ansible playbooks.  It would be *wonderful* to be able to separate configuration from provisioning logic.

rocodes:
[...]

PR #1392, opened 2025-08-06
Configure apt environment based on bootstrap environment choice

### LINKED ISSUE #1058: Protect production users from setting up a dev or staging environment in error using production RPM
Currently the RPM includes the test signing key and repo config - setting up a dev or staging environment involves changing the env setting involves changing one  value in `config.json` and running `sdw-admin --apply`. This increases the risk that users may configure a test environment in error for a system intended for production.

There are a few options for guardrails we could put in place:

- [ ] https://github.com/freedomofpress/securedrop-workstation/issues/1053 
- [ ] Build separate RPMs for dev/staging and production, containing the correct keys in each case.
- [ ] dev and staging envs will have a copy of the source available in dom0 - we could overwrite the RPM-provided key and repo config in the `prep-dev` script invoked via `make {dev,staging}` prior to running `sdw-admin --apply`

We should discuss and pick one.

### LINKED ISSUE #1232: Decide on dev/staging provisioning strategy compatible with securedrop-workstation-keyring
(Converted to issue from the release checklist)
With securedrop-workstation-keyring providing the yum prod .repo file and release signing key, the dev/staging provisioning will need to be adjusted in order to maintain compatibility with existing make targets.

Ideally, the workstation-keyring package will be solely in charge of the file `securedrop-workstation.repo` in `/etc/yum.repos.d/`, and will not be dynamically provisioned with a staging repo config at provisioning time, since that could lead to confusing behaviour: RPM will check that the file is present, but will not overwrite its contents if it is, so the best case is for staging/dev to provision a file with a different name (eg securedrop-workstation-staging.repo), probably removing the prod .repo file so as not to confuse dom0 updates, and additionally importing the staging key into the rpm database.

rocodes:
Ready for a look, depends on #1210 [edit: merged!]

Fixes #1058 
Fixes #1232 
Refs #1053  (can supersede/close with another approach in a followup PR)

Use bootstrap rpm instead of config.json to decide what environment (dev, staging, prod) we're using. 
- Build apt .sources file dynamically, from a single j2 template (drop apt_test j2 template).   
- Always default to prod
- Remove pubkey and test key from this repo, and source them from the file managed by the bootstrap rpm. (One fewer place to lose track of pubkey)
- (chore) Clean up unused config.json imports in a couple of VM sls files

## Test plan
- [ ] base branch is `feat/sdw-keyring-compat` 
- [ ] #1210 is merged
- [ ] CI (openqa) passing (clean install)
- [ ] visual review 

## Checklist

This change accounts for:


rocodes:
(triggered openqa, will rebase once #1210 is merged then mark ready for review)

tests/test_dom0_and_templates_correct_key.py
```diff
@@ -0,0 +1,248 @@
...
+    ) as tmp_file:
+        tmp_file.write("\n".join(key_lines) + "\n")
+
+    return tmp_file.name
+
+
+### BEGIN TESTS ###
+
+# Get config, needed for test
+ENV, CONFIG = _get_expected_config()
```
legoktm: Let's turn these into pytest fixtures.

tests/test_dom0_and_templates_correct_key.py
```diff
@@ -0,0 +1,248 @@
...
+            exp = parts[6]
+            if exp:
+                key_info["expires"] = exp
+        elif tag == "uid":
+            key_info["uids"].append(parts[9])
+
+    return key_info
+
+
+def _validate_key(info: dict, expected_fpr: str, expected_uid: str, expected_expiry: str):
```
legoktm, line 68: instead of checking things and raising exceptions, I think we can just use normal pytest assertions

tests/test_dom0_and_templates_correct_key.py
```diff
@@ -0,0 +1,248 @@
...
+        CONFIG["signing_key_fpr"],
+        CONFIG["signing_key_uid"],
+        CONFIG["signing_key_exp"],
+    )
+
+
+def test_key_apt_sources():
+    """Check the key in the apt_sources file used for provisioning Debian-based templates."""
+    app = Qubes()
+    for template in ["small", "large"]:
```
legoktm: pytest parameterization?

tests/test_dom0_and_templates_correct_key.py
```diff
@@ -0,0 +1,248 @@
...
+
+        # Know which template we failed on
+        try:
+            _read_and_validate_key(
+                tmp_keyfile,
+                CONFIG["signing_key_fpr"],
+                CONFIG["signing_key_uid"],
+                CONFIG["signing_key_exp"],
+            )
+        except ValueError as e:
```
legoktm: Above I suggested using assertions in `_read_and_validate_key` so then we don't need to catch exception and then immediately fail. Plus if we use parameterization then the test name will also have the VM name.

securedrop_salt/sd-default-config.sls
```diff
@@ -2,33 +2,49 @@
...
+    {% set env = "dev" %}
+    {% break %}
+  {% elif pkg.split('-')[-1] == "staging" %}
+    {% set env = "staging" %}
+  {% endif %}
+{% endfor %}
 
-# Append repo URL with appropriate distribution
-{% set _ = sdvars.update({"distribution": "bookworm"}) %}
+{% set apt_config = environments.get(env) %}
```
legoktm, line 43: I'm wondering if we just leave this called `sdvars` even though it's different, just to cut down on the number of salt files modified in this PR that are entirely just `s/sdvars/apt_config/`.

securedrop_salt/sd-base-template.sls
```diff
@@ -19,9 +19,16 @@ sd-base-template:
...
-        - sd-{{ sdvars.distribution }}
+        - sd-{{ apt_config.distribution }}
     - features:
       - enable:
         - service.paxctld
     - require:
       - qvm: dom0-install-debian-minimal-template
+
+# Store signing key in salt cache so that it can later be provisioned to templates in their apt-sources files.
+# Using cp.push or cp.cache_file don't work due to the minion-minion setup.
```
legoktm: Sorry, this is going over my head. Why can't the step in fpf-apt-repo.sls just read the original `apt_config['keyfile']`?

tests/test_dom0_and_templates_correct_key.py
```diff
@@ -0,0 +1,248 @@
...
+        subprocess.check_call(
+            ["rpm", "-q", pkg], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL
+        )
+
+        return True
+    except subprocess.CalledProcessError:
+        return False
+
+
+# A bit clunky, but we have limited tooling options in dom0.
```
legoktm, line 28: Cursed but if it works, it works :) 

Could you add an explicit `TODO: use sequoia here instead` - I think we can do that once we're on 4.3.

legoktm (changes requested):
Overall it looks good, most of my comments are related to the tests.

securedrop_salt/sd-base-template.sls
rocodes: `apt_config['keyfile']` is a reference to the signing key in `/etc/pki` in dom0, and which one it is depends on the keyring package (test key or real release key).

The fpf-apt-repo.sls steps are run on the Debian VMs, but the bootstrap keyring files live in dom0. 
The debian vms have access to /srv/salt (as `salt://`) and to the cache dir but not to dom0 files, so somewhere in a state that's run on dom0 (I just picked this file for simplicity since then we only do it once, at initial provisioning time, and after that the key is managed via the keyring package), we cache the key, and then the VMs can read it.

securedrop_salt/sd-default-config.sls
rocodes: It's up to you, but I just found `sdvars` such a confusing/nondescriptive name that it seemed like this would be long-term easier to understand.

securedrop_salt/sd-base-template.sls
legoktm: Gotcha, appreciate the explanation. How about:

```
# Debian VMs need access to the signing key for initial provisioning, store it in salt cache since they
# only have access to the cache and `/srv/salt`, not all of dom0.
# Using cp.push or cp.cache_file don't work due to the minion-minion setup.
```

rocodes:
The CI failure shows that the installation is looking for the prod key, when it should be looking for the test key. Here's why: sd-default-config.sls file correctly selects the apt configuration we want when it's run on dom0, but even when it's imported `with context` in the VM files, it still tries to run the same checks on the vm, as opposed to remembering the configuration we selected. 

The solution is to write the configuration to the same cachedir we write the key. (Note: this is also where Salt files are cached when they're being applied; we aren't adding directories or files anywhere new).

I'm now just debugging why Salt can read the key file but not the json config file that I wrote to that directory, when all the permissions and file owners match what's already in there.

PR #1405, opened 2025-08-13
Create boostrap make target for keyring installation

rocodes:
Developer-facing changes:

Configure .repo file and download keyring bootstrap (dev and staging only) via bootstrap-keyring.py script and `make` target.  Note: bootstrap-keyring.py is based on an AI-generated Python script which I then lightly refactored.

This also puts a copy of the test key in the `scripts/` directory. The test key will stop being included in the built rpm (see https://github.com/freedomofpress/securedrop-workstation/pull/1210) but this PR can be reviewed independently.

## Test plan
- [ ] visual review
- [x] base branch is `feat/sdw-keyring-compat`
- [x] `make bootstrap-dev` and `make bootstrap-staging` work, and install respective keyring packages in a Qubes environment (this is for testing locally on Qubes or in openQA). (Note that you still need to install the dom0 config rpm package yourself afterwards with `sudo qubes-dom0-update securedrop-workstation-dom0-config`)
- [x] `make boostrap-anythingelse` fails with instructive error message

## Checklist

This change accounts for:

### DIFF
--- a/.github/workflows/openqa.yml
+++ b/.github/workflows/openqa.yml
@@ -42,7 +42,7 @@ jobs:
             GIT_REF="$GIT_REF" \
             FLAVOR=securedrop \
             NEEDLES_DIR="%%CASEDIR%%/needles" \
-            CASEDIR="https://github.com/QubesOS/openqa-tests-qubesos.git#main"\
+            CASEDIR="https://github.com/rocodes/openqa-tests-qubesos.git#securedrop-keyring-bootstrap"\
             MAX_JOB_TIME=10800 | tee openqa.json
 
           for id in $(jq -r '.ids[]' openqa.json); do


scripts/bootstrap-keyring.py
```diff
@@ -0,0 +1,91 @@
...
+        repo_file_path = Path(temp_dir) / f"{KEYRING_PACKAGENAME}-{args.env}.repo"
+
+        create_repo_file(args.env, repo_file_path, fedora_version)
+        repo_dest_path = Path(YUM_REPOS_DIR) / f"{KEYRING_PACKAGENAME}-{args.env}.repo"
+        subprocess.check_call(
+            ["sudo", "install", "-m", "0644", str(repo_file_path), str(repo_dest_path)]
+        )
+
+    # Install keyring package
+    rpm_import(TEST_KEY_FILENAME)
```
deeplow: You'll figure this out right away and it's not ready for review, but since I stumbled upon it while running the OpenQA test...

```suggestion
    rpm_import(key_file)
```

Feel free to close.

scripts/bootstrap-keyring.py
```diff
@@ -0,0 +1,95 @@
...
+
+        create_repo_file(args.env, repo_file_path, fedora_version)
+        repo_dest_path = Path(YUM_REPOS_DIR) / f"{KEYRING_PACKAGENAME}-{args.env}.repo"
+        subprocess.check_call(
+            ["sudo", "install", "-m", "0644", str(repo_file_path), str(repo_dest_path)]
+        )
+
+    # Install keyring package
+    rpm_import(key_file)
+    dom0_install_keyring(args.env)
```
deeplow, line 102: Do we want to add a delay of 15s for the key to be installed? Ortherwise, the user (or OpenQA)

Or perhaps instead checking if the key is already installed and if not, looking until it is. Something like this:
```
rpm -qi gpg-pubkey-* | gpg --import --import-options show-only --dry-run - | grep "83127F68BABB04F3FE9A69AA545E94503FAB65AB"
```

rocodes: addressed in https://github.com/freedomofpress/securedrop-workstation/pull/1405/commits/b83f3429ccf89be5ac19b7e2a087a7958e60ba7a

scripts/bootstrap-keyring.py
```diff
@@ -0,0 +1,114 @@
...
+TEST_KEY_RPMID = "gpg-pubkey-3fab65ab-660f2beb"
+
+# Only support test rpms; for prod, use official install instructions
+BASEURL = "https://yum-test.securedrop.org"
+
+
+def get_fedora_version():
+    try:
+        return subprocess.check_output(["rpm", "--eval", "%{fedora}"]).decode().strip()
+    except subprocess.CalledProcessError as e:
```
legoktm: nit: since this is a dev/CI-facing script I don't think catching the exception and then immediately exiting is useful, the traceback is probably more useful for when something does go wrong. (I've found Claude tends to over do this, not sure if that's what happened here)

scripts/bootstrap-keyring.py
```diff
@@ -0,0 +1,114 @@
...
+        print(f"Error retrieving Fedora version: {e}")
+        sys.exit(1)
+
+    print("No Fedora version number found.")
```
legoktm: This is unreachable right?

scripts/bootstrap-keyring.py
```diff
@@ -0,0 +1,114 @@
...
+    """Create .repo file based on environment."""
+    repo_content = f"""
+[{KEYRING_PACKAGENAME}-{env}]
+enabled=0
+gpgcheck=1
+baseurl={BASEURL}/workstation/dom0/f{ver}{'-nightlies' if env == 'dev' else ''}
+name=SecureDrop Workstation Keyring ({env})
+"""
+    with open(repo_file_path, "w") as repo_file:
+        repo_file.write(repo_content.strip())
```
legoktm: nit: maybe lstrip() so there's still a trailing newline?

scripts/bootstrap-keyring.py
```diff
@@ -0,0 +1,114 @@
...
+    script_dir = Path(__file__).resolve().parent
+    key_file = script_dir / TEST_KEY_FILENAME
+    if not key_file.exists():
+        print(f"'{key_file}' not found.")
+        sys.exit(1)
+
+    fedora_version = get_fedora_version()
+
+    # Build and install .repo file
+    with tempfile.TemporaryDirectory() as temp_dir:
```
legoktm, line 82: Can you add a comment explaining why we can't just write the file directly? I *think* it's because we're running as the local user and then need to shell out to sudo to install it?

scripts/bootstrap-keyring.py
legoktm: Maybe we bump it to 17 or 20 seconds? Just because the keyring's unit only starts after 15 seconds, so if it doesn't start/finish immediately, there's still a bit of a race right?

scripts/bootstrap-keyring.py
rocodes: nope, this was me :upside_down_face: I can let it bubble up

scripts/bootstrap-keyring.py
rocodes: Sorry, yes - I had a different check previously that parsed /etc/os-release and set the fedora  version conditional on the qubes version; have fixed this up.

scripts/bootstrap-keyring.py
```diff
@@ -0,0 +1,106 @@
...
+
+# Test key: ID is deterministic and based on fpr + key creation timestamp
+TEST_KEY_RPMID = "gpg-pubkey-3fab65ab-660f2beb"
+
+# Only support test rpms; for prod, use official install instructions
+BASEURL = "https://yum-test.securedrop.org"
+
+
+def get_fedora_version():
+    return subprocess.check_output(["rpm", "--eval", "%{fedora}"])
```
legoktm: I think we still need `text=True` or `.decode()` here, it's bytes being stringified, so https://openqa.qubes-os.org/tests/149692#step/install_workstation/26 failed with a URL of `/fb'37n'-nightlies` (not sure where the "n" is coming from...)

legoktm: The n is from `\n`, so probably also a `.strip()`

rocodes: ugh, sorry to let that sneak in. one moment.

rocodes: okay. The perils of my back and forth between dom0 workflow. addressed (and squashed) in https://github.com/freedomofpress/securedrop-workstation/pull/1405/commits/82a50b5a91835fad68a6ae394816735a36c28125

.github/workflows/openqa.yml
```diff
@@ -42,7 +42,7 @@ jobs:
             GIT_REF="$GIT_REF" \
             FLAVOR=securedrop \
             NEEDLES_DIR="%%CASEDIR%%/needles" \
-            CASEDIR="https://github.com/QubesOS/openqa-tests-qubesos.git#main"\
+            CASEDIR="https://github.com/rocodes/openqa-tests-qubesos.git#securedrop-keyring-bootstrap"\
```
deeplow, line 45: Actually, we need to get this upstreamed prior to merging. I'd advocate again for keeping our own fork of https://github.com/QubesOS/openqa-tests-qubesos. This way we can better coordinate regarding the timings of these.

deeplow: Basically we have to synchronize the upstreaming of https://github.com/QubesOS/openqa-tests-qubesos/compare/main...deeplow:openqa-tests-qubesos:securedrop-keyring-bootstrap with this PR. Otherwise, the nightly job will fail.

deeplow: OK. Never mind. I see the strategy here now. We don't want this in the main testing environment, but rather in this feature branch alone. So we just have to be careful not to merge this into main when we merge the feature branch itself.

rocodes: Yep - that's what we discussed in our meeting yesterday :) (reverting just this commit)

rocodes:
merged per @legoktm yesterday (in DM); it's a feature branch so we are ok bypassing the SDCI runner in favour of openqa.

legoktm:
> I was going to merge this, since OpenQA passed, but merging seems to be blocked. I thought it was my stale review, but after dismissing it, I still see:

It's an artifact of our branch protection rules, because we're trying to merge into a feature branch instead of main. There's a rule that when you git push to update the branch, it automatically bypasses the "cannot update protected ref" whereas in a PR, you have to explicitly use the scary red override.

For now it's fine to use so we can land stuff, just bad UX from GitHub.

PR #1414, opened 2025-08-20
Prepares path for `sd-whonix` whonix removal

deeplow:
Towards https://github.com/freedomofpress/securedrop-workstation/issues/456. Adds minimal dependencies to test its client counterpart https://github.com/freedomofpress/securedrop-client/pull/2561. Whonix removal will be in a separate PR.

## Test plan

Apply with `sdw-admin --apply` after installing RPM and then see https://github.com/freedomofpress/securedrop-workstation/issues/456.

## Checklist

This change accounts for:
- [ ] any necessary RPM packaging updates (e.g., added/removed files, see `MANIFEST.in` and `rpm-build/SPECS/securedrop-workstation-dom0-config.spec`)
- [ ] any required documentation

PR #1415, opened 2025-08-22
Remove sd-whonix and whonix-gateway FPF packages and repos

deeplow:
Follow-up to https://github.com/freedomofpress/securedrop-workstation/pull/1414 (to be rebased after it is merged).
- Removes `sd-whonix` and tries to remove all code tries between the workstation and its whonix dependency.
- Added unofficial support for an installation where Whonix is not installed

TODO:
- [ ] update threat model in `README`
- [x] remove `sd-sys-whonix-vms.sls` (including removing apparmor kernel arguments)
- [ ] Conditional removal of packages from whonix templates, depending on presence of whonix VMs

## Test plan

- [ ] visual review
- [ ] Ensure no other lingering references to Whonix exist
- [ ] ensure FPF packages, repositories and keyring absent from sys-whonix or whonix-gateway-17
- [ ] `make dev` does not run `qvm.anon-whonix` or other system default formulas

## Checklist

This change accounts for:


.github/workflows/openqa.yml
```diff
@@ -42,7 +42,7 @@ jobs:
             GIT_REF="$GIT_REF" \
             FLAVOR=securedrop \
             NEEDLES_DIR="%%CASEDIR%%/needles" \
-            CASEDIR="https://github.com/freedomofpress/openqa-tests-qubesos.git#main"\
+            CASEDIR="https://github.com/freedomofpress/openqa-tests-qubesos.git#sd-whonix-removal"\
```
deeplow: This must be removed before merging.

securedrop_salt/sd-remove-deprecated-qubes.sls
```diff
@@ -5,25 +5,25 @@
 # WARNING: only remove when complete reinstall is assumed (e.g. 1.0.0 release)
```
legoktm: Did you mean to delete this file?

tests/test_vms_exist.py
```diff
@@ -92,23 +91,37 @@ def test_default_dispvm(self):
...
-        assert vm.provides_network
-        assert vm.autostart
-        assert not vm.template_for_dispvms
-        assert "sd-workstation" in vm.tags
+    def test_sd_whonix_absent(self):
+        """
+        The sd-whonix once existed to proxy sd-proxy's traffic throgh Tor.
+        But we've since removed it and included a Tor proxy in sd-proxy.
+        """
+        with pytest.raises(KeyError):
```
legoktm: nit: A simpler way to write this might be `assert "sd-whonix" not in self.app.domains`

tests/test_vms_exist.py
```diff
@@ -92,23 +91,37 @@ def test_default_dispvm(self):
...
+            "whonix-workstation-17",
+            "whonix-gateway-17",
+            "sys-whonix",
+            "anon-whonix",
+            "whonix-workstation-17-dvm",
+        ]
+        for qube_name in whonix_qubes:
+            try:
+                qube = self.app.domains[qube_name]
+                qube.property_is_default("kernelopts")
```
legoktm: I assume you meant:
```suggestion
                assert qube.property_is_default("kernelopts")
```

tests/test_vms_exist.py
```diff
@@ -92,23 +91,37 @@ def test_default_dispvm(self):
...
+            except KeyError:  # qube does not exist when accessing 'app.domains'
+                # Notify developer when this test may be OK to go away.
+                pytest.fail(
```
legoktm: Hmm, I don't think this is useful, like, we should be able to have people test the release that includes without having whonix installed. Can we just continue/skip instead?

tests/test_vms_exist.py
```diff
@@ -92,23 +91,37 @@ def test_default_dispvm(self):
...
+        """
+        Whonix templates used to be modified by the workstation (<=1.4.0).
+        Ensure they were properly reset.
+        """
 
-        assert vm.template.features.get("os-version") == CURRENT_WHONIX_VERSION
+        whonix_qubes = [
+            "whonix-workstation-17",
+            "whonix-gateway-17",
+            "sys-whonix",
```
legoktm, line 109: I don't remember what we did in the past, but did we apply the change to all the individual VMs or do we only need to check the templates for this?

securedrop_salt/sd-remove-whonix-packages.sls
```diff
@@ -0,0 +1,18 @@
...
+#  Workstations miss some updates (due to not being in use for a while), the
+#  mere presence of the repo/package, does not affect regular system behavior.
+##
+
+# NOTE: 'cmd.run' necessary since 'pkg.del_repo' or 'pkg.purge' fail due to:
+#    "The pkg module could not be loaded: unsupported OS family"
+sd-cleanup-whonix-gateway:
+  cmd.run:
+    - names:
+      - "sudo apt purge --yes securedrop-keyring securedrop-qubesdb-tools securedrop-whonix-config ||:"
```
legoktm: `apt` isn't supposed to be used in scripts, so:

```suggestion
      - "sudo apt-get purge --yes securedrop-keyring securedrop-qubesdb-tools securedrop-whonix-config ||:"
```

securedrop_salt/sd-remove-deprecated-qubes.sls
deeplow: No, that wasn't my intention. I'm treating this file as a slightly more sustainable way to remove qubes we used in the past but should no longer be present. Let me know if you need more clarifications.

tests/test_vms_exist.py
deeplow: I checked for all past whonix modifications in `sd-sys-whonix-vms.sls`. We only applied in the templates, but I've included all the "descendent" qubes just for extra checking. Qubes should in fact propagate the kernelopts from the template. Otherwise it would have not been doing what we wanted it to do.

Since it doesn't take much more time in the test I included it, just an an extra checking.

securedrop_salt/sd-remove-whonix-packages.sls
deeplow: Good catch. I had taken this from `clean-all.sls` (which is not super well supported) forgot to check this. Thanks!

deeplow: Done in 3151ab5

tests/test_vms_exist.py
deeplow: I keep on forgetting the whonixless install case. We don't yet know if this is something that we want to suggest or not. But my inclination is only have one install story (which can be whonixless). But if we go that route, we need to change the install docs.

That works. Done in 68fd305.

tests/test_vms_exist.py
deeplow: :bangbang: Fixed in bbe9279

tests/test_vms_exist.py
deeplow: 87098b

securedrop_salt/sd-remove-deprecated-qubes.sls
legoktm: well AFAICT securedrop_salt/sd-remove-unused-qubes.sls and securedrop_salt/sd-remove-deprecated-qubes.sls are identical right? do we need both?

tests/test_vms_exist.py
legoktm: Sounds good.

files/sdw-admin.py
```diff
@@ -111,11 +109,11 @@ def provision_and_configure():
     provision("Provisioning base template", "securedrop_salt.sd-base-template")
     configure("Configuring base template", ["sd-base-bookworm-template"])
     provision_all()
-    configure("Enabling Whonix customizations", [f"whonix-gateway-{Util.get_whonix_version()}"])
     configure(
         "Configure all SecureDrop Workstation VMs with service-specific configs",
         [q.name for q in Qubes().domains if "sd-workstation" in q.tags],
     )
+    configure("Disabling Whonix customizations", ["whonix-gateway-17"])
```
legoktm: I still see a failure on a whonixless install, and I think this is the culprit, because configure() does `run_cmd(["qvm-shutdown", "--wait", "--"] + targets)`, which fails with a non-zero exit because no such VM/template exists.

Maybe we can have configure learn some kind of `skip_if_missing` where it doesn't do anything if the target(s) don't exist?

(fwiw I commented out this line and then the install worked fine all the way through so this should be the last/only thing to fix)

legoktm:
> I have included logic to remove it in `sd-remove-whonix-packages.sls` on an updater run. Are you suggesting we also remove that logic?

No, I got confused and thought it was being installed and it isn't.

securedrop_salt/sd-remove-deprecated-qubes.sls
deeplow: Outch. You're right. I had not realized that's what you meant. Fixed in 109a910

files/sdw-admin.py
deeplow: I think this is kind off an edge-case that is going away pretty soon IMO. So my preference would be to just add an if before we go into that configure step. Once it goes away we just remove it. Otherwise we're adding logic to `configure()` to handle this specific situation, which we're unlikely to encounter in the future.

legoktm: Works for me!

deeplow: Because in general we do want the targets to exist, otherwise it may be doing something we don't want it to. But perhaps that's more on the testing territory.

deeplow: Done in 4952035. Thanks for testing the code ahead of time :)

legoktm (approved):
LGTM, my whonixless fresh install worked :) ty!

PR #1439, opened 2025-09-25
Update nightlies GHA workflow to use more narrowly-scoped GH app

adaFPF:
Relates to: https://github.com/freedomofpress/infrastructure/issues/5789

## Test plan
Successful push of nightly build to securedrop-yum-test on next cronjob.

## Checklist

This change accounts for:


.github/workflows/nightlies.yml
```diff
@@ -46,29 +46,41 @@ jobs:
...
         run: |
           git config --global user.email "securedrop@freedom.press"
-          git config --global user.name "sdcibot"
+          git config --global user.name "sdcibot-nightlies[bot]"
           cd securedrop-yum-test
           mkdir -p workstation/dom0/f37-nightlies
           cp -v ../rpm-build/*.rpm workstation/dom0/f37-nightlies/
           git add .
           git diff-index --quiet HEAD || git commit -m "Automated SecureDrop workstation build"
+          git remote set-url origin https://x-access-token:${GH_TOKEN}@github.com/${TARGET_REPO}.git
```
legoktm: This means it gets saved in `.git/config`, which is why we normally set persist-credentials: false to avoid that...could we do:

`git push https://x-access-token:${GH_TOKEN}@github.com/${TARGET_REPO}.git main` instead so it isn't ever persisted?

adaFPF: Extremely good point.  I'm so used to the "set remote -> push" pattern that I missed an obvious improvement in the zone I was already aiming for by limiting the envvar scope in the first place.  I'll amend, 2 shakes.

adaFPF: As a side note, a bunch of the workflows in the repos I was working with don't explicitly set `persist-credentials: false` for the checkout action.  Mostly it's a DiD superfluous improvement (the targets are public and don't invoke secrets in the first place), and so I mostly only changed it on the workflows I was working with, but I can make that sort of thing a rolling target as I go.

legoktm: I would assume that list is basically wherever we don't have zizmor set up yet: https://github.com/freedomofpress/securedrop-dev/issues/18

legoktm (approved):
LGTM, thank you! (I just squashed it down to one commit)

PR #1440, opened 2025-09-29
Simplify and improve testing of apt sources

### LINKED ISSUE #1438: test_sd_vm_apt_sources test is reliably flaky in CI
## Description

This test has been reliably flaky in CI (both SDW CI and OpenQA) ever since 44ed477b6626477bb11411949dad768e74dd1c8a which switched it from parsing files to looking at apt's state. We should go back to the earlier file-parsing approach.

legoktm:
For reasons that aren't super clear, the apt-get indextargets approach has been flaky on both SDW CI and now OpenQA.

Simplify the test by just looking at the file we provision and not the apt cache. AFAIK this part has always been relatively robust so it shouldn't be flaky anymore.

Also take advantage of pytest's parameterization.

Fixes #1438

## Checklist

This change accounts for:


tests/test_vms_platform.py
```diff
@@ -192,25 +161,41 @@ def test_sd_vm_apt_sources(self):
...
+                components.append("nightlies")
+
+            self.assert_apt_source(
+                vm,
+                " ".join(components),
+                "https://apt-test.freedom.press",
+                "/etc/apt/sources.list.d/apt-test_freedom_press.sources",
+            )
+
+    def assert_apt_source(self, vm, component, url, filename):
```
legoktm, line 193: This is a refactoring of _validate_apt_sources.

tests/test_vms_platform.py
```diff
@@ -192,25 +161,41 @@ def test_sd_vm_apt_sources(self):
...
-                for component in test_components:
-                    assert (
-                        f"Description: https://apt-test.freedom.press bookworm/{component} "
-                        + "amd64 Packages\n"
-                        in contents
-                    )
+                # No test sources should be present
+                stdout, stderr = vm.run("ls /etc/apt/sources.list.d/")
+                contents = stdout.decode("utf-8").rstrip("\n")
+                assert "apt-test_freedom_press.sources" not in contents
```
deeplow: Nit: can we differentiate these "contents" with a more specific variable name? Perhaps `sources_list`.

legoktm: Done.

Issue #1426, opened 2025-09-10
dev/staging environments: consistent and clear set up


deeplow:
## Description

There are many GH issues regarding development environments, each focuses on particular bits but it would be good to design a solution that tackles the most pressing. This issue has that goal:

1. Evaluate the various needs and come up with a solution
2. Update development docs to reflect changes

Bugs as a result:
- https://github.com/freedomofpress/securedrop-workstation/issues/1055
- https://github.com/freedomofpress/securedrop-workstation/issues/1135

Possibly beyond the scope but good to consider:
- https://github.com/freedomofpress/securedrop-workstation/issues/1066

legoktm:
@deeplow now that we have switch-apt-source, is there more you'd like us to do here?

PR #1449, opened 2025-10-02
Add switch-apt-source script to flip between APT repos

legoktm:
A mostly independent Python script that switches our templates' APT repos between dev, staging, and prod. It does nothing else, so you can switch e.g. from prod to staging, and then run the updater to simulate what a real upgrade would look like.

Refs #1426.

## Checklist

This change accounts for:


zenmonkeykstop (approved):
Script works as expected, LGTM

PR #1450, opened 2025-10-03
Merge keyring feature branch

### LINKED ISSUE #1442: Merge keyring feature branch into main
We're ready to merge the `feat/sdw-keyring-compat` branch into `main` now. Some conflicts might need to be resolved, and we'll also have to port over some OpenQA modifications too.

legoktm:
Fixes #1442.

## Test plan

TK

## Checklist

This change accounts for:


.github/workflows/openqa.yml
deeplow, line 1: We must remove these changes before merging. In place of this, we need to merge https://github.com/freedomofpress/openqa-tests-qubesos/pull/5/.

legoktm:
> but the reverse path is not entirely clear. But I don't think a solution needs be in scope, but we need to document this for QA of the respective release.

I don't think there can ever be a fully guaranteed path, aside from uninstall and reinstall.

Makefile
```diff
@@ -25,10 +25,43 @@ all: assert-dom0
 
 dev staging: assert-dom0 ## Configures and builds a dev or staging environment
 	./scripts/configure-environment.py --env $@
+	$(MAKE) assert-keyring-%
 	@./scripts/prep-dev
```
zenmonkeykstop: `prep-dev` tries to install the securedrop-workstation-dom0-config package, already installed by the bootstrap-<env> make target. It will fail if the config package has dependencies, as it uses dnf, which cannot resolve them in dom0.

Makefile
```diff
@@ -25,10 +25,43 @@ all: assert-dom0
 
 dev staging: assert-dom0 ## Configures and builds a dev or staging environment
 	./scripts/configure-environment.py --env $@
+	$(MAKE) assert-keyring-%
```
zenmonkeykstop: The `%` wildcard doesn't get resolved properly here - it should likely be `$@`

Makefile
zenmonkeykstop: Actually it won't *consistently* fail, as if grub2-xen-pvh is still in local cache it *will* get installed.

deeplow: There seems to be problem here: `dev` and `staging` are both installing a local RPM due to `prep-dev`. This likely needs to be changed to

```suggestion
	$(MAKE) install-rpm RPM_INSTALL_STRATEGY=$@
```

Makefile
```diff
@@ -25,10 +25,43 @@ all: assert-dom0
 
 dev staging: assert-dom0 ## Configures and builds a dev or staging environment
 	./scripts/configure-environment.py --env $@
+	$(MAKE) assert-keyring-$@
```
deeplow, line 32: Nit: this could be the first thing running, since we probably don't want any environment changes before we even know that we're using the right keyring. Maybe even a prerequite (but doing it as that would require more changes since `$@` means nothing in the prerequisites).

Makefile
deeplow: Yup. I had also noticed this and fixed it in https://github.com/freedomofpress/securedrop-workstation/pull/1450/commits/e8b481e0713c82e9b36af34584995e98e6b70001.

PR #1466, opened 2025-10-31
Suppress GPG prompt

### LINKED ISSUE #960: Split-GPG: consider making GPG keys access never expire
The following prompt appears when sd-app is first start:

gpg_key

If the user clicks <kbd>No</kbd> this could lead for an unusable workstation (at least until sd-app restart) and I think it doesn't add anything from the security perspective since the user is always expected to say <kbd>Yes</kbd>.

This was set to 8 hours [in 2018](https://github.com/freedomofpress/securedrop-workstation/commit/7cbdb86c47a76fdf09dea8d3c6d826e5b533ee18) and has stayed like that since then, my suggestion would be to reconsider if this choice still makes sense from the UX and security perspectives.

deeplow:
Open a fix, per discussion with @legoktm. This suppresses the following prompt that showed up on first start and then every now and then:

gpg_key

Prompt was an unnecessary extra UX hoop as there was no reason for denying the prompt (after all, the user always wants sd-app to be able to have its submissions decrypted). The qrexec policy is a more effective counter-measure, which is already in place: if it exists, the GPG decryption is allowed. If it's not there, then it's not allowed.

Suppression is done by setting the max linux epoch value (2**31-1). This works because When the qubes.Gpg RPC service is called, it compares the expiry date with the current time.

Fixes #960

(test on: openqa)

[1]: https://github.com/QubesOS/qubes-app-linux-split-gpg/blob/bf8d6b38b58ad63bada2fce3c78a56aec2a421f6/qubes.Gpg.service#L32

## Test plan

- upgrade path
  - [x] make dev (over existing dev install)
  - [x] check that sd-gpg does not show any prompt
- "clean install"
  - [x] `qvm-remove sd-gpg`
  - [x] repeat "upgrade path" plan

## Checklist

This change accounts for:

### DIFF
--- a/securedrop_salt/sd-gpg-files.sls
+++ b/securedrop_salt/sd-gpg-files.sls
@@ -16,7 +16,8 @@ sd-gpg-increase-keyring-access-timeout:
     - marker_start: "### BEGIN securedrop-workstation ###"
     - marker_end: "### END securedrop-workstation ###"
     - content: |
-        export QUBES_GPG_AUTOACCEPT=2147483647 # hides GPG prompt (max epoch)
+        # hides GPG prompt (max epoch)
+        export QUBES_GPG_AUTOACCEPT=2147483647
 
 sd-gpg-create-keyring-directory:
   file.directory:
--- a/tests/test_gpg.py
+++ b/tests/test_gpg.py
@@ -63,7 +63,7 @@ def vm_fingerprint(qube):
 
 
 def test_sd_gpg_timeout(qube):
-    line = "export QUBES_GPG_AUTOACCEPT=28800"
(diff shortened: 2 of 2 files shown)


deeplow:
Looping the support in, just to flag support implications. I don't think this is documented anywhere, so it doesn't need removal. But I think this is a welcome thing.

nathandyer:
Just adding my +1 here from a support perspective, I don't see any downsides to this

legoktm (changes requested):
This is beautiful, please update the test and then this is ready to go.

P.S. It would be nice if qubes let us configure this via QubesDB vm-config so we don't need to write to `.profile` like this...

PR #1485, opened 2025-11-12
Start building Qubes 4.3 (Fedora 41) nightlies

### LINKED ISSUE #1476: Publish Qubes 4.3 nightly packages
Good to have for 4.3 testing and necessary for testing 4.2 -> 4.3 upgrade.

legoktm:
Provides the base infrastructure needed to build the RPM on Fedora 41 and push it to yum-test, with no attention paid to whether that RPM works or not.

Fixes #1476.

## Test plan

* [x] Review https://github.com/freedomofpress/securedrop-workstation/actions/runs/19312766805/job/55237007016, see that it created both f37 and f41 packages and would've pushed it to yum-test (had I not commented that part out)

Note that once this is ready to merge, we'll need infra to update the "Required" CI levels to use the new job names.

## Checklist

This change accounts for:

### DIFF
--- a/scripts/container.sh
+++ b/scripts/container.sh
@@ -7,14 +7,13 @@ set -eu
 
 source "$(dirname "$0")/common.sh"
 
+OCI_BUILD_ARGUMENTS="${OCI_BUILD_ARGUMENTS:-} --build-arg=FEDORA_VERSION=${FEDORA_VERSION}"
+
 # Whenever we're not on the platform we expect, explicitly tell the container
 # runtime what platform we need
-# This is also relevant for CI, as the --platform argument is only available
-# from Docker API 1.32 onward, while at the time of writing CircleCI still on
-# API 1.23
 if [[ "$(uname -sm)" != "Linux x86_64" ]]; then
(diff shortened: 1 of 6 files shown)


scripts/container.sh
```diff
@@ -15,8 +15,12 @@ source "$(dirname "$0")/common.sh"
 if [[ "$(uname -sm)" != "Linux x86_64" ]]; then
     OCI_RUN_ARGUMENTS="${OCI_RUN_ARGUMENTS} --platform linux/amd64"
     OCI_BUILD_ARGUMENTS="${OCI_RUN_ARGUMENTS} --platform linux/amd64"
+else
+    OCI_BUILD_ARGUMENTS=""
```
deeplow: Can you explain a bit this one? Do we need to reset `OCI_BUILD_ARGUMENTS` when running on Linux, where in the past we didn't?

deeplow: Maybe something the following if we want to make sure it exists at this point:

```suggestion
else
    OCI_BUILD_ARGUMENTS="${OCI_BUILD_ARGUMENTS:-}"
```

legoktm: Hmm, there's a typo in the existing code. Let me fix...

PR #1487, opened 2025-11-18
 Partial pytest conversion to centralize list of present VMs

deeplow:
Partially addresses #1338 by the logic around obtaining all qubes managed by the workstation (`sdw_tagged_vms`) or present in the system (`all_vms`). This was blocking for #1373 as there we need to exclude preloaded disposables as they should be automatically assessed by the system.

## Test plan

- [ ] Ensure number of executed tests does not change
- [ ] Visual review: test semantics do not change
- [ ] Tests passing

NOTE: at some point I was running into issues on exceeding the number of open files. I simply increased the ulimit on my system, but this could be a bigger problem. I didn't further investigate the root cause.

## Checklist

This change accounts for:

### DIFF
--- a/tests/test_vms_exist.py
+++ b/tests/test_vms_exist.py
@@ -29,8 +29,8 @@ def test_expected(all_vms, sdw_tagged_vms):
 
 @unittest.skipIf(CONFIG["environment"] != "prod", "Skipping on non-prod system")
 def test_internal(all_vms):
-    all_vms["sd-proxy-dvm"].features.get("internal") == "1"
-    all_vms["sd-viewer"].features.get("internal") == "1"
+    assert all_vms["sd-proxy-dvm"].features.get("internal") == "1"
+    assert all_vms["sd-viewer"].features.get("internal") == "1"
 
 
 def test_grsec_kernel(sdw_tagged_vms):


tests/test_vms_platform.py
```diff
@@ -21,177 +19,171 @@
...
+        stdout_lines = results.split("\n")
+        # filter out gpgcheck warning present on all dnf operations
+        stdout_lines = [
+            x for x in stdout_lines if not x.startswith("Warning: Enforcing GPG signature check")
+        ]
+        results = "".join(stdout_lines)
+        assert results == "", fail_msg
+
+
+@unittest.skipIf(IS_CI, "Skipping on CI")
```
legoktm, line 77: You can use https://docs.pytest.org/en/6.2.x/skipping.html#id1

tests/test_vms_exist.py
```diff
@@ -18,242 +17,244 @@
...
+    sdw_tagged_vm_names = [vm.name for vm in sdw_tagged_vms]
+    expected_vms = set(SD_VMS + SD_DVM_TEMPLATES + SD_TEMPLATES)
+    assert set(sdw_tagged_vm_names) == set(expected_vms)
+
+    # Check for untagged VMs
+    for vm_name in SD_UNTAGGED_DEPRECATED_VMS:
+        assert vm_name not in all_vms
+
+
+@unittest.skipIf(CONFIG["environment"] != "prod", "Skipping on non-prod system")
```
legoktm, line 30: use https://docs.pytest.org/en/6.2.x/skipping.html#id1

tests/conftest.py
```diff
@@ -11,3 +14,24 @@ def dom0_config():
     """Make the dom0 "config.json" available to tests."""
     with open(PROJ_ROOT / "config.json") as config_file:
         return json.load(config_file)
+
+
+@pytest.fixture
+def all_vms():
+    """Obtain all SecureDrop Workstation-exclusive qubes"""
+    return Qubes().domains
```
legoktm, line 22: I think the comment is wrong?

It's also not really clear what advantage we get from this being a fixture versus calling this directly where it's needed.

tests/test_vms_exist.py
```diff
@@ -18,242 +17,244 @@
...
+    for vm in sdw_tagged_vms:
+        if vm.name == "sd-app":
+            assert vm.default_dispvm.name == "sd-viewer"
+        else:
+            assert vm.default_dispvm is None, f"{vm.name} has dispVM set"
+
+
+def test_sd_whonix_absent(all_vms):
+    """
+    The sd-whonix once existed to proxy sd-proxy's traffic throgh Tor.
```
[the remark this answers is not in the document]
deeplow: Fixed in c2cabe9 (even though it was not introduced in this PR)

tests/test_vms_exist.py
```diff
@@ -18,242 +17,244 @@
...
+    whonix_qubes = [
+        "whonix-workstation-17",
+        "whonix-gateway-17",
+        "sys-whonix",
+        "anon-whonix",
+        "whonix-workstation-17-dvm",
+    ]
+    for qube_name in whonix_qubes:
+        if qube_name not in all_vms:
+            # skip check on nonexitent qubes
```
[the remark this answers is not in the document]
deeplow: Fixed in c2cabe9

tests/test_qubes_rpc.py
```diff
@@ -1,68 +1,58 @@
...
+        if vm.name != "sd-log":
+            assert policy_exists(vm, "sd-log", "securedrop.Log")
+
+
+# securedrop.Log from anything else to sd-log should be denied
+def test_sdlog_from_other_to_sdlog_denied(all_vms, sdw_tagged_vms):
+    non_sd_workstation_vms = set(all_vms).difference(set(sdw_tagged_vms))
+    for vm in non_sd_workstation_vms:
+        if vm.name != "sd-log":
+            assert not policy_exists(vm, "sd-log", "securedrop.Log")
```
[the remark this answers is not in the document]
deeplow: Not necessary. VM has `__repr__`.

tests/test_qubes_rpc.py
```diff
@@ -1,68 +1,58 @@
...
+    """verify the policies are installed"""
+    assert os.path.exists("/etc/qubes/policy.d/31-securedrop-workstation.policy")
+    assert os.path.exists("/etc/qubes/policy.d/32-securedrop-workstation.policy")
+
+
+# securedrop.Log from @tag:sd-workstation to sd-log should be allowed
+def test_sdlog_from_sdw_to_sdlog_allowed(sdw_tagged_vms):
+    for vm in sdw_tagged_vms:
+        if vm.name != "sd-log":
+            assert policy_exists(vm, "sd-log", "securedrop.Log")
```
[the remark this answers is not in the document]
deeplow: Not necessary. VM has `__repr__`.

tests/base.py
```diff
@@ -10,6 +10,9 @@
 SD_TEMPLATE_LARGE = f"sd-large-{DEBIAN_VERSION}-template"
 SD_TEMPLATE_SMALL = f"sd-small-{DEBIAN_VERSION}-template"
 
+SD_TAG = "sd-workstation"  # Tag identifying SecureDrop Workstation-managed VMs
+
+# Expectations regarding VMs' existance and versions
```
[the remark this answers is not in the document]
deeplow: Fixed in c2cabe9 (even though it was not introduced in this PR)

tests/test_viewer.py
```diff
@@ -12,9 +11,9 @@
 )
 
 
-def _create_test_qube(dispvm_template_name):
+def _create_test_qube(dispvm_template_name, all_vms):
```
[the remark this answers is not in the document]
deeplow: `all_vms` has a function-scope and so we can't really import it here. But it was in fact a problem. The practical solution was to just use `Qubes().domains` directly. I'm not pleased with this end result, but for this smaller-scoped PR it doesn't make sense looking for a better solution to deal with the various scopes.

tests/test_vms_exist.py
```diff
@@ -18,242 +17,244 @@
...
+
+    # Check for untagged VMs
+    for vm_name in SD_UNTAGGED_DEPRECATED_VMS:
+        assert vm_name not in all_vms
+
+
+@unittest.skipIf(CONFIG["environment"] != "prod", "Skipping on non-prod system")
+def test_internal(all_vms):
+    all_vms["sd-proxy-dvm"].features.get("internal") == "1"
+    all_vms["sd-viewer"].features.get("internal") == "1"
```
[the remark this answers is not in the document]
deeplow: Not really part of the original changes, but important enough to skip over. I would have opened this as a separate PR, but the indentation change introduced by this one would have made that PR dependent on this one, and at that point one might as well sneak those changes in here (via ca5554c). CC @legoktm

tests/test_vms_platform.py
deeplow: I had originally considering including these, but I ultimately decided to focus on just what was on the critical path for #1373. There are many other things I'd want address, but I had to inhibit myself from doing so, otherwise this would have been a massive PR. I only included [this scope-creep suggestion](https://github.com/freedomofpress/securedrop-workstation/pull/1487#discussion_r2540931203) because it was clearly an issue with the test and would lead to conflicts (due to indentation changes) if done in another PR.

So if you wouldn't mind, I'd prefer to address this separately. I have added this as a note in https://github.com/freedomofpress/securedrop-workstation/pull/1487#discussion_r2539418618, so we don't forget.

tests/test_vms_exist.py
deeplow: See https://github.com/freedomofpress/securedrop-workstation/pull/1487#discussion_r2540963828

tests/conftest.py
deeplow: One of the commit messages explains a bit:

```
commit 1581adb4def4f3eb24779e8d7d267406caa1f280
Author: deeplow <francisco@freedom.press>
Date:   Mon Nov 17 11:52:04 2025 +0000

    Tests: pytest fixtures for multiple tests
    
    Factors out some duplicate unittest class 'setUp', as a step towards
    moving away from unittest and into pytest (#1338). This refactor is
    necessary for the Qubes 4.3 work to avoid checking preloaded qubes
    during tests. Without consolidated fixtures with would have involved too
    much code duplication.
```

TL;DR for practical purposes for practically all tests we'll want to exclude preloaded qubes from being considered. So I wanted to centralize this `all_vms` to add the exclusion there, rather than doing it every time. The only exception I think will be actual preload tests. Does this make sense?

tests/test_vms_platform.py
legoktm: Sounds good.

tests/conftest.py
legoktm: Yes, just that I think we can have conslidation without it being a fixture. But yeah, can be done in a follow-up.

PR #1493, opened 2025-11-24
Improve GitHub actions behavior on "unkown" OpenQA status

deeplow:
## Test plan

 - [ ] visual review

Not sure how to go about simulating all of these scenarios. In my mind, this can be just a visual review and if it leads to problems we revisit it. It seems straightforward enough and it was directly copied from the docs.


legoktm (approved):
Per Conor's review

PR #1495, opened 2025-11-24
PyQt6 compatibility for Qubes 4.3

### LINKED ISSUE #1494: Support PyQt6 (default in Qubes 4.3)
## Description

PyQt6 is available by default in Qubes 4.3. Theoretically we could also install PyQt5 there, but if we can spare dom0 from another dependency, that's a positive. Therefore we should make our current code work with both versions, at least while Qubes 4.2 is around.

## Relevant PyQt5 → PyQt6 changes

They are practically all backwards compatible, but some are not forward compatible. In order words, PyQt6 is more stringent, but this should keep working on PyQt5. Here are some of the relevant changes to be adopted:

- replace "exec_()"  with exec() (Qt5 targeted also python2, which meant that it had "exec" reserved. This is no longer the case, so we can use the more aptly named "exec()" (**NOTE**: check if this works on PyQt5, otherwise add a `try...except`)
- Expand enums like `QMessageBox.Ok` to have its full name: `QMessageBox.StandardButtons.Ok`
    
A good overview of the upgrade notes can be found at https://www.pythonguis.com/faq/pyqt5-vs-pyqt6/

deeplow:
Fixes #1494. 
- Updates code to use be compatible with both PyQt5 and 6
- Adds the necessary modifications to be able to run tests on Qubes 4.3's dom0 equivalent CI environment

This is necessary for #1373 if we don't want to add pyqt6 as a dom0 dependency.
    
## Test plan

- [ ] lint test passes (it'll only check for pyqt6 syntax — do we want to create another poetry group that excludes PyQt6 in order to test PyQt5?)
- [ ] Building with `FEDORA_VERSION=37` (Qubes 4.2) and running `rpm -qR path/to/rpm` shows PyQt5 as depdency
- [ ] Building with `FEDORA_VERSION=41` (Qubes 4.3) and running `rpm -qR path/to/rpm` shows PyQt6 as depdency  
- [ ] test launcher application

## Checklist

This change accounts for:
   - There are some references that could be removed from the dev docs. I have filed a ticket for that: https://github.com/freedomofpress/securedrop-dev-docs/issues/275

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,8 +58,13 @@ jobs:
           make reprotest
   launcher-tests:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        qubes_release:
+          - { fedora_ver: "37", python_ver: "python3.11" }  # Qubes 4.2
+          - { fedora_ver: "41", python_ver: "python3.13" }  # Qubes 4.3
     container:
-      image: quay.io/fedora/fedora:37
+      image: quay.io/fedora/fedora:${{ matrix.qubes_release.fedora_ver }}
     steps:
       - run: dnf install -y make
       - uses: actions/checkout@v5
@@ -69,6 +74,7 @@ jobs:
         run: |
           make test-deps
           pip install poetry==2.1.1
+          poetry env use ${{ matrix.qubes_release.python_ver }}
           poetry install --no-ansi
       - name: Run launcher tests
         run: |


deeplow:
https://github.com/freedomofpress/securedrop-workstation/pull/1495/commits/151882ad0eba3c47bb16938adaa96a1aa3a8d533 introduces an issue where a `launcher-tests` job hangs:
Screenshot 2025-11-25 at 14-49-28 PyQt6 compatibility for Qubes 4 3 by deeplow · Pull Request #1495 · freedomofpress_securedrop-workstation

I think this is just a `.yml` misconfiguration, and that is actually a "ghost" job, but I can't spot what the problem is. Some help is welcome on this.

conorsch:
Used a 4.2 workstation install to test:

- [x] lint test passes (it'll only check for pyqt6 syntax — do we want to create another poetry group that excludes PyQt6 in order to test PyQt5?)
- [x] Building with `FEDORA_VERSION=37` (Qubes 4.2) and running `rpm -qR path/to/rpm` shows PyQt5 as depdency
- [x] Building with `FEDORA_VERSION=41` (Qubes 4.3) and running `rpm -qR path/to/rpm` shows PyQt6 as depdency  
- [x] test launcher application

I didn't check the 4.3 behavior, but 4.2 is higher priority for now, anyway. Nice conditional logic throughout, no complaints.

> https://github.com/freedomofpress/securedrop-workstation/commit/151882ad0eba3c47bb16938adaa96a1aa3a8d533 introduces an issue where a launcher-tests job hangs

Haven't resolved that yet, but will push some attempts, in the spirit of collaboration.

deeplow:
If merged today, I can follow up with infra tomorrow morning if nobody has filed it until then.

deeplow:
CI passed. But we need to have someone else also approve and merge this (CC @legoktm).

PR #1504, opened 2025-11-26
[unbreak nightlies] `openqa.yml`: add workflow inputs for `SECUREDROP_ENV` and `GIT_REF`

deeplow:
This accomplishes two things:
1. **Unbreaks nightlies** — since https://github.com/freedomofpress/securedrop-workstation/pull/1407 did not work and had to be reverted. It passes the git reference via workflow input, since this way we can explicitly set it to the main hash on nighties
    - **NOTE**: I needed to obtain the `main` commit hash because in `openqa.yml` it needs the specific hash to post a commit status message. Otherwise it fails [with error 422](https://github.com/freedomofpress/securedrop-workstation/actions/runs/19700982727/job/56437179439#step:5:53). About obtaining the main's hash, there is for sure a better way of doing it than what I did here. Improvements are welcome.
3. **Adds "environment"** also as input — the goal is to have this complement https://github.com/freedomofpress/openqa-tests-qubesos/pull/7 such that we can specify the type of installation we want to run on OpenQA: `dev` `staging` `prod` or `prod-qa`

I combined both into the same PR, since they use the same mechanism.

## Test plan

https://github.com/freedomofpress/securedrop-workstation/actions/runs/19702351741/job/56441426280
- [ ] Confirm 'nightly' was successfully simulated in bf63878 ([these](https://github.com/freedomofpress/securedrop-workstation/actions/runs/19702351741/job/56441426280), which I cancelled afterwards)
- [ ] OpenQA job runs on PRs successfully

## Checklist

This change accounts for: *(irrelevant for this PR)*

### DIFF
--- a/.github/workflows/nightlies.yml
+++ b/.github/workflows/nightlies.yml
@@ -90,9 +90,27 @@ jobs:
           git diff-index --quiet HEAD || git commit -m "Automated SecureDrop workstation build"
           git push https://x-access-token:${GH_TOKEN}@github.com/${TARGET_REPO}.git main
 
+  get-main-commit:
+    runs-on: ubuntu-latest
+    outputs:
+      main-commit: ${{ steps.get_main_sha.outputs.result }}
+    steps:
+      - name: Get SHA of main branch
+        id: get_main_sha
+        uses: actions/github-script@v8
+        with:
+          result-encoding: string
+          script: |
+            const { data: branch } = await github.rest.repos.getBranch({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              branch: 'main'
+            });
+            return branch.commit.sha;
   openqa-nightly-dev:
     uses: ./.github/workflows/openqa.yml
     secrets: inherit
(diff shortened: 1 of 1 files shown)


conorsch (approved):
LGTM. I've still got some learning to do on OpenQA specifically, but no concerns with this change set landing. We'll continue to monitor the QA and improve where we can.

conorsch:
> About obtaining the main's hash, there is for sure a better way of doing it than what I did here. Improvements are welcome.

The simplest approach that occurs to me is using `${{ github.sha }}` or `$GITHUB_SHA` wherever we need the commit SHA1; [docs here](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context). There's a bit of indirection in the OpenQA workflows, via `workflow_call`, so we'd have to test whether we can reuse the GITHUB_SHA in a workflow_call context and inherit the caller's context.

However, having the explicit dependencies across the workflows isn't a bad thing. If we can excise some JS-in-YAML, let's, but the most important thing is that It Works™.

PR #1506, opened 2025-12-01
Update sdw-admin `configure` action to check for packaged Tails config

### LINKED ISSUE #1502: Update sdwadmin configure to account for Tails config location changes in securedrop 2.13.0

zenmonkeykstop:
Fixes #1502 

Post 2.13.0 Tails Admin sticks will store the `app-journalist.auth-private` file under `~/.config/securedrop-admin`. This change checks there first, falling back to the old location on error, and failing if neither location works.

## Test plan
- [x] Run `sdw-admin configure` using a 2.13.0-style admin stick and confirm that a valid config.json is created with the correct auth hidden service info
- [x] Run `sdw-admin configure` using a 2.12.10-or-earlier style admin stick and confirm that a valid config.json is created with the correct auth hidden service info
- [x] Run it with the admin USB stick not attached at the right time, and confirm that the script fails with a useful error. 

## Checklist

This change accounts for:

### DIFF
--- a/files/sdw-admin.py
+++ b/files/sdw-admin.py
@@ -27,7 +27,8 @@ BASE_TEMPLATE = "debian-12-minimal"
 SUBMISSION_KEY = "sd-journalist.sec"
 TAILS_PATH = "/run/media/user/TailsData/"
 TAILS_GNUPG_PATH = TAILS_PATH + "gnupg/"
-TAILS_JOURNALIST_INTERFACE_CONFIG = (
+TAILS_PKG_JOURNALIST_INTERFACE_CONFIG = TAILS_PATH + "securedrop-admin/app-journalist.auth_private"
+TAILS_GIT_JOURNALIST_INTERFACE_CONFIG = (
     TAILS_PATH + "Persistent/securedrop/install_files/ansible-base/app-journalist.auth_private"
 )
 
@@ -396,15 +397,36 @@ def import_journalist_interface_config():
     Assumes that USB drive is attached to vault VM and decrypted.
     Returns (hostname, key) of the journalist interface hidserv
     """
-    journalist_interface_config = subprocess.check_output(
-        [
-            "qvm-run",
-            "--pass-io",
-            "vault",
-            f"cat {TAILS_JOURNALIST_INTERFACE_CONFIG}",
-        ],
-        text=True,
-    )
+    journalist_interface_config = ""
+    try:
+        # First, check for the 2.13.0+ location
+        journalist_interface_config = subprocess.check_output(
+            [
+                "qvm-run",
+                "--pass-io",
+                "vault",
+                f"cat {TAILS_PKG_JOURNALIST_INTERFACE_CONFIG}",
+            ],
+            text=True,
+        )
+    except subprocess.CalledProcessError:
+        try:
+            # Fall back to the legacy location
+            journalist_interface_config = subprocess.check_output(
+                [
+                    "qvm-run",
+                    "--pass-io",
+                    "vault",
+                    f"cat {TAILS_GIT_JOURNALIST_INTERFACE_CONFIG}",
+                ],
+                text=True,
+            )
+        except subprocess.CalledProcessError:
+            raise SDWAdminException(
+                "Failed to find a valid journalist interface config.\n"
+                "Check the attached USB key and try again."
+            )
+
     fields = journalist_interface_config.strip().split(":")
     addr = fields[0]
     auth_token = fields[3]
@@ -461,7 +483,12 @@ def import_config():
         if response.lower() != "y":
             print("Exiting.")
             return
-        ji_addr, ji_auth_token = import_journalist_interface_config()
+        try:
+            ji_addr, ji_auth_token = import_journalist_interface_config()
+        except SDWAdminException as e:
+            print(f"Error importing configuration: {e}")
+            sys.exit(1)
+
         print(
             "Journalist Interface details imported.\n\n"
             f"Onion address: {ji_addr}.onion\n"


files/sdw-admin.py
```diff
@@ -27,7 +27,8 @@
 SUBMISSION_KEY = "sd-journalist.sec"
 TAILS_PATH = "/run/media/user/TailsData/"
 TAILS_GNUPG_PATH = TAILS_PATH + "gnupg/"
-TAILS_JOURNALIST_INTERFACE_CONFIG = (
+TAILS_PKG_JOURNALIST_INTERFACE_CONFIG = TAILS_PATH + "/securedrop-admin/app-journalist.auth_private"
```
deeplow:
```suggestion
TAILS_PKG_JOURNALIST_INTERFACE_CONFIG = TAILS_PATH + "securedrop-admin/app-journalist.auth_private"
```

The path still resolves, but it'll have a double-slash: `/run/media/user/TailsData//securedrop-admin/app-journalist.auth_private`.

files/sdw-admin.py
```diff
@@ -396,15 +397,34 @@ def import_journalist_interface_config():
...
+        journalist_interface_config = subprocess.check_output(
+            [
+                "qvm-run",
+                "--pass-io",
+                "vault",
+                f"cat {TAILS_GIT_JOURNALIST_INTERFACE_CONFIG}",
+            ],
+            text=True,
+        )
+    except subprocess.CalledProcessError:
```
deeplow: I think you meant to catch this one inside the exception processing (one indentation level to the right). Otherwise it'll just raise the exception up the chain without triggering the sys.exit(1).

files/sdw-admin.py
```diff
@@ -396,15 +397,34 @@ def import_journalist_interface_config():
...
+        print(
+            "Failed to find a valid journalist interface config. Check the attached\n"
+            "USB key and try again."
+        )
+        sys.exit(1)
```
deeplow: I know in this script we don't have a well-established exception handling procedure (this code [is already inside some exception handling](https://github.com/freedomofpress/securedrop-workstation/blob/1.5.0/files/sdw-admin.py#L447)). But adding `sys.exit(1)` in the in the high stack depths of the script sounds like a case for surfacing an exception and `sys.exit(1)` lower in the stack. This way it's clearer that we're not abruptly terminating something.

```suggestion
    except subprocess.CalledProcessError as e:
        raise SDWAdminConfigException(
            "Failed to find a valid journalist interface config. Check the attached\n"
            "USB key and try again."
        )
```

files/sdw-admin.py
zenmonkeykstop: Will fix, that might catch us out down the line

files/sdw-admin.py
```diff
@@ -396,15 +397,34 @@ def import_journalist_interface_config():
...
+    journalist_interface_config = ""
+    try:
+        journalist_interface_config = subprocess.check_output(
+            [
+                "qvm-run",
+                "--pass-io",
+                "vault",
+                f"cat {TAILS_PKG_JOURNALIST_INTERFACE_CONFIG}",
+            ],
+            text=True,
```
deeplow, line 410: Stderr here is actually being printed out:

devnull

We should perhaps redirect stderr to stdout or just devnull

I'm guessing this applies to all other `qvm-run --pass-io ... cat` statements.

files/sdw-admin.py
```diff
@@ -396,15 +397,34 @@ def import_journalist_interface_config():
...
-            "qvm-run",
-            "--pass-io",
-            "vault",
-            f"cat {TAILS_JOURNALIST_INTERFACE_CONFIG}",
-        ],
-        text=True,
-    )
+    journalist_interface_config = ""
+    try:
+        journalist_interface_config = subprocess.check_output(
```
deeplow, line 403: I think a comment here could be useful to note that this is post-2.13.0 format and the other one before that.

files/sdw-admin.py
```diff
@@ -396,15 +397,34 @@ def import_journalist_interface_config():
...
+                    "--pass-io",
+                    "vault",
+                    f"cat {TAILS_GIT_JOURNALIST_INTERFACE_CONFIG}",
+                ],
+                text=True,
+            )
+        except subprocess.CalledProcessError:
+            raise SDWAdminException(
+                "Failed to find a valid journalist interface config. Check the attached\n"
+                "USB key and try again."
```
deeplow: Nit:

```suggestion
                "Failed to find a valid journalist interface config.\n"
                "Check the attached USB key and try again."
```

Because the exception text already prepends some text, looking like this:

```
Error importing configuration: Failed to find a valid journalist interface config. Check the attached
USB key and try again.
```

files/sdw-admin.py
deeplow: Fixed in https://github.com/freedomofpress/securedrop-workstation/compare/0ff6511d144b19a81486e30760f14115ed522634..6c1da0cdf02d2710fbe62fa6bc012216ac1170da#diff-6104c690fa1d1c1b216b58c8744593b8b49a9ec12c8c7eb4b8d101524cb02451L30-L30

files/sdw-admin.py
deeplow: Fixed in https://github.com/freedomofpress/securedrop-workstation/compare/0ff6511d144b19a81486e30760f14115ed522634..6c1da0cdf02d2710fbe62fa6bc012216ac1170da#diff-6104c690fa1d1c1b216b58c8744593b8b49a9ec12c8c7eb4b8d101524cb02451R488

files/sdw-admin.py
zenmonkeykstop: Let's keep the scope of this PR as is and fix this in a subsequent release. There might be some places where it's useful to have stderr, or we might wanna add an explicit debug mode.

deeplow: OK https://github.com/freedomofpress/securedrop-workstation/issues/1516

Issue #1339, opened 2025-06-05
Review existing python tests


deeplow:
Python tests need a review / improvement. While implementing #1329 I have come across many tests that could be improved.

Examples improvements:
- Better distinguish tests that run fast from slow ones (usually ones that require starting VMs). Perhaps this can best be done with `pytest.mark` and the ability to run smoke tests (the fast ones) and then a longer test run.
- `test_proxy_vm.test_do_not_open_here` is [no longer necessary](https://github.com/freedomofpress/securedrop-workstation/pull/1329/files/ca096ce83fb600da2275a3c4f058386818f4c5ab#diff-ede474fd05b08ab3877968a9b0faa29de80c151146858518870f46697605d829R18).
- #1331 
- `provisioning` separate from `configuration` to ensure that provisioning tests are fast (just checking settings/prefs/etc.), while the later is slower and needs to start qubes in order to ensure proper configuration.
- Shut down qubes on test teardown (to avoid leaving them hanging, consuming unecessary ammounts of RAM in the runner)
- Consider replacing generic method patching with [`mocker.spy`](https://pytest-mock.readthedocs.io/en/latest/usage.html#spy), when we don't necessarily want to isolate behavior, but rather test if methods have been called.
   - See in https://github.com/freedomofpress/securedrop-workstation/issues/1498 an example of where mocking, due to its power, actually suppressed issues that would have otherwise been caught by spy.
- **Better test pattern for comparing expected vs. existing** - many tests compare the existing state (obtained indirectly via `qubesadmin.Qubes().domains`) with a set of [expected qubes](https://github.com/freedomofpress/securedrop-workstation/blob/1.5.0/tests/base.py#L8-L22). The expected qubes are simply set by name and it's not really clear from the variable names that they are "expected".
  - Is there a way of making it clear that we're comparing expected state versus existing state?
  - Can we make use of Qubes' own [qube mocking methods](https://github.com/QubesOS/qubes-core-admin-client/blob/main/qubesadmin/tests/mock_app.py) to not treat these qubes as strings, but instead have stated properties and compare with the real ones? That would be a better way to phrase tests [like these](https://github.com/freedomofpress/securedrop-workstation/blob/1.5.0/tests/test_vms_exist.py#L185-L195). For example, fedora's system upgrade tool prefixes with `system_*` and `target_*` to distinguish them.
[...]

PR #1512, opened 2025-12-05
test: parallelizes integration test suite 

conorsch:
Incremental progress towards modernizing the integration tests that are run in dom0. The bulk of the diff in this PR is simply cutting-and-pasting test logic into discrete files, so that tests for e.g. `sd-viewer` reside in `test_vm_sd_viewer.py`. I was careful in preparing this diff to preserve the existing test functionality; the goal of submitting this refactor is to keep PRs small, thereby aiding in review of more substantive changes to come. 

Hopefully changes are unobjectionable. There's additional detail in the commit messages, which may aid review, and is OK to squash down during merge IMO. 

There's ~2x speed-up in the integration test run, and the test suite now displays the slowest tests. The use of xdist introduces an additional dom0 dependency, which I've documented and even bottled up into a new makefile target. What's still to come is a corresponding PR to the [openqa test repo](https://github.com/freedomofpress/openqa-tests-qubesos/). 

Towards #1339. Also includes nods to #1338.

### Local testing as sanity check

I recommend setting `export CI=true` before running the tests, to skip the package checks, which often fail. All other tests should pass! Future work can mark the package tests to make them easier to include or exclude, via pytest markers, as described in #1338. 

1. Clone to dom0: `make clone-norpm`
2. Run `make test`; observe informative error message about missing dependency
3. Install `python3-pytest-xdist` in dom0 as described by informative error message
4. Run `make test`; observe nice parallelized test output, including slow test info

### OpenQA

As this test suite is primarily invoked via OpenQA, corresponding changes must be made to the [openqa test suite repo](https://github.com/freedomofpress/openqa-tests-qubesos/), specifically in the [dom0 setup block](https://github.com/freedomofpress/openqa-tests-qubesos/blob/f3d96453b40f343ab200eb0436c417c9fec85e4b/tests/securedrop/test_dom0.pm#L34-L37).

In this PR, I've included a new `make dom0-install-test-prereqs` target to this repo's Makefile, so it's callable from the openqa suite. I haven't yet filed that PR, pending push rights, but will shortly. Thus the draft status. 

## Checklist

This change accounts for:
- [x] includes reference to corresponding changes in OpenQA repo: https://github.com/freedomofpress/openqa-tests-qubesos/pull/8

### DIFF
--- a/.github/workflows/openqa.yml
+++ b/.github/workflows/openqa.yml
@@ -25,6 +25,11 @@ jobs:
       GIT_REF: ${{ inputs.git_ref }}
       GH_TOKEN: ${{ github.token }}
       SECUREDROP_ENV: ${{ inputs.environment }}
+      # The branch to clone on the FPF-managed remote OpenQA repo:
+      # https://github.com/freedomofpress/openqa-tests-qubesos
+      SECUREDROP_OPENQA_REPO_BRANCH: "main"
+      # The URL to clone the remote OpenQA repo from (useful for forks):
+      SECUREDROP_OPENQA_REPO_URL: "https://github.com/freedomofpress/openqa-tests-qubesos"
     container:
       image: debian:trixie
     steps:
@@ -50,7 +55,7 @@ jobs:
             SECUREDROP_ENV="$SECUREDROP_ENV" \
             FLAVOR=securedrop \
             NEEDLES_DIR="%%CASEDIR%%/needles" \
-            CASEDIR="https://github.com/freedomofpress/openqa-tests-qubesos.git#main"\
+            CASEDIR="${SECUREDROP_OPENQA_REPO_URL}#${SECUREDROP_OPENQA_REPO_BRANCH}" \
             MAX_JOB_TIME=10800 | tee openqa.json
 
           for id in $(jq -r '.ids[]' openqa.json); do


conorsch:
Marking RFR. We should expect CI to be green across the board. @deeplow I think the order of operations should be as follows:

1. Review this PR. 
2. Review https://github.com/freedomofpress/openqa-tests-qubesos/pull/8 
3. Approve and merge https://github.com/freedomofpress/openqa-tests-qubesos/pull/8
4. Drop the "TEMPORARY" commit on this PR, pointing to the feature branch over in https://github.com/freedomofpress/openqa-tests-qubesos/pull/8
5. Wait again for CI on this repo. :clock1030: 
6. Approve and merge this PR.

What do you think?

Makefile
```diff
@@ -112,17 +112,21 @@ clone-norpm: assert-dom0 ## As above, but skip creating RPM
...
+	@test -e sd-journalist.sec || (echo "Ensure sd-journalist.sec is in this directory" && exit 1)
+	@rpm -q python3-pytest python3-pytest-cov python3-pytest-xdist || (echo 'please install test dependencies with "make install-dom0-test-prereqs"' && exit 1)
+
+.PHONY: install-dom0-test-prereqs
+install-dom0-test-prereqs: assert-dom0 ## Installs pytest dependencies in dom0
+	@rpm -q python3-pytest python3-pytest-cov python3-pytest-xdist || sudo qubes-dom0-update -y python3-pytest python3-pytest-cov python3-pytest-xdist
 
 test: test-prereqs ## Runs all application tests (no integration tests yet)
-	pytest -v tests -v launcher/tests
+	pytest -v tests -v launcher/tests -n auto --dist=loadfile --durations=5
```
deeplow, line 129: I've previously stored pytest in `pyproject.toml` to avoid duplication in the launcher tests. Could we move this there as well?

NOTE: I've now realized that the launcher tests actually also have extra options there. We should remove those as well (maybe out of scope for this one, but good to know).

Makefile
deeplow, line 1: This made me wonder how many vcpus dom0 had by default. The answer is great: the max available! :raised_hands:, according to the [xen wiki](https://wiki.xenproject.org/wiki/Tuning_Xen_for_Performance#Dom0_vCPUs). But I have a also confirmed with `xl vcpu-list Domain-0`.

Sadly these results don't translate to OpenQA, since it has only two CPUs available (see `QEMUCPUS 	2` in [the test settings](https://openqa.qubes-os.org/tests/161725#settings)). We could consider increasing these, but we better check with the Qubes team first, as [all other tests have it set to 2](https://github.com/search?q=repo%3AQubesOS%2Fopenqa-tests-qubesos+QEMUCPU&type=code). But checking the specs on some of the test machines, it should definitely be possible from a physical cores perspective.

It may even make sense to have a higher value just for the [`test_dom0`](https://openqa.qubes-os.org/tests/161725#dependencies) job and not the others.

tests/test_vms_exist.py
```diff
@@ -17,9 +17,19 @@
...
+    Seeks to detect errors in provisioning that result in VMs
+    failing to be created. Compares to a hardcoded list in fixtures.
+    """
     sdw_tagged_vm_names = [vm.name for vm in sdw_tagged_vms]
     expected_vms = set(SD_VMS + SD_DVM_TEMPLATES + SD_TEMPLATES)
+
+    # This integration test suite will create an ephemeral "sd-viewer-disposable" VM,
+    # and then destroy it, post-test-run.
+    sdw_tagged_vm_names = [vm for vm in sdw_tagged_vms if vm != "sd-viewer-disposable"]
+
```
deeplow, line 34: I'd mark this as a side-effect of `_create_test_qube` and the fact that tests are parallelized now.

My suggestion would be to exclude tests that create side-effects and remove this artifact from here. The Qubes Team's has [a way of isolating individual](https://dev.qubes-os.org/projects/core-admin/en/latest/qubes-tests.html#qubes.tests.SystemTestCase) tests from the rest of the system, but it's still in unittest format, which we're trying to get rid of and we'd have to evaluate it it suits our needs.

tests/test_vm_sd_viewer.py
```diff
@@ -28,8 +41,11 @@ def _create_test_qube(dispvm_template_name):
     return qube_name
 
 
-def _kill_test_qube(qube_name):
-    subprocess.run(["qvm-kill", qube_name], check=True)
+def _shutdown_test_qube(qube_name):
+    """
+    Gracefully power off the DispVM created for testing.
+    """
+    subprocess.run(["qvm-shutdown", "--wait", qube_name], check=True)
```
deeplow, line 48: What's the reason behind this change? Since this is a disposable, we don't really care about any state.

tests/test_vms_exist.py
```diff
@@ -109,141 +119,6 @@ def test_whonix_vms_reset(all_vms):
...
-    vm_type = vm.klass
-    assert vm_type == "DispVM"
-    assert SD_TAG in vm.tags
-
-    assert vm.features["service.avahi"] == "1"
-
-    # MIME handling
-    assert vm.features["service.securedrop-mime-handling"] == "1"
-    assert vm.features["vm-config.SD_MIME_HANDLING"] == "sd-devices"
-    check_service_running(vm, "securedrop-mime-handling")
```
deeplow, line 244: General comment on service assertions: at some point, I think it would be useful for us to also distinguish between provisioning assertions vs. configuration ones. Just because the former is much faster. My mental conception of `test_vms_exist` was to basically ensure the provisioning bit, but as you can see, it clearly got mixed with some service assertions.

I think the move to to have them in the same file makes sense because they are logically close. But maybe we can explore doing this with pytest markers at some point.

tests/test_vm_sd_app.py
deeplow, line 1: Idea: would it make sense to have a `tests/vm/` folder just to group all of these together? No strong feelings.

Makefile
conorsch: With only two vCPUs on the CI setup, I don't expect the CI runtimes to be much reduced, but on my dinky laptop that I'm using for SDW development, dom0 `nproc` reports `6`, which provides a significant speed-up of test runs locally over the non-xdist version.

Basically, I'm advocating for the parallelization not to speed up CI, but to speed up test runs locally for me, so I can get feedback faster after making changes.

Open to making asks of Qubes, but I don't think this is pressing, and I'd prefer to observe the new logic in CI for a while before making an ask for more resources.

tests/test_vm_sd_viewer.py
conorsch: The motivation to switch from `qvm-kill` to `qvm-shutdown` is simply to prefer graceful termination wherever possible. Specifically, I was remembering a problem where use of `qvm-kill` increased the likelihood of bugs in `qmemman` surfacing, discussed here: https://github.com/QubesOS/qubes-issues/issues/4890#issuecomment-606972698 Those bugs have since been patched, of course, so updating the test suite to abjure kills is not strictly necessary, but I think a good look. 

Also, the cross-ref should absolutely have been in the commit message—I got lazy while tracking down the issue and decided it wasn't worth it. As usual, your thoughtful review brought it back up into discussion. :slightly_smiling_face:

tests/test_vms_exist.py
conorsch: > My mental conception of test_vms_exist was to basically ensure the provisioning bit, but as you can see, it clearly got mixed with some service assertions.

Yes! It's good to check whether the VMs exist—but let's have a test named that that does just that! 

> But maybe we can explore doing this with pytest markers at some point.

100%, next PR will have a few marker additions around, that should give us a solid start.

tests/test_vm_sd_app.py
conorsch: I thought the same thing, and for now, decided against it, mostly out of adherence to "flat is better than nested," but also because we're not yet structuring the tests into cohesive submodules that would warrant their own directories. But my feeling isn't very strong in this regard, and I welcome any cleanup that aids discoverability by devs.

tests/test_vms_exist.py
conorsch: > My suggestion would be to exclude tests that create side-effects and remove this artifact from here.

I agree that there's a side-effect in the DispVM creation, but I'm not convinced that we should disable the test leveraging the DispVM creation, because testing those "ephemeral" VM environments is a valuable part of the test suite. Perhaps the most durable solution, that supports both comprehensive testing, and avoids transient errors caused by race conditions when running tests in parallel, is to use session-scoped fixtures for the VM lists: then the tag lookup will run once, at the start of the pytest suite, and that list of VMs will be used everywhere, removing the possibility of different lists being used. 

There's work in the "slice-2" forthcoming PR that does this, but let me try to adapt that change into this PR, so we're not taking any steps backwards.

conorsch: Turns out the implementation was a bit more complicated than just "use session-scoped vars", because we don't want to share the definition of "session" across pytest invocations of the launcher tests and the dom0 integration tests. To keep the test suites separate, I opted to create `pytest.ini` files for each suite, and provide the opts there. See details in https://github.com/freedomofpress/securedrop-workstation/pull/1512/commits/4f95203d4ccffd46f1f17507657eae4e782b1b19

Makefile
conorsch: Great point! While addressing another review comment, I moved the pytest options _out_ of the top-level pyproject.toml and into pytest.ini files for each suite; see details here: https://github.com/freedomofpress/securedrop-workstation/pull/1512/commits/4f95203d4ccffd46f1f17507657eae4e782b1b19

tests/test_vms_exist.py
conorsch: > To keep the test suites separate, I opted to create pytest.ini files for each suite

No, that didn't work: I've backed that change out. The attempt and reversion are backed up in [1339-slice-1-session-fixtures](https://github.com/freedomofpress/securedrop-workstation/tree/1339-slice-1-session-fixtures), if anyone's interested.

I've expanded the comments around the fixture for the DispVM to explain that it's ephemeral, and restored the exemption of the ephemeral VM from the generalized state fixture. That's by far the simplest solution here, but let's keep looking for improvements, especially since there's another follow-up in the wings.

tests/test_vm_sd_app.py
deeplow: Makes sense!

tests/test_vm_sd_viewer.py
deeplow: OK. That sounds reasonable.

tests/test_vms_exist.py
deeplow: OK. I guess we'll have to work with that for now, then. Thanks for trying.

Makefile
deeplow: I agree. This already makes a difference running locally.

Makefile
deeplow: That works. One thing I've noticed now is that the launcher tests that run in the GitHub CI are not [inheriting these parameters](https://github.com/freedomofpress/securedrop-workstation/blob/main/.github/workflows/ci.yml#L81) because they call out directly `make test-launcher` and we'd have `pytest-xdist` in poetry (which that CI job uses). Because of this coincidence of missing both the xdist parameters and the dependencies, it all works.

Honestly, I don't think it's worth addressing because our aim is not even to run the launcher tests faster in GitHub. I'll file a ticket about retiring that make target since they're now run in the workstation anyways...

conorsch: Yeah, I noticed the same when flipping between `pytest.ini` strategies. My focus on these changes has been entirely on the dom0 integration tests, because that's the tight feedback loop I want before making significant template changes. Once the dust settles on the integration tests, it should be comparatively little work to circle back and port any of those changes to the launcher tests, if we want.

deeplow:
> Marking RFR. We should expect CI to be green across the board. @deeplow I think the order of operations should be as follows:
> 
>    - [x] Review this PR.
>    - [] Review [test: install SDW dom0 deps via makefile openqa-tests-qubesos#8](https://github.com/freedomofpress/openqa-tests-qubesos/pull/8)
>    - [ ] Approve and merge [test: install SDW dom0 deps via makefile openqa-tests-qubesos#8](https://github.com/freedomofpress/openqa-tests-qubesos/pull/8)
>    - [ ] Drop the "TEMPORARY" commit on this PR, pointing to the feature branch over in [test: install SDW dom0 deps via makefile openqa-tests-qubesos#8](https://github.com/freedomofpress/openqa-tests-qubesos/pull/8)
>    - [ ] Wait again for CI on this repo. 🕥

I don't think it's worth waiting. Without the `install-dom0-test-prereqs` make target, main will essentially be broken, even if for a short while. This branch was recently rebased, so I don't think there's a risk.

 My suggestion would be approve and merge both after dropping the "TEMPORARY". Let me know if that works for you.

deeplow (approved):

PR #1520, opened 2025-12-16
Add `make run-app`

legoktm:
Same as `run-client`, but just for the app

## Test plan

* [x] After installing the app, try it with `make run-app`
## Checklist

This change accounts for:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,8 +139,8 @@ test-gpg: test-prereqs ## Runs tests for SD GPG functionality
 # Client autologin variables
 XDOTOOL_PATH=$(shell command -v xdotool)
 OATHTOOL_PATH=$(shell command -v oathtool)
-PHONY: run-client
-run-client: assert-dom0 ## Run client application (automatic login)
+.PHONY: run-deps
+run-deps:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -160,6 +160,26 @@ endif
...
+run-app: assert-dom0 ## Run SecureDrop App (automatic login)
+ifeq ($(XDOTOOL_PATH),)
+	@echo $(XDOTOOL_PATH)
+	@echo 'please install xdotool with "sudo qubes-dom0-update xdotool"'
+	@false
+endif
+ifeq ($(OATHTOOL_PATH),)
+	@echo 'please install oathtool with "sudo qubes-dom0-update oathtool"'
+	@false
+endif
```
deeplow: I think the the duplication in the remaining part is OK, but on this one my DRY suggestion here would be to just put them under the `run-deps` make target (without a "`##`" comment).

deeplow (changes requested):

Makefile
legoktm: Done.

deeplow (approved):

PR #1521, opened 2025-12-16
Import sd-gpg key from  dom0 on boot

### LINKED ISSUE #1505: Implement new RPC `securedrop.GetSecretKeys` (`sd-gpg`→ `@adminvm`) to replace salt's configuration
## Description

This is a follow-up to an architectural proposal in https://github.com/freedomofpress/securedrop-workstation/issues/1325, where we're aiming for "Proposal 0" in an initial phase.

> ### Proposal 0 |new RPC `securedrop.GetSecretKeys` (`sd-gpg`→ `@adminvm`) to replace salt's configuration
> 
> **NOTE:** This proposal does not comment on whether or not we should aim for some disposability of `sd-gpg`. Separating the commons parts between proposal 1 and 2 was one of the conclusions of a meeting on the topic.
> 
> Have a trivial dom0 RPC policy in `/etc/qubes-rpc/securedrop.GetSecretKeys` with the following content:
> 
> ```
> #/bin/sh
> cat  /usr/share/securedrop-workstation-dom0-config/secret-keys/*.sec
> ```
>
> And then an RPC policy allowing only `sd-gpg` to call it:
>
> ``` 
> # service_name			argument	source_qube	target_qube	action
> securedrop.GetSecretKeys	*		sd-gpg		@adminvm	allow
> ```
>
> This call would be done on-boot, for example.
> 
> Tradeoffs:
>     * (-) Adds attack surface to dom0, although it should be trivial enough to audit (read-only, and the attacker can only call without any input, which is better than many other domU -> dom0 qrexec services)
>     * (+) Gets rid of saltstack configuration of `sd-gpg`, which makes it be on par with all other qubes (except for templates)

 _Originally posted by @deeplow in [#1325](https://github.com/freedomofpress/securedrop-workstation/issues/1197#issue-2625375452)_

>     * (-) Adds attack surface to dom0, although it should be trivial enough to audit (read-only, and the attacker can only call without any input, which is better than many other domU -> dom0 qrexec services)

_Originally posted by @deeplow in [#1325](https://github.com/freedomofpress/securedrop-workstation/issues/1197#issue-2625375452)_

Theoretically, it adds some attack surface as this is one more service exposed in dom0. In practice, the calling qube (`sd-gpg`) should have no control overy any extra information beyond getting the result of that RPC's execution (which simply prints some files). See Marek's feedback [here](https://github.com/freedomofpress/securedrop-workstation/issues/1505#issuecomment-3636314623).

If one gets the ability to create arbitrary symlinks in dom0, then this further expose dom0 into revealing more significant information (by symlinking a sensitive file to the location where this new RPC cats the files), but a threat like that will already be quite significant. Therefore, this increased exposure alone, does not represent a significant security tradeoff.

deeplow:
Fixes https://github.com/freedomofpress/securedrop-workstation/issues/1505 by importing sd-gpg key on boot. We can't make it yet a disposable, because in some cases multiple keys could be present in `sd-gpg` and those are not backed up in dom0.

NOTE: I named the RPC policy as `keys`  and not `key` because in the future we'll probably want to add support for multiple ones.

<strike>Also [adds a mechanism to run scripts on boot](https://github.com/freedomofpress/securedrop-workstation/pull/1521/files#diff-296179b3a998617cd803e9293c9bb52dd4e748fae47065b28b74ad17ec4e34c3R42-R45) (it needs a counterpart component in the client repo). Some other alternatives being documented in https://github.com/freedomofpress/securedrop-workstation/issues/1522.</strike>

## Test plan

Setup:
- [ ] Do not delete your existing `sd-gpg`
- [ ] `./scripts/try-client-pr.py 3096`

Testing:
  - [ ] run `make dev`
- [x] open a terminal in `sd-gpg` and:
  - [x] check that the key is there (`gpg -K` should list it)
  - [x] `env | grep QUBES_GPG_AUTOACCEPT` should list it as set to `2147483647`
  - [x] `tail /home/user/.profile` — it should not have any `QUBES_GPG_AUTOACCEPT=...`
  - [x] remove the key you have there (either nuking the keyring `rm /home/user/.gnupg/pubring.kbx` or removing it with `gpg`)
- [x] Restart `sd-gpg` and open a terminal there:
  - [x] check that the key is there (`gpg -K` should list it)

## Checklist

This change accounts for:


deeplow:
@conorsch I decided to take a look at what exists out there. Turns out the answer is "a lot". I've documented these in https://github.com/freedomofpress/securedrop-workstation/issues/1522.

tests/test_vm_sd_gpg.py
```diff
@@ -86,13 +107,40 @@ def test_local_key_in_remote_keyring(config_fingerprint, dom0_fingerprint, vm_fi
     assert vm_fingerprint == [config_fingerprint]
 
 
+@pytest.mark.configuration
+def test_local_key_in_remote_keyring_clean(qube, config_fingerprint):
+    """
+    Confirm key presence in sd-gpg, but simulate clean environment.
+
+    This test exists because until sd-gpg is disposable or we run the test in a
+    perfectly new deployment, there is now way to tell if the key was already
```
legoktm:
```suggestion
    perfectly new deployment, there is no way to tell if the key was already
```

files/securedrop.GetJournalistSecretKeys
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
legoktm: I think I'd prefer this to be in Python, just given how security sensitive this is.

I think that would allow us to hardcode MAIN_SECRET_KEY_PATH and instead rely on mocking/monkeypatching in tests. (When I first saw it could be overridden by an env variable I spent a minute trying to figure out if qrexec/RPC allowed passing along environment variables...)

legoktm (changes requested):
I haven't tested this yet but I like where it's going.

files/securedrop.GetJournalistSecretKeys
deeplow: I've re-written this is python now as it's clearly superior for clarity / auditability. I also added an extra sanity check that ensures the request comes from `sd-gpg` (for the case of policy misconfiguration).

> (When I first saw it could be overridden by an env variable I spent a minute trying to figure out if qrexec/RPC allowed passing along environment variables...)

The same thought occurred to me at the time, but I dismissed it as the qrexec protocol is intended to be very thin in attack surface. Exposing environment variables to an attacker would be pretty bad. Furthermore qrexec / Qubes RPC itself relies on the [setting specific environment variables](https://github.com/QubesOS/qubes-core-qrexec/blob/e5a5b0a/libqrexec/exec.c#L188-L204). Therefore, an attacker being able to mess with environment variables in the target qube (dom0 in this case) would already be game over in this security model.

In any case, I also took a peek [at the docs for this specific case](https://doc.qubes-os.org/en/latest/developer/services/qrexec-internals.html#domx-request-execution-of-service-admin-service-in-dom0) and there's no mention of environment variables being in the mix.

> I think that would allow us to hardcode MAIN_SECRET_KEY_PATH and instead rely on mocking/monkeypatching in tests.

I don't think we can do this, at least the usual way for a few reasons:
1. We need to load a module from a source file
2. The RPC service prints directly to stdout (we need to catch those outputs)
3. It throws a `sys.exit` and we need to detect that within pytest (although it's likely pytest is built do be able to deal with this).

We can probably find a way but from experimenting a bit it didn't seem to be immediately obvious how to do it and many existing guides are outdated. I can look into it, if we really want to avoid environment variables.

legoktm:
(Rebased)

files/securedrop.GetJournalistSecretKeys
legoktm: I think all those 3 things are doable :) Let me give it a try...

tests/test_vm_sd_gpg.py
```diff
@@ -51,28 +51,49 @@ def dom0_fingerprint():
...
+
+def test_sd_gpg_timeout_not_in_home_profile(qube):
+    """This is mostly a test against lingering state"""
+    if qube.vm.klass == "DispVM":
+        pytest.fail("Test may no longer be needed")
+    assert "QUBES_GPG_AUTOACCEPT" not in qube.run("cat /home/user/.profile")
+
+
+@pytest.mark.configuration
+def test_dot_profile_was_reset(qube):
```
legoktm: This should be removed now

tests/test_vm_sd_gpg.py
```diff
@@ -86,13 +107,40 @@ def test_local_key_in_remote_keyring(config_fingerprint, dom0_fingerprint, vm_fi
...
+    perfectly new deployment, there is no way to tell if the key was already
+    there from before or if it was in fact placed there by the workstation version
+    under testing.
+    """
+    if qube.vm.klass == "DispVM":
+        # This condition is just a note for developers
+        pytest.fail("Test can be removed, as we no longer need to test against clean slate")
+
+    tmp_dir = qube.run("mktemp -d")
+    qube.run(f"GNUPGHOME={tmp_dir} securedrop-get-secret-keys")
```
legoktm, line 104: This wrapper no longer exists, right?

legoktm (changes requested):
Functionally this is great. A few remaining things:
* You should review my changes and approve + squash them into yours.
* I flagged some stuff about the tests, once those are fixed I think we can try and OpenQA run now that the securedrop-gpg-config package should be on apt-test.

launcher/tests/test_getsecretkeys_rpc.py
deeplow, line 1: This file was in the wrong place. I have now merged it.

tests/test_vm_sd_gpg.py
deeplow: True. I had forgotten that it was also useful in testing (monorepo, where are you?). Overriding env variables on system seems to much more messy (e.g. overrides via `.d` drop-in directory and then remove override).

Unless you have a trick up your sleeve, I'd it's probably just easier to do this test with a separate file. Would you agree?

tests/test_vm_sd_gpg.py
deeplow: Done in dd341d4.

deeplow (commented):
> You should review my changes and approve + squash them into yours.

I've done this now. The code included some parts that didn't feel relevant given the context, so I refactored it quite a bit. Please take a look.

tests/test_vm_sd_gpg.py
legoktm: Yep, I agree that having a standalone `securedrop-get-secret-keys` is useful for testing with no real downsides.

deeplow: I have now separated it (again) :sweat_smile: 

https://github.com/freedomofpress/securedrop-client/pull/3118

files/securedrop.GetJournalistSecretKeys
deeplow: Circling back with some findings.

When looking at your code, I realized that I had actually tested the same thing regarding module importing. But with one key difference: I was loading the RPC service from `/etc/qubes-rpc/securedrop.GetJournalistSecretKeys` and it was complaining about a missing loader. After some time I gave up.

Had I first tried with the script in `files/securedrop.GetJournalistSecretKeys.py` to begin with, it **would have worked** :cry:. The reason is that `importlib.utils.spec_from_file_location()` does not like python files that don't end in  `.py`.

However, since all of your tests test the actually deployed system, I think we want to test the real path and not the dev path. Therefore we needed to create a loader first:

```
loader = importlib.machinery.SourceFileLoader(MODULE_NAME, MODULE_PATH)
spec = importlib.util.spec_from_loader(MODULE_NAME, loader)
```

:rabbit: :hole:

legoktm: Sounds good to me. I had done it has a launcher test that way it could run locally but having it run against the real thing in dom0 works as well. Will review+test tomorrow.

legoktm (changes requested):
I'm having the same failure as 4.3 OpenQA locally:
```
# test_policy_from_sdgpg_to_dom0_allowed[sd-gpg]
# failure: 

# timestamp None
src_qube_name = 'sd-gpg'

    @pytest.mark.parametrize(
        "src_qube_name",
        [vm.name for vm in Qubes().domains if SD_TAG in vm.tags],
    )
    @pytest.mark.provisioning
    def test_policy_from_sdgpg_to_dom0_allowed(src_qube_name):
        allowed = policy_exists(src_qube_name, "@adminvm", "securedrop.GetSecretKeys")
    
        if src_qube_name == "sd-gpg":
>           assert allowed
E           assert False

tests/test_rpc_policy.py:101: AssertionError
```

deeplow:
This was in fact happening. I'll through it in my TODO list to investigate why that was happening. In the meantime, replacing `@adminvm` with `dom0` worked just fine.

legoktm (approved):
Let's do this!

PR #1515, opened 2025-12-09
OpenQA: start with PR labels

deeplow:
Fixes https://github.com/freedomofpress/securedrop-workstation/issues/1360 by:
  - Starting OpenQA test when <kbd>openqa-pending-X.Y</kbd> label is added 
  - Canceling when label removed
  - Adding Qubes version to OpenQA GitHub statuses (so it's clear which test is which)

## Options for discussion
- Label names? (current: `openqa-pending-X.Y`)
  - we can probably have spaces here
  - since labels are alphabetically sorted, starting with the version number may avoid a text search on the labels list.
- Cancelation implemented through the `concurrency` directive. This has some side-effects: "cancel" job is "skipped" every time it is supposed to run.

## Implementation Tasks

_**When code is pushed** (no labels)_
  - [x] Ensure the PR has no labels
  - [x] Push code (empty commit is fine) **without any labels**: no OpenQA test should start.

_**When label is added**_
  - [x] Ensure the PR has no labels
  - [x] Add <kbd>openqa-pending-4.2</kbd> label
     - [x] Qubes 4.2 OpenQA workflow is started
        (**NOTE:** do not cancel this workflow manually yet)
        Screenshot 2026-01-13 at 15-32-52 OpenQA start with PR labels by deeplow · Pull Request #1515 · freedomofpress_securedrop-workstation
     - [ ] Open the above mentioned workflow in a **new tab** (will be necessary later)
     - [x] After a few seconds (up to 1 minute) three new status checks appear in the format `OpenQA / 4.2 / <TEST_NAME>` 
        (one for each OpenQA "sub" test)
       Screenshot 2026-01-13 at 15-37-23 OpenQA start with PR labels by deeplow · Pull Request #1515 · freedomofpress_securedrop-workstation
     - [x] Check that this indeed is running on OpenQA by clicking `securedrop_install` and then on OpenQA going to settings and checking that the various variables are correct (`GIT_REF: (the head commit)` and `VERSION: 4.2`)
  - [x] Add <kbd>openqa-pending-**4.3**</kbd> label
    - [x] Qubes **4.3** OpenQA workflow is started (**NOTE:** do not cancel this workflow manually yet)
       Screenshot 2026-01-13 at 15-43-20 OpenQA start with PR labels by deeplow · Pull Request #1515 · freedomofpress_securedrop-workstation
    - [x] Wait 2 minutes for the tests to settle in place
[...]

PR #1525, opened 2025-12-17
test: use parametrization in dom0 integration tests 

### LINKED ISSUE #1339: Review existing python tests

conorsch:
Following up on the changes in #1512, this PR further refactors the dom0 integration tests, to make them more reliable when run repeatedly, and more pleasant to work with in general. Specifically, the changes are:

* reducing code duplication across tests
* bolstering the mimetype checks to check against known-good state
* automatically shutting down VMs after the test suite runs
* liberal application of custom pytest markers, for meaningful test filtering

There's more detail in the commit message for each of those changes. As with #1512, I've tried to keep the changes relatively unobjectionable, but some opinions did slip in. I like @deeplow's framing of "provisioning" i.e. are VMs created vs "configuration" i.e. do VMs contain the correct files; the markers reflect that, but so far my approach has been "just run all of the tests," and that takes about 2.5m on my 6-core Qubes laptop. Good enough for feedback between changes.

There's a tiny bit more work required in order to fulfill the asks in #1338; but I contend that the changes here, in #1512, plus the recent test improvements in #1329 and #1487, are sufficient for us to close #1339. 

## Test plan
CI passing is good enough, I recommend taking for a spin:

```
# in dom0
make clone-norpm && make test
```

You'll see a _lot_ more tests, mostly due to the parametrization of the mimetypes. As the commit message explains, parametrizing the mimetype checks doesn't reduce overall wall time of the test suite, which is a bummer, but it does add a lot more granularity to the tests, which I think is a win. If folks find this change too verbose, backing out the params is a very small diff to add on top of this: the other DRY refactors can stay IMHO.

## Checklist

This change accounts for:

### DIFF
--- a/tests/test_qubes_vms.py
+++ b/tests/test_qubes_vms.py
@@ -22,9 +22,8 @@ def test_current_fedora_for_sys_vms(sys_vm_name, all_vms):
     """
     Checks that all sys-* VMs are configured to use an up-to-date version of Fedora.
     """
-    # Look up all VMs that use the current Fedora as their TemplateVM.
-    fedora_based_vms = all_vms[CURRENT_FEDORA_TEMPLATE].derived_vms
     sys_vm = all_vms[sys_vm_name]
-    assert (
-        sys_vm in fedora_based_vms
-    ), f"VM {sys_vm_name} should have {CURRENT_FEDORA_TEMPLATE} as TemplateVM"
+    if sys_vm.klass == "DispVM":
+        assert sys_vm.template.template.name == CURRENT_FEDORA_TEMPLATE
+    else:
+        assert sys_vm.template.name == CURRENT_FEDORA_TEMPLATE


tests/test_qubes_vms.py
```diff
@@ -6,26 +8,24 @@
...
 
 
-def test_current_fedora_for_sys_vms(all_vms):
+@pytest.mark.provisioning
+@pytest.mark.parametrize(
+    ("sys_vm", "expected_templates"),
+    [
+        ("sys-firewall", [CURRENT_FEDORA_TEMPLATE, CURRENT_FEDORA_DVM]),
+        ("sys-net", [CURRENT_FEDORA_TEMPLATE]),
+        ("sys-usb", [CURRENT_FEDORA_TEMPLATE, f"sd-{CURRENT_FEDORA_DVM}"]),
```
deeplow: This change makes it a bit less evident that we're also testing for the case where sys-usb was made persistent through the settings. But I don't think we'd want to keep this configuration anyways, so I think it's fine to keep the code as-is.

I've just filed an issue to remove this possibility of a persistent sys-usb (see https://github.com/freedomofpress/securedrop-workstation/issues/1528).

deeplow: It's not a regression in this PR, but I've noticed two problems in this test (and it's original). I can follow-up with a separate PR as not to slip in too many additional changes.

1. `CURRENT_FEDORA_DVM` is wrongly set to be `fedora-XX-dvm` (the current fedora disposable is now named `default-dvm`).
2. `sys-usb` uses `default-dvm`. The test is missing this nuance.

May be related to https://github.com/freedomofpress/securedrop-workstation/issues/1528.

deeplow: In practice, I think we should probably drop this test in favor of a test like this:

```
# NOTE: Depending on operation time, it may be OK to pytest-parameterize
up_to_date_fedora_vms = all_vms[CURRENT_FEDORA_TEMPLATE]).derived_vms

for qube in [`sys-net`, `sys-usb`, `sys-firewall`, `default-mgmt-dvm`]:
  assert qube in up_to_date_fedora_vms
```

Or alternatively from each qube, we could obtain its template (or template's template) and check against `qvm-features <QUBE> os-eol`. It's not a perfect because this variable is set when the template is built and the actual EOL date may later be changed (depending on when the next version is released).

We can follow-up in a future PR.

tests/test_vms_platform.py
```diff
@@ -16,50 +14,34 @@
...
-    Retrieve PRETTY_NAME for an AppVM.
-    """
-    stdout, stderr = vm.run("cat /etc/os-release")
-    search = re.search(r'^PRETTY_NAME="(.*)"', stdout.decode())
-    if not search:
-        raise RuntimeError(f"Unable to determine platform for {vm.name}")
-    return search.group(1)
-
+    Checks that all available package updates are installed;
+    so upgrades pending. Assumes VM is Debian-based, so uses apt,
```
deeplow: Nit: typo

```suggestion
    no upgrades pending. Assumes VM is Debian-based, so uses apt,
```

tests/base.py
```diff
@@ -241,7 +267,45 @@ def test_lsm_enabled(self, qube):
...
         for profile in qube.enforced_apparmor_profiles:
             assert results["profiles"][profile] == "enforce"
+
+    def test_grsec_kernel(self, qube):
+        """
+        Confirms expected grsecurity-patched kernel is running.
+        """
+        # base doesn't have kernel configured
+        # TODO: test in sd-viewer based dispVM
+        exceptions = [SD_TEMPLATE_BASE, "sd-viewer", "sys-usb"]
```
deeplow: This is now solved, since it runs from the disposable :raised_hands:

```suggestion
        exceptions = [SD_TEMPLATE_BASE, "sys-usb"]
```

tests/test_vm_sd_devices.py
```diff
@@ -37,18 +37,17 @@ def test_logging_configured(qube):
...
+
+
+@pytest.mark.parametrize(("mime_type", "expected_app"), SD_DEVICES_MIME_TYPE_VARS)
+def test_mime_types(mime_type, expected_app, qube):
+    """
+    Functionally verifies that the VM config handles specific filetypes correctly,
+    opening them with the appropriate program.
+    """
+    actual_app = qube.run(f"xdg-mime query default {mime_type}")
+    assert actual_app == expected_app
```
deeplow: These mimetype changes are a pretty good improvement, but I think we can even further de-duplicate it.
`test_mime_types` is the same as in the viewer. It almost feels like it should be moved to `base.py` kind of how `linux_security_modules` is done. What do you think? (totally good if you think this is pattern not worth expanding)

_(I know it won't work for sd-proxy. But maybe we could have it in `sd-proxy` as well and then just implement one that checks that the last line of the original file is always `open-in-dvm.desktop;`. After all, that's what we're doing in practice)_

tests/test_vm_sys_usb.py
```diff
@@ -18,6 +18,8 @@ def qube():
     return QubeWrapper("sys-usb", linux_security_modules="selinux")
 
 
+@pytest.mark.configuration
+@pytest.mark.packages
```
deeplow: FYI, this one is done by Salt :cry:

```suggestion
```

pyproject.toml
```diff
@@ -125,3 +125,12 @@ exclude = [
...
 [tool.pytest.ini_options]
 addopts = "--cov-report term-missing --cov=sdw_notify --cov=sdw_updater --cov=sdw_util --junitxml=test-data.xml"
+
+# Registering custom pytest markers for tests. Allows selecting or skipping by marked groups.
+markers = [
+  "mime: checks for filetype/application associations; requires shell access",
+  "slow: long-running tests, longer than 1m in runtime",
+  "packages: checks that all packages are up to date; quite slow, generally",
+  "provisioning: checks that VMs exist and their extra-VM prefs are set correctly",
+  "configuration: assumes that VMs exist inspects their intra-VM files and folders",
```
deeplow, line 135: Works for me. The `provisioning` is a really good proxy for `fast`.

I'm wondering if the need for a `package` marker is indicative of some tests needing change (not in this PR, ofc). My thinking is that if we ever add one more package or one more mime-related tests we may forget to add the marker. This feels especially true in the case of packages.

Also regarding the packages, I didn't really notice that much slowness in running these. It's mostly a side-effect of it starting all the templates. Is there a different split that would be possible here? Just thinking out lout, but it could be for example separating the templates (where we'd check packages and updatedness) and the rest of the qubes. Not sure, just thought I'd mentioned it.

conorsch:
Blocked on https://github.com/freedomofpress/securedrop-workstation/pull/1533, which in turn is blocked on https://github.com/freedomofpress/securedrop-workstation-keyring/pull/41.

pyproject.toml
conorsch: Yeah, `packages` is somewhat redundant given that it is a proxy for "slow" and also for "the VM must be booted to have its state inspected." We already have `configuration`, after all. But I think it's worth keeping for now, as a first pass on having label-markers at all. I've been running these test suites a _lot_ locally, so I expect one or both of us to have opinions about which labels are useful and which are not, at which point we can refine them further. 

All that to say: merely a first pass on what's a suitable label!

deeplow: OK. Sounds good to me. You're right in that only with these in, can we know which are useful. Let's keep it then.

tests/test_qubes_vms.py
conorsch: Wow, I wasn't aware of the `os-eol` qvm feature—that's definitely something we should be checking! And delightfully, it's a very quick check via dom0, doesn't involve booting the VM or anything. Added in a test that references the EOL feature; it's set to fail if the EOL date is less than 1 month in the future. Could cause early test failures, but I'd kinda rather know that, and silence the test, if need be.

Also took your excellent suggestion about using the `.derived_vms` API call—also a TIL!—and the test is much more readable as a result.

tests/test_vm_sd_devices.py
conorsch: You're right, the mime types refactor went most of the way, but it can indeed be folded into `base.py`. Great call. That's done now. As for the extra "open-in-dvm" assert, I just removed it, since that's what's hardcoded in the `tests/vars/sd-proxy.mimeapps` file fixture anyway.

conorsch:
Implemented all review feedback. Amended the final commit to contain "run on openqa" to trigger the 4.2 openqa suite; we want to make sure this change passes there, because only the openqa suite exercises these modified integration tests. (Soon we'll have #1515 merged and can switch to using labels to trigger openqa runs.)

Assuming CI is green across the board, this is ready for another look, @deeplow!

tests/test_qubes_vms.py
```diff
@@ -1,31 +1,30 @@
...
-            f"Unexpected template for {sys_vm}\n"
-            + f"Current: {vm.template.name}\n"
-            + "Expected: {}".format(", ".join(wanted_templates))
-        )
+    # Look up all VMs that use the current Fedora as their TemplateVM.
+    fedora_based_vms = all_vms[CURRENT_FEDORA_TEMPLATE].derived_vms
+    sys_vm = all_vms[sys_vm_name]
+    assert (
+        sys_vm in fedora_based_vms
+    ), f"VM {sys_vm_name} should have {CURRENT_FEDORA_TEMPLATE} as TemplateVM"
```
deeplow: Thanks for implemented it. I know this was per my suggestion, but I had not realized how low the `.devived_vms` was. The tests took on average about 10.5 seconds to run. We could either cache this via a module-wide fixture or we could change it to check from the sys-vm's down like so:

```suggestion
    sys_vm = all_vms[sys_vm_name]
    if sys_vm.klass == "DispVM":
        assert sys_vm.template.template.name == CURRENT_FEDORA_TEMPLATE
    else:
        assert sys_vm.template.name == CURRENT_FEDORA_TEMPLATE
```

This implementation runs in 0.33s (total of 4 tests). Apologies for the inefficient suggestion.

deeplow (commented):
Thanks for implementing the suggestions. I had not actually realized my suggestion on the mimetypes would actually make it pretty impossible to parameterize.

Minor suggestion to fix the inefficiency I suggested. Other than that LGTM

tests/test_qubes_vms.py
conorsch: Heard, revised to the newer, leaner approach! Fix is on the branch, now rebased to include the labels CI functionality from #1515.

PR #1534, opened 2026-01-08
Enable SecureDrop App if `app: true` set in config.json

### LINKED ISSUE #1526: Implement initial app integration
We still need to finalize the details in https://github.com/freedomofpress/securedrop-client/issues/2907 but I think at a baseline we can have an `app: true` option in config.json, that when enabled:

1. Installs the `securedrop-app` package in the small template
2. Creates a second desktop icon for the app
3. Switches the launcher to start the app

This would only work in dev installs (once https://github.com/freedomofpress/securedrop-client/issues/2932 is done) and not prod ones because there is no prod securedrop-app package yet.

The `app` setting would just be temporary and will be removed before the real production release.

legoktm:
Have the launcher switch to launching the app and using it in strings based on `--launch-app`.
    
A separate "SecureDrop Client (legacy)" entry is added to the dom0 applications menu (unconditionally).

The config.json setting is expected to be temporary during development and if we have an opt-in period.

Fixes #1526.

## Test plan

* [ ] with config.json "app: false"
   * [ ] check out this branch and provision it with `make dev` 
   * [ ] the desktop icon launches the Client
   * [ ] in the Qubes menu -> {gear} -> Other, see two entries for "SecureDrop" and "SecureDrop Client (legacy)". The former should launch the app and the latter the client.
* [ ] switch config.json "app: true" and provision with `make dev`
   * [ ] the desktop icon launches the App
   * [ ] in the Qubes menu -> {gear} -> Other, see two entries for "SecureDrop" and "SecureDrop Client (legacy)". The former should launch the app and the latter the client.

## Checklist

This change accounts for:

### DIFF
--- a/securedrop_salt/sd-app-files.sls
+++ b/securedrop_salt/sd-app-files.sls
@@ -20,7 +20,7 @@ install-securedrop-client-package:
   pkg.installed:
     - pkgs:
       - securedrop-client
-      {% if d.app %}
+      {% if d.get('app', false) %}
       - securedrop-app
       {% endif %}
     - require:
--- a/securedrop_salt/sd-dom0-files.sls
+++ b/securedrop_salt/sd-dom0-files.sls
@@ -30,7 +30,7 @@ dom0-login-autostart-desktop-file:
     - context:
         desktop_name: SDWLogin
         desktop_comment: Updates SecureDrop Workstation DispVMs at login
-        desktop_exec: /usr/bin/sdw-login{% if d.app %} --launch-app{% endif %}
+        desktop_exec: /usr/bin/sdw-login{% if d.get('app', false) %} --launch-app{% endif %}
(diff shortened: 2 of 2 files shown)


securedrop_salt/sd-app-files.sls
```diff
@@ -12,11 +12,16 @@ include:
...
 
+{% import_json "securedrop_salt/config.json" as d %}
+
 # FPF repo is setup in "securedrop-workstation-$sdvars.distribution" template,
 # and then cloned as "sd-small-$sdvars.distribution-template"
 install-securedrop-client-package:
   pkg.installed:
     - pkgs:
       - securedrop-client
+      {% if d.app %}
```
legoktm: I thought jinja would automatically have `d.app` be False if no `app` property was set, but maybe I need to manually handle it somehow?

Issue #1530, opened 2026-01-06
failing test on dom0 yum repos


conorsch:
Over the past week or so, I've started seeing a failing test in the `tests/test_dom0_rpm_repo.py` checks:

```
>       assert found_lines == wanted_lines
E       AssertionError: assert ['[securedrop-workstation-dom0]',\n 'gpgcheck=1',\n 'skip_if_unavailable=False',\n 'gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation-test',\n 'enabled=1',\n 'baseurl=https://yum-test.securedrop.org/workstation/dom0/r$releasever',\n 'name=SecureDrop Workstation Qubes dom0 repo'] == ['[securedrop-workstation-dom0]',\n 'gpgcheck=1',\n 'skip_if_unavailable=False',\n 'gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation-test',\n 'enabled=1',\n 'baseurl=https://yum-test.securedrop.org/workstation/dom0/f37-nightlies/',\n 'name=SecureDrop Workstation Qubes dom0 repo']
E         At index 5 diff: 'baseurl=https://yum-test.securedrop.org/workstation/dom0/r$releasever' != 'baseurl=https://yum-test.securedrop.org/workstation/dom0/f37-nightlies/'
E         Full diff:
E           [
E            '[securedrop-workstation-dom0]',
E            'gpgcheck=1',
E            'skip_if_unavailable=False',
E            'gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation-test',
E            'enabled=1',
E         -  'baseurl=https://yum-test.securedrop.org/workstation/dom0/f37-nightlies/',
E         ?                                                            ^^^^^^^^^ -  ^
E         +  'baseurl=https://yum-test.securedrop.org/workstation/dom0/r$releasever',
E         ?                                                            ^^^^  + ^^^^
E            'name=SecureDrop Workstation Qubes dom0 repo',
E           ]

tests/test_dom0_rpm_repo.py:64: AssertionError
```

## Steps to reproduce

I'm able to reproduce this locally on commit 47a3065a11f2d4618f2f09f06c9f02ff9e32af50, merged this morning. The STR are:

```
make clone
make clean
# then additional cleaning, to be sure:
[...]

PR #1537, opened 2026-01-14
test: temporarily disable dom0 yum checks

conorsch:
Disables the dom0 yum repo checks as a stopgap measure to unblock other work streams. We have fixes in the pipeline, but until they're implemented and shipped, we cannot expect the yum repo checks to pass in the dom0 integration test suite.

We'll toggle them back on when the dust has settled on the rpm changes.

Refs https://github.com/freedomofpress/securedrop-workstation/issues/1530

## Test plan
Visual review is sufficient: notice the added underscore, which prevents unittest from processing the function as a test to run. This effectively ignores the test. 

Usually we'd want to use `pytest.mark.skip()` but conversion from unittest to pytest is ongoing, and blocked by #1530. I'll own following up to revert this change in a few days, whenever the blocking issue is resolved.

## Checklist

This change accounts for:

### DIFF
--- a/tests/test_dom0_rpm_repo.py
+++ b/tests/test_dom0_rpm_repo.py
@@ -45,7 +45,9 @@ class SD_Dom0_Rpm_Repo_Tests(unittest.TestCase):
             {"FEDORA_VERSION": FEDORA_VERSION}
         )
 
-    def test_rpm_repo_config(self):
+    # Temporarily disabling this test pending backend fixes:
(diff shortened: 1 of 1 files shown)


conorsch:
@deeplow I should have opted for this stopgap measure back when I reported the blockage; even though we're nearly out of the woods, I still think it's worthwhile to merge, since there are other PRs that would benefit, e.g. https://github.com/freedomofpress/securedrop-workstation/pull/1515

deeplow (approved):

PR #1542, opened 2026-01-22
tests: convert unittest to pytest for dom0 VM integration suite

conorsch:
Finishes the long-running refactor of dom0 VM integration tests, migrating them from `unittest` to `pytest`. Most of the relevant work was already merged in recent PRs:

  * https://github.com/freedomofpress/securedrop-workstation/pull/1512
  * https://github.com/freedomofpress/securedrop-workstation/pull/1525
  
and this small batch of related changes elides the final "unittest" references within the VM tests. Notably the SDW Launcher tests still use unittest; I'm less familiar with that test suite, and didn't include it in the aggressive refactoring.

With this changeset, I believe we can finally close https://github.com/freedomofpress/securedrop-workstation/issues/1338 on top of the already-closed https://github.com/freedomofpress/securedrop-workstation/issues/1339.

## Test plan

If you want to verify locally:

```
make clone-norpm
make test
```

and watch the output. But in terms of verifying that the tests are passing, let's make sure the OpenQA job (triggered via labels, yay! #1515) passes on this PR prior to merging. 

## Checklist

This change accounts for:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -22,11 +22,11 @@ def dom0_config():
     with open(os.path.join(PROJ_ROOT, "config.json")) as c:
         config = json.load(c)
         # TODO: in the future, when "config.json" does not include an env declaration,
-        # we should either assume "prod" or else infer the env from the keyring
-        # packages that are installed. See the `_get_env_by_package` function
-        # elsewhere in this test suite for details.
+        # If the "environment" key is absent from the "config.json" file, assume prod,
+        # as a sane default. Dev environments will have it set explicitly.
         if "environment" not in config:
-            config["environment"] = "dev"
+            pytest.warn("no 'environment' detected in config.json, assuming prod")
+            config["environment"] = "prod"
     return config
 
 
--- a/tests/test_dom0_rpm_repo.py
+++ b/tests/test_dom0_rpm_repo.py
@@ -35,18 +35,6 @@ def _is_installed(pkg: str) -> bool:
         return False
 
 
-def _get_env_by_package():
-    """
-    Check which environment we are using by checking keyring package.
-    The order matters; due to package versioning, an installed dev package
-    means a dev setup, so check dev, staging, then prod.
-    """
-    for env, suffix in [("dev", "-dev"), ("staging", "-staging"), ("prod", "")]:
-        if _is_installed(f"securedrop-workstation-keyring{suffix}"):
-            return env
-    pytest.fail("No keyring package")
-
-
 def test_rpm_releasever_substitution():
     """
     Ensure DNF is understanding $releasever as we expect in the repository
@@ -56,15 +44,13 @@ def test_rpm_releasever_substitution():
     assert qubes_version in ["4.2", "4.3"]
 
 
-def get_repo_config(dom0_config):
+@pytest.fixture(scope="session")
+def repo_config(dom0_config):
     """
     Look up the appropriate Yum repo configuration, based on config.json.
-
-    We avoid declaring this function as a pytest fixture, because we want
-    to make use of the existing fixture for the dom0 config.json, and pytest
-    fixtures cannot depend on other fixtures.
+    Depends on the `dom0_config` fixture, for lookup on the environment,
+    which determines the rpm config.
     """
-
     # The dom0 fixture defaults to "dev" for the environment key.
     # It's possible that we can fall back to inferring the env based on
     # which keyring RPM packages are installed in dom0.
@@ -72,13 +58,12 @@ def get_repo_config(dom0_config):
     return REPO_CONFIG[env].copy()
 
 
-def test_rpm_repo_config(dom0_config):
+def test_rpm_repo_config(repo_config):
     """
     Inspect the dom0 yum repo config for the SecureDrop Workstation RPM repository,
     and verify the settings are what we expect. Some of the attributes vary
     by env; see top-level REPO_CONFIG.
     """
-    repo_config = get_repo_config(dom0_config)
     repo = repo_config["repo_file_name"]
     baseurl = repo_config["yum_repo_url"]
     repo_file = f"/etc/yum.repos.d/{repo}"
--- a/tests/test_dom0_validate.py
+++ b/tests/test_dom0_validate.py
@@ -1,6 +1,4 @@
-import os
 import shutil
-import tempfile
 from pathlib import Path
 
 import pytest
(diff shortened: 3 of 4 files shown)


tests/test_dom0_rpm_repo.py
```diff
@@ -23,83 +22,96 @@
...
+    assert qubes_version in ["4.2", "4.3"]
+
+
+def get_repo_config(dom0_config):
+    """
+    Look up the appropriate Yum repo configuration, based on config.json.
+
+    We avoid declaring this function as a pytest fixture, because we want
+    to make use of the existing fixture for the dom0 config.json, and pytest
+    fixtures cannot depend on other fixtures.
```
deeplow: I think fixtures can be based on fixtures. Take a look at the [cleanup fixture](https://github.com/freedomofpress/securedrop-workstation/pull/1542/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R50). It's just a matter of passing it as an argument in the fixture, just like a regular fixture.

tests/conftest.py
```diff
@@ -18,8 +19,15 @@
...
-    with open(PROJ_ROOT / "config.json") as config_file:
-        return json.load(config_file)
+    with open(os.path.join(PROJ_ROOT, "config.json")) as c:
+        config = json.load(c)
+        # TODO: in the future, when "config.json" does not include an env declaration,
+        # we should either assume "prod" or else infer the env from the keyring
+        # packages that are installed. See the `_get_env_by_package` function
+        # elsewhere in this test suite for details.
+        if "environment" not in config:
+            config["environment"] = "dev"
```
deeplow: I know we had this already elsewhere, but now that I think about it, wouldn't it be riskier to default to `dev` rather than defaulting to `prod`? My thinking is that that the prod environment is necessarily more strict than `dev` (even though in practice we're not running these tests in prod.

Also, for context I was personally never too keen on having the keyring package as a proxy for the environment, as it just complicates things. But I agree with the goal of not needing `environment = 'prod'` in production. We should just assume prod unless otherwise stated.

What do you think about defaulting to `prod` or simply `pytest.fail('config.json without "environment" not yet supported')`? In any case, if removing the `_get_env_by_package` from here, it should be removed entirely as it's no longer called anywhere else.

tests/test_dom0_rpm_repo.py
```diff
@@ -23,83 +22,96 @@
...
+def _get_env_by_package():
+    """
+    Check which environment we are using by checking keyring package.
+    The order matters; due to package versioning, an installed dev package
+    means a dev setup, so check dev, staging, then prod.
+    """
+    for env, suffix in [("dev", "-dev"), ("staging", "-staging"), ("prod", "")]:
+        if _is_installed(f"securedrop-workstation-keyring{suffix}"):
+            return env
+    pytest.fail("No keyring package")
```
deeplow: As of right now, this is dead code and we could remove it.

tests/conftest.py
deeplow: Maybe this is one where others have thoughts?

tests/test_dom0_rpm_repo.py
deeplow: No strong opinions on whether or not this is implemented. Just wanted to mention.

tests/test_dom0_validate.py
```diff
@@ -1,68 +1,66 @@
...
-            shutil.copy(f"{self.test_resources}/files/example_key.asc", f"{dir}/sd-journalist.sec")
 
-            # Validator currently runs checks in constructor
+def test_good_config(test_resources_dir):
+    with tempfile.TemporaryDirectory() as dir:
+        shutil.copy(f"{test_resources_dir}/testconfig.json", f"{dir}/config.json")
+        shutil.copy(f"{test_resources_dir}/example_key.asc", f"{dir}/sd-journalist.sec")
+
+        # Validator currently runs checks in constructor
+        SDWConfigValidator(dir)
```
deeplow: Has [the perfect fixture](https://docs.pytest.org/en/6.2.x/tmpdir.html) for this :raised_hands:
```suggestion
def test_good_config(test_resources_dir, tmpdir):
    shutil.copy(f"{test_resources_dir}/testconfig.json", f"{tmpdir}/config.json")
    shutil.copy(f"{test_resources_dir}/example_key.asc", f"{tmpdir}/sd-journalist.sec")

    # Validator currently runs checks in constructor
    SDWConfigValidator(tmpdir)
```

tests/test_dom0_validate.py
```diff
@@ -1,68 +1,66 @@
...
-    def test_missing_config(self):
-        with tempfile.TemporaryDirectory() as dir:
-            with pytest.raises(ValidationError) as exc_info:
-                SDWConfigValidator(dir)
+        assert "Config file does not exist" in exc_info.exconly()
+
 
-            assert "Config file does not exist" in exc_info.exconly()
+def test_config_malformed_key(test_resources_dir):
+    with tempfile.TemporaryDirectory() as dir:
```
deeplow: See `tmpdir` fixture comment.

tests/test_dom0_validate.py
```diff
@@ -1,68 +1,66 @@
...
+        assert "PGP secret key file provided is not an armored private key" in exc_info.exconly()
 
-            with pytest.raises(ValidationError) as exc_info:
-                SDWConfigValidator(dir)
 
-            assert (
-                "PGP secret key file provided is not an armored private key" in exc_info.exconly()
-            )
+def test_config_malformed_onion_json(test_resources_dir):
+    with tempfile.TemporaryDirectory() as dir:
```
deeplow: See `tmpdir` fixture comment.

tests/test_dom0_validate.py
```diff
@@ -1,68 +1,66 @@
...
-            assert "Invalid hidden service hostname specified" in exc_info.exconly()
 
-    def test_config_malformed_fpr_json(self):
-        with tempfile.TemporaryDirectory() as dir:
-            shutil.copy(
-                f"{self.test_resources}/files/testconfig.json.malformedfpr", f"{dir}/config.json"
-            )
-            shutil.copy(f"{self.test_resources}/files/example_key.asc", f"{dir}/sd-journalist.sec")
+def test_config_malformed_fpr_json(test_resources_dir):
+    with tempfile.TemporaryDirectory() as dir:
```
deeplow: See `tmpdir` fixture comment.

tests/test_dom0_validate.py
```diff
@@ -1,68 +1,66 @@
...
-    def setUp(self):
-        # Enable full diff output in test report, to aid in debugging
-        self.maxDiff = None
-        self.test_resources = Path(__file__).parent.resolve()
+@pytest.fixture
+def test_resources_dir():
+    """
+    Return path to directory hard-coded test data files.
+    """
+    return os.path.join(Path(__file__).parent.resolve(), "files")
```
deeplow: Out of scope / nit suggestion, but since we're already adding "files" to it:
```suggestion
    return Path(__file__).parent.resolve() / "files"
```

(I mention it just because another suggestion of mine is the pytest's fixture `tmpdir` which is also a Path and just makes it easier to work with)

Also it might just be personal opinion of `Path` over `os.path`. Pet peeve; totally fine to keep as is.

tests/test_dom0_salt_config.py
```diff
@@ -1,19 +1,15 @@
 import subprocess
-import unittest
 
+import pytest
 
-class SD_Dom0_Salt_Config_Tests(unittest.TestCase):
-    def setUp(self):
-        # Enable full diff output in test report, to aid in debugging
-        self.maxDiff = None
```
deeplow, line 8: This has helped in the past. Do we want to fully removed it? The [pytest docs](https://docs.pytest.org/en/7.1.x/how-to/output.html) didn't really give much clarity on how to achieve this. But [someone](https://til.unessa.net/testing/pytest-maxdiff/) figured out that a way around that: (note I haven't tested it)

```
import unittest
unittest.TestCase.maxDiff = None
```

It's a bit "dirty", through.

tests/test_dom0_rpm_repo.py
conorsch: You are so right. I was confused. Appreciate you pointing this out, because the code is much cleaner reusing the fixtures.

tests/conftest.py
conorsch: I'm not sure how often folks will be running the pytest suite in a prod environment, but your point stands: dev is explicitly configured as dev in config.json, so I've updated the default assumption to be prod. Existing dev setups will continue to be tested as dev setups, appropriately.

tests/test_dom0_salt_config.py
conorsch: I added an explanatory comment about the `-vv` in the pytest invocation that we're already using, which should provide maximum diffing already. Personally I'd like a bit _less_ verbosity in the tests by default, but that's mostly an artifact of me having run them a bunch recently, so let's keep as-is for now.

tests/test_dom0_rpm_repo.py
```diff
@@ -35,19 +35,7 @@ def _is_installed(pkg: str) -> bool:
...
-    means a dev setup, so check dev, staging, then prod.
-    """
-    for env, suffix in [("dev", "-dev"), ("staging", "-staging"), ("prod", "")]:
-        if _is_installed(f"securedrop-workstation-keyring{suffix}"):
-            return env
-    pytest.fail("No keyring package")
-
-
-def test_rpm_releasever_substitution():
+def _test_rpm_releasever_substitution():
```
deeplow: Might have been by mistake, but this now has a `_` prefix.

tests/test_dom0_validate.py
deeplow: Thanks! I just noticed that that these three could be folded into parameters to avoid repetition, but it's really minor. Let's leave that for a future PR, if we bother to do it.

tests/test_dom0_rpm_repo.py
deeplow: I have double-checked and this is no longer getting tested:

Screenshot 2026-01-26 at 19-47-09 Qubes OS openQA qubesos-4 2-securedrop-x86_64-Build202601261802-4 2-securedrop_test_dom0@64bit test results

Other than that LGTM

tests/test_dom0_salt_config.py
deeplow: Oh. I see. Not sure if it does the exact same thing. Some of the VM list comparisons may get truncated, but we can deal with that if it becomes a problem. I agree some of the verbosity is too much (e.g `tests/test_vm_sd_log.py::test_logs_are_flowing` prints the whole log file in `sd-log`).

tests/test_dom0_rpm_repo.py
conorsch: Eek, that was indeed a typo in the latest commit. Fixed, pushed.

PR #1546, opened 2026-01-22
Test nightlies.yml whenever workflow is edited

### LINKED ISSUE #1545: Simplify testin `nightlies.yml` changes
Some [recent](https://github.com/freedomofpress/securedrop-workstation/issues/1470) changes have broken nightlies unintentionally. In generally it's hard to test some CI action that is not triggered when we're trying to fix it.

**Solution:** Can we trigger nightly runs whenever `nightlies.yml` is changed in a PR?

deeplow:
Some issues have been caused in the past due to workflow changes that affected the nightlies. This triggers a limited nightly run whenever nightlies.yml is edited.

Fixes #1545

## Test plan

- [ ] check that nightlies is triggered (**note:** failing job is acceptable here if due to https://github.com/freedomofpress/securedrop-workstation/pull/1544)
- [ ] push empty commit and check that nightlies does NOT run
- [ ] (after merging) check that the next nightly is correctly triggered

## Checklist

This change accounts for: (NOT NECESSARY)

### DIFF
--- a/.github/workflows/nightlies.yml
+++ b/.github/workflows/nightlies.yml
@@ -5,11 +5,14 @@ on:
   push:
     branches:
       - main
+  pull_request:  # To test changes in this workflow
+    paths:
+      - '.github/workflows/nightlies.yml'
 
 # Only allow one job to run at a time because we're pushing to git repos;
 # the string value doesn't matter, just that it's a fixed string.
 concurrency:
-  group: "just-one-please"
+  group: "just-one-please-${{ github.ref }}"
 
 defaults:
   run:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightlies.yml
```diff
@@ -41,12 +44,14 @@ jobs:
           make build-rpm
       - uses: actions/upload-artifact@v6
         id: upload
+        if: ${{ github.ref == 'refs/heads/main' }} # Skip when testing workflow
```
legoktm: Actually this is fine, we don't need to skip here, we only need to skip the commit-and-push job below.

legoktm (changes requested):
One minor thing but otherwise LGTM. Good idea, probably worth doing in securedrop-client as well :)

.github/workflows/nightlies.yml
deeplow: OK. I have now removed it (via force-push).

PR #1551, opened 2026-01-27
Have try-client-pr autostart sd-proxy

### LINKED ISSUE #1550: app: start up sd-proxy VM on app open
We should start the sd-proxy VM when the app opens instead of on the first login request

legoktm:
Since it shuts down all the VMs, apply the same-ish hack as 428222f559 so that the proxy and Tor are running by the time we try to log in.

Fixes #1550.

## Test plan

* Use the script and see that it still shuts down all the VMs but then starts sd-proxy afterwards.

## Checklist

This change accounts for:


deeplow (approved):
LGTM. We could also (or alternatively) put this in `make run-client`. But fine either way.

PR #1553, opened 2026-01-30
Deny attachment of devices except to `sd-devices`

deeplow:
Deny attachment of devices except to `sd-devices`. This is a temporary workaround while there is no salt support for https://github.com/QubesOS/qubes-mgmt-salt-dom0-qvm/pull/36. Once we have that we can add it under `qvm.prefs`.

This also adds a test where it creates a block device exposed in sys-usb and then attempts to attach it the expected qube. Unfortunately there was not enough information in the `qvm-block attach` command and systemd journal logs were necessary to confirm the attach denial.

Fixes #1428. Depends on https://github.com/freedomofpress/securedrop-workstation/pull/1373 (**change base to `main` once that's merged**)

## Test plan
- [ ] (Qubes 4.2) Passing on OpenQA should be sufficient, since it shows that devices_denied is not being applied (otherwise it would fail)
Qubes 4.3:
- [x] Insert USB drive and attempt to attach to sd-app 
- [x] The devices widget should prevent it (*blocked by policy*)
   policy

## Checklist

This change accounts for:

### DIFF
--- a/securedrop_salt/sd-app.sls
+++ b/securedrop_salt/sd-app.sls
@@ -30,6 +30,9 @@ sd-app:
       - template: sd-small-{{ sdvars.distribution }}-template
       - netvm: ""
       - default_dispvm: "sd-viewer"
+      {% if grains['osrelease'] != '4.2' %}
+      - devices_denied: '*******'
+      {% endif %}
     - tags:
       - add:
         - sd-client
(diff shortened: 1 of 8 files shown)


legoktm:
(Rebased)

legoktm:
I'll wait for the 4.2 OpenQA run to pass before merging. I verified locally that `make test` passes on 4.3.

PR #1555, opened 2026-01-30
Custom persist: `sd-app`, `sd-gpg` and `sd-log`

### LINKED ISSUE #1265: Make use of custom-persist on app qubes
## Description

Limit persistence on certain directories in the app qubes with [`custom-persist`](https://github.com/QubesOS/qubes-issues/issues/1006).

No impact.

It adds a security mitigation that makes malware persistence from something trivial to accomplish (writing to `/rw/config/rc.local` or `~/.config/autostart/`) to something slightly more involved (using directories which execute on file dropping as opposed to machine restart, or compromising the `/dev/xvdb` [disk itself](https://github.com/QubesOS/qubes-issues/issues/1006#issuecomment-2567294118)). But it should not change the threat model itself.

deeplow:
Fixes #1265.

Custom-persist is a feature introduced in Qubes 4.3 that, as the name implies, enables persisting particular files. This turns these qubes into semi-disposables. This makes an attack slightly more challenging if persistence is the goal, although it is by no means equivalent to the properties of full disposables (e.g. here the persistence is enforced by the kernel and not the hypervisor).

Properties:
  - qubes should retain only specified files and directories
  - data previously stored in the user's home is preserved but not mounted in `/home` (you can find it under `/rw/home/user`)
  - arbitrary data can be persisted in `/rw/`, but it won't be mounted in critical locations.

## Test plan

**Before running make-dev**
- [x] Start client application and send and receive some messages (to ensure that there is some state
- [x] (repeat for the app)
- [x] `sdw-admin --apply` or `make dev` with this branch
- [x] Start client and ensure state persisted. Play around and see if there are any glitches caused by files going away. Also watch out for slowness due to missing caches.
- [x] (repeat for the app)

**Clean installation** — also test in a clean environment (`sdw-admin --uninstall` is fine`). This is important because some in existing installations the files have already been created under `/rw/` and it's possible we run into some edge-conditions where being saved for the first time can make then not persist.

**Code review:** (particular things to watch out for)
- [ ] Confirm that no other `sd-workstation`-tagged remain with full persistence (except disposable templates — should we also add it to them?)
- [ ] Are there any other directories/files that should also be included
- [ ] Of the persisted files / directories does any allow the persistence of easily executable code (.e.g `/home/user/.profile`) — we don't want this kind of behavior

## Checklist

This change accounts for:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,6 +18,10 @@ from tests.base import (
 
 PROJ_ROOT = Path(__file__).parent.parent
 
+skip_on_qubes_4_2 = pytest.mark.skipif(
+    dnf.rpm.detect_releasever("/") == "4.2", reason="Feature only available in Qubes >= 4.3"
+)
+
 
 @pytest.fixture
 def qubes_ver():
--- a/tests/test_vms_exist.py
+++ b/tests/test_vms_exist.py
@@ -9,6 +9,7 @@ from tests.base import (
     SD_UNTAGGED_DEPRECATED_VMS,
     SD_VMS,
 )
+from tests.conftest import skip_on_qubes_4_2
 
 
 @pytest.mark.provisioning
@@ -31,6 +32,35 @@ def test_all_sdw_vms_present(all_vms, sdw_tagged_vms):
     # Check for untagged VMs
     for vm_name in SD_UNTAGGED_DEPRECATED_VMS:
         assert all_vms.get(vm_name) is None, f"Qube '{vm_name}' expected"
+
+
+@skip_on_qubes_4_2
+@pytest.mark.provisioning
+def test_expected_persistence(sdw_tagged_vms):
+    """Make sure SD qubes are either disposable or have custom-persist enabled"""
+    for qube in sdw_tagged_vms:
+        if qube.klass == "DispVM":
+            continue
+
+        elif qube.klass == "AppVM" and qube.template_for_dispvms:
+            # Persistence is acceptable because disposable templates are not
+            # expected to be used directly. We could explore setting custom-persist,
+            # just to ensure no state is kept.
+            continue
+
+        elif qube.klass == "AppVM" and not qube.template_for_dispvms:
+            # 1. Custom persist must be enabled
+            assert qube.features.get("service.custom-persist", False)
+
+            # 2. Must have at least one entry, setting which files/dirs to persist
+            assert any(feat.startswith("custom-persist.") for feat in qube.features)
+
+        elif qube.klass == "TemplateVM":
+            # Everything in a template is expected to persist
+            continue
+
+        else:
+            pytest.fail(f"Qube of unexpected type: {qube.name}")
 
 
 @pytest.mark.provisioning
--- a/tests/test_vm_sd_viewer.py
+++ b/tests/test_vm_sd_viewer.py
@@ -5,7 +5,6 @@ specifically for the "sd-viewer" VM and related functionality.
 
 import subprocess
 
-import dnf
 import pytest
 from qubesadmin import Qubes
 
@@ -17,15 +16,12 @@ from tests.base import (
 from tests.base import (
     Test_SD_VM_Common as Test_SD_Viewer_Common,  # noqa: F401 [HACK: import so base tests run]
 )
+from tests.conftest import skip_on_qubes_4_2
 
 EXPECTED_N_PRELOADED_VMS = [
     1,  # 16GB system (less than recommended)
     2,  # 32GB+ system
 ]
-
-skip_on_qubes_4_2 = pytest.mark.skipif(
-    dnf.rpm.detect_releasever("/") == "4.2", reason="Feature only available in Qubes >= 4.3"
-)
 
 
 def _create_test_qube(dispvm_template_name):


securedrop_salt/sd-app.sls
```diff
@@ -45,6 +45,18 @@ sd-app:
...
       - sls: securedrop_salt.sd-viewer
 
+{% if grains['osrelease'] != '4.2' %}
+sd-app-custom-persist:
+  qvm.features:
+    - name: sd-app
+    - enable:
+      - service.custom-persist
+    - set:
+      - custom-persist.client_dir: /home/user/.securedrop_client/
```
deeplow, line 58: Since the client is going away, this is not too critical, but is there something here that we don't fully need?

securedrop_salt/sd-gpg.sls
```diff
@@ -45,3 +45,13 @@ sd-gpg:
...
+
+{% if grains['osrelease'] != '4.2' %}
+sd-gpg-custom-persist:
+  qvm.features:
+    - name: sd-gpg
+    - enable:
+      - service.custom-persist
+    - set:
+      - custom-persist.gnupg_dir: /home/user/.gnupg
+{% endif %}
```
deeplow, line 62: Because we stopped persisting `~/.profile`, we're essentially undoing the gpg prompt removal introduced in https://github.com/freedomofpress/securedrop-workstation/pull/1466. To work around this, we have two options:

### Option 1 
Include in this PR some on-boot step that adds `export QUBES_GPG_AUTOACCEPT=2147483647` to `~/.profile`.

To avoid a complementary client package, this implementation may share the functionality proposed for https://github.com/freedomofpress/securedrop-workstation/issues/1505.

### Option 2 (probably easier)

Add `export QUBES_GPG_AUTOACCEPT=2147483647` to `/etc/.profile` in the template, under `/etc/profile.d`.

This would ideally be done through a package as opposed to salt, but since it's it's in the template we're not incurring extra installation time. Furthermore, we can target only `sd-gpg` either by its name ([kind of how the client does](https://github.com/freedomofpress/securedrop-client/blob/0.17.0/client/files/sd-app-qubes-gpg-domain.sh)) to set's `sd-gpg` as its backend domain. But arguably preferably is to implement this by a qubes-service. This way, should we ever need to remove this behavior in `sd-gpg` we don't have to care about whether or not the file exists in the system (see problem described bellow), but simply remove the qubes service...

- **Via salt/ansible:** We'll forever need to care about this file in the template
- **Through package:** One more package, but we can change whatever we want (including remove that file). However, because the package is itself installed through salt, we get to the same exact problem ­should we no longer need it, we need to tell salt to remove it :/

securedrop_salt/sd-gpg.sls
```diff
@@ -45,3 +45,13 @@ sd-gpg:
...
       - sls: securedrop_salt.sd-upgrade-templates
+
+{% if grains['osrelease'] != '4.2' %}
+sd-gpg-custom-persist:
+  qvm.features:
+    - name: sd-gpg
+    - enable:
+      - service.custom-persist
+    - set:
+      - custom-persist.gnupg_dir: /home/user/.gnupg
```
deeplow, line 61: Can we make this more fine-grained? Or do we risk loosing important files?

legoktm: Once the app is default + we have the dom0 RPC to get the secret key the entire VM will be made disposable so I don't think we need to worry too much about this.

securedrop_salt/sd-gpg.sls
legoktm: For reference, we ended up going with freedomofpress/securedrop-client#3096.

securedrop_salt/sd-app.sls
legoktm: Same thought as GPG, let's just leave this as is for now and not invest time figuring out the client since it'll go away soon. Once the migration is done we'll wipe this folder and stop persisting it.

securedrop_salt/sd-log.sls
```diff
@@ -72,6 +72,16 @@ install-sd-log:
     - require:
       - qvm: sd-small-{{ sdvars.distribution }}-template
 
+{% if grains['osrelease'] != '4.2' %}
+sd-log-custom-persist:
+  qvm.features:
+    - name: sd-app
```
legoktm:
```suggestion
    - name: sd-log
```

legoktm:
> Confirm that no other sd-workstation-tagged remain with full persistence (except disposable templates — should we also add it to them?)

Actually could we add a test for this? Iterate through all the tagged VMs and verify that it's either a dispVM or a custom persist?

tests/test_vms_exist.py
```diff
@@ -33,6 +34,39 @@ def test_all_sdw_vms_present(all_vms, sdw_tagged_vms):
...
+            # just to ensure no state is kept.
+            pass
+
+        elif qube.klass == "AppVM" and not qube.template_for_dispvms:
+            # 1. Custom persist must be enabled
+            assert qube.features.get("service.custom-persist", False)
+
+            # 2. Must have at least one entry, setting which files/dirs to persist
+            has_custom_persist_entry = False
+            for feat in qube.features:
```
legoktm: Maybe `assert any(feat.startswith("custom-persist.") for feat in qube.features)`

tests/test_vms_exist.py
```diff
@@ -33,6 +34,39 @@ def test_all_sdw_vms_present(all_vms, sdw_tagged_vms):
         assert all_vms.get(vm_name) is None, f"Qube '{vm_name}' expected"
 
 
+@skip_on_qubes_4_2
+@pytest.mark.provisioning
+def test_expected_persistence(sdw_tagged_vms):
+    """Make sure SD qubes are either disposable or have custom-persist enabled"""
+    for qube in sdw_tagged_vms:
+        if qube.klass == "DispVM":
+            pass
```
legoktm: Can we use `continue` here (and in the other blocks) just to indicate that this is an acceptable state and we're done processing?

Just so it's easier to read, right now with `pass` I need to keep going down to understand the control flow.

deeplow: Makes sense. Thanks for the suggestion.

tests/test_vms_exist.py
deeplow: That's much nicer. I always forget about the `any`. Thanks!

tests/conftest.py
```diff
@@ -18,6 +18,10 @@
 
 PROJ_ROOT = Path(__file__).parent.parent
 
+skip_on_qubes_4_2 = pytest.mark.skipif(
```
legoktm, line 21: This is a little duplicate now with the qubes_ver() fixture but I'm going to leave it like this just to land it and we can DRY later as a follow-up.

Issue #1294, opened 2025-04-27
Preload sd-viewer


deeplow:
## Description

Preload a couple of sd-viewer qubes to speed up document viewing. Make use of the imminent https://github.com/QubesOS/qubes-issues/issues/1512.

## How will this impact [SecureDrop/SecureDrop Workstation users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?

Shorter time opening untrusted documents.

## How would this affect the SecureDrop Workstation [threat model](https://github.com/freedomofpress/securedrop-workstation/#threat-model)?

Should not affect

## User Stories

As a journalist I don't like waiting for documents to show up.

deeplow:
### Feature exploration

I'll update this comment with the details I find as following #1458.

**Must**:
- **preload X `sd-viewer` qubes** when enough RAM is available

**Should**:
- disable default disposable preloading (to save RAM)

PR #1557, opened 2026-02-02
Test preloading of sd-viewer qubes (4.3+)

### LINKED ISSUE #1541: Test preloaded disposables in 4.3
Test preloaded disposables working as correctly. Preloaded disposables came 'for free' in #1294, given that `sd-viewer` is the default disposable and default disposables get preloaded. But we have not yet implemented tests for it. This issue is about that.

deeplow:
In Qubes 4.3 sd-viewer [is preloaded by default](https://github.com/freedomofpress/securedrop-workstation/issues/1294) as a consequence of it being the default disposable (otherwise we could also make it explicitly get preloaded). However, we weren't yet testing for this.

Fixes #1541 

**Blocked by**:  #1373 (switch base to `main` after it is merged)

## Test plan

- [ ] Visual review
- [ ] Tests pass
- [ ]  Tests fault injection or other kind of meta-testing

## Checklist

This change accounts for:

### DIFF
--- a/tests/test_vm_sd_viewer.py
+++ b/tests/test_vm_sd_viewer.py
@@ -5,6 +5,7 @@ specifically for the "sd-viewer" VM and related functionality.
 
 import subprocess
 
+import dnf
 import pytest
 from qubesadmin import Qubes
 
@@ -15,6 +16,15 @@ from tests.base import (
 )
 from tests.base import (
     Test_SD_VM_Common as Test_SD_Viewer_Common,  # noqa: F401 [HACK: import so base tests run]
+)
+
+EXPECTED_N_PRELOADED_VMS = [
+    1,  # 16GB system (less than recommended)
+    2,  # 32GB+ system
+]
+
+skip_on_qubes_4_2 = pytest.mark.skipif(
+    dnf.rpm.detect_releasever("/") == "4.2", reason="Feature only available in Qubes >= 4.3"
 )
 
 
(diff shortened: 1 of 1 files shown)


tests/test_vm_sd_viewer.py
```diff
@@ -153,3 +159,29 @@ def test_sd_viewer_config(all_vms, dom0_config):
...
+    # sd-viewer as default_dispvm is makes it preload implicitly
+    assert Qubes().default_dispvm.name == "sd-viewer"
+
+    # Confirm expected number of preloaded qubes
+    preload_dispvm_max = int(all_vms["dom0"].features["preload-dispvm-max"])
+    assert preload_dispvm_max == max(EXPECTED_N_PRELOADED_VMS)
+
+
+@pytest.mark.skipif(
+    dnf.rpm.detect_releasever("/") == "4.2", reason="Feature only available in Qubes >= 4.3"
```
legoktm: nit: I would probably written some very basic parsing of `/etc/os-release` to figure out 4.2 vs 4.3 but this is perfectly fine!

tests/test_vm_sd_viewer.py
```diff
@@ -153,3 +159,29 @@ def test_sd_viewer_config(all_vms, dom0_config):
...
+
+
+@pytest.mark.provisioning
+def test_preloading_assumptions(all_vms):
+    # sd-viewer as default_dispvm is makes it preload implicitly
+    assert Qubes().default_dispvm.name == "sd-viewer"
+
+    # Confirm expected number of preloaded qubes
+    preload_dispvm_max = int(all_vms["dom0"].features["preload-dispvm-max"])
+    assert preload_dispvm_max == max(EXPECTED_N_PRELOADED_VMS)
```
legoktm: I don't think the max() logic here is right...`max([1,2])` will always be 2. But I happen to be testing on a 16GB machine, so there's only one preloaded VM.

Instead I'd suggest:

```suggestion
    assert preload_dispvm_max in EXPECTED_N_PRELOADED_VMS
```

deeplow: I don't know where I was going with that :/ Should be fixed now.

tests/test_vm_sd_viewer.py
deeplow: Probably overkill, you're right. I had this already somewhere in another PR, so I simply copied it here.

Issue #1507, opened 2025-12-01
spike: create sd-admin VM with admin tooling


conorsch:
The traditionally Tails-based SecureDrop Admin tooling was recently made available [as a Debian package](https://github.com/freedomofpress/securedrop/pull/7606), slated for [release in SecureDrop v2.13.0](https://github.com/freedomofpress/securedrop/issues/7721). Now that the admin tooling is available in apt repositories, we can reuse the tooling in Qubes contexts, too. Let's spike on prototyping an `sd-admin` VM to feel out the implementation.

We don't yet have a clear specification for the `sd-admin` requirements; the spike is intended to gather details that should inform a full spec. For example, do we want to require all orgs to provision a separate Qubes laptop for admin use? Should existing SDW laptops be optionally configurable to also serve as an Admin Workstation?

As described in #1396, we're in the early stages of feeling out Debian Trixie support. Notably we _don't_ want to commit to migrating all of the SDW VMs from Debian 12 Bookworm to Debian 13 Trixie, as that would require repackaging `securedrop-client` with a newer Python version. Given the leaner, `qubes-builder`-less approach described in #969, we should be able to `qvm-clone` the Qubes-provided `debian-13` template or its minimal variant and go from there. 

## Subtasks

The spike handle at least the following:

- [ ] Create a Trixie-based SDW template
- [ ] Add provisioning logic for `sd-admin` VM based on `debian-13`
- [ ] Install `securedrop-admin` tooling inside that VM via Salt or similar

## Stretch goals

In order to inform the full implementation, we also want to _use_ the tooling, e.g.:

- [ ] Use the `sd-admin` VM to configure remote SecureDrop servers
- [ ] Use the `sd-admin` VM to import pre-existing Tails configs.

deeplow:
> ## Subtasks
> 
> The spike handle at least the following:
> 
> * [ ]  Create a Trixie-based SDW template
> * [ ] Add provisioning logic for `sd-admin` VM based on `debian-13`\
> * [ ] Install `securedrop-admin` tooling inside that VM via Salt or similar

The base template will [soon be Debian 13](https://github.com/freedomofpress/securedrop-workstation/pull/1719#discussion_r3382027329), so we may need to add some optional integration as per initial spike comments (in code).


Post spike tasks:
- [ ] (from code comment) consider extracting shared apt config into a common location to avoid duplication with securedrop_salt/sd-default-config.sls.
[...]

PR #1572, opened 2026-02-17
spike: sd-admin VM via RPM

conorsch:
This PR adopts the "admin-config" RPM spec file by @zenmonkeykstop in #1554. It also pulls in the simplified RPM build logic from #1565; therefore, #1565 should be reviewed and merged first.

Based on @zenmonkeykstop's work, I added an `sd-admin` VM based on Debian 13 Trixie. There's no grsec kernel or keyring package yet, but the `securedrop-admin` tooling is installed:

<details>
<summary>
securedrop-admin help text
</summary>

```
[conor@dom0 securedrop-workstation]$ qvm-run -p sd-admin "securedrop-admin --help"
usage: securedrop-admin [-h] [-v] [-d] [--force]
                        {sdconfig,install,localconfig,tailsconfig,qubesconfig,generate_v3_keys,backup,restore,check_for_updates,logs,reset_admin_access,verify} ...

SecureDrop Admin Toolkit.

For use by administrators to install, maintain, and manage their SD
instances.

positional arguments:
  {sdconfig,install,localconfig,tailsconfig,qubesconfig,generate_v3_keys,backup,restore,check_for_updates,logs,reset_admin_access,verify}
    sdconfig            Configure SD site settings
    install             Install/Update SecureDrop
    localconfig (tailsconfig, qubesconfig)
                        Configure either Tails or Qubes environment post SD install
    generate_v3_keys    
                        This method will either read v3 Tor onion service keys if found or generate
                        a new public/private keypair.
    backup              Perform backup of the SecureDrop Application Server.
                        Creates a tarball of submissions and server config, and fetches
                        back to the Admin Workstation. Future `restore` actions can be performed
                        with the backup tarball.
    restore             Perform restore of the SecureDrop Application Server.
                        Requires a tarball of submissions and server config, created via
                        the `backup` action.
    check_for_updates   Check for SecureDrop updates
    logs                Get logs for forensics and debugging purposes
    reset_admin_access  Resets SSH access to the SecureDrop servers, locking it to
                        this Admin Workstation.
    verify              Run configuration tests against SecureDrop servers

options:
  -h, --help            show this help message and exit
  -v                    Increase verbosity on output (default: False)
  -d                    Developer mode. Not to be used in production. (default: False)
  --force               force command execution without update check (default: False)
[conor@dom0 securedrop-workstation]$ 
```
</details>

The RPM setup reuses Salt state files between the two RPMs, but installs them to unique dest paths, ensuring both RPMs can be installed independently of one another, without triggering dnf conflicts. We can lean into this and port more of the pre-existing Salt config if we desire. 

Towards #1507. 

## Test plan
1. Ensure that `make clone && make dev` passes without error
2. Run `make sd-admin` to configure the `sd-admin` VM; check that `securedrop-admin` CLI is installed within it.
3. Build the RPMs locally and run `rpm -qlp <path>` on them; inspect their file contents and ensure no conflicts or overlaps between them, because any duplication of dest paths will cause problems.

## Checklist

This change accounts for:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,6 +83,18 @@ build-rpm: ## Build RPM package
 	@echo
 	@echo "Build log available at $(OUT)"
 
+.PHONY: build-admin-rpm
+build-admin-rpm: OUT:=build-log/securedrop-admin-$(shell date +%Y%m%d).log
+build-admin-rpm: ## Build admin RPM package (opt-in)
+	@mkdir -p build-log
+	@echo "Building SecureDrop Admin RPM..."
+	@export TERM=dumb
+	@USE_BUILD_CONTAINER=true script \
+		--command "$(CONTAINER) ./scripts/build-admin-rpm.sh" \
+		--return $(OUT)
+	@echo
+	@echo "Build log available at $(OUT)"
+
 # FIXME: the time variations have been temporarily removed from reprotest
 # Suspecting upstream issues in rpm land is causing issues with 1 file\'s modification time not being clamped correctly only in a reprotest environment
 .PHONY: reprotest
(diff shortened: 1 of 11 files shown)


micahflee (dismissed):
This looks good to me. I haven't looked into why the GitHub checks aren't running/passing, and `main` needs to be merged into this branch still.

As far as the test plan:

1. `make clone && make dev` works without error
2. `make sd-admin` configures the `sd-admin` VM, and `securedrop-client` is installed and worked
3. after building the RPMs locally, I confirm that they don't have any conflicting files

After creating `sd-admin`, I also went ahead and used it to install a SecureDrop server, and it works! One thing we might want to put some thought into is what tags to put on `sd-admin`, because the lack of copy/paste and file transfer made it very difficult at first to copy my ssh key into `sd-admin`. There might be various legit reasons the admin needs to use copy/paste and to transfer files.

legoktm (dismissed):
I'm slightly concerned about having this be enabled by default just because in the past we've worked hard to avoid having things that get built but intentionally discarded; we're at the point that it should be a blind thing that you just copy all the packages that get built to apt/yum repo (especially once we have CI do it) and I'd like to preserve that.

RPM makes it easier to have [conditional builds](https://rpm.org/docs/latest/manual/conditionalbuilds.html), could we try adding `%bcond_with admin` so that builds need to have `--with=admin` to include that RPM? And then a separate `make build-rpm-admin` target.

conorsch:
Heard. I didn't get to this today, but I'll adjust the logic to ensure that the new RPM is off-by-default for both building and installing.

conorsch:
Alright, I rewrote the RPM build logic to use `%bcond`. It's a significant enough change to the PR to require a full re-review.

There's still a bit of non-DRY duplication in the shell scripts that build the RPM, but I feel that's warranted here during the WIP phase, to ensure that the pre-existing logic for building the `securedrop-workstation-dom0-config` remains completely separate from the new logic for `securedrop-admin-dom0-config`. If folks disagree, we could easily toggle the `--with admin` flag to be conditional on an env var. Given the state of the spike, I suggest we punt on reconciling the duplication until we're ready to build and host the package somewhere like yum-test.

Ready for re-review, @micahflee!

micahflee (approved):
I tested again and everything works great. I built both RPMs and compared them, and they still don't have any overlapping files.

Issue #1499, opened 2025-11-25
SaltStack `cmd.run` doesn't get `args` passed along (Qubes 4.3)


deeplow:
## Description

*(necessary for https://github.com/freedomofpress/securedrop-workstation/issues/1475)*

When running "sdw-admin --apply" "run-prep-upgrade-scripts" in "sd-upgrade-templates.sls" and "sd-remove-unused-templates" as it fails with the message:

```    
stdout: Please specify prepare or remove
```

This indicated that the `args` parameter is not being properly recognized. I looked at the SaltStack documentation for the `cmd` module and nothing indicates an 'args' parameters change.

I've checked with Salt's changelogs as well comparing the version in Qubes 4.2 with the one in Qubes 4.3 to no avail. Instead of investigating how to get "cmd.script" this simply uses another way to call the command as a temporary fix.

## Steps to Reproduce

Try to deploy SecureDrop Workstation in Qubes 4.3, but essentially it failed in 

## Expected Behavior

`sdw-admin --apply` works

## Actual Behavior

I had made a commit to fix it and I'm only now trying to document it as a separate issue / PR from #1475. So I don't have too much context, but the text in the above description was captured in original commit where I fixed this.

## Comments

For now we just need a workaround with `cmd.script` instead of `cmd.run` as it doesn't suffer from missing args. For a more permanent solution we need to fix this in all different places and if we choose to go the 'cmd.run' route, we should place the script in some place other than `/srv/salt/`.

legoktm:
This landed in https://github.com/freedomofpress/securedrop-workstation/pull/1373.

PR #1582, opened 2026-02-27
Set xfce-exe-checksum on desktop launcher

### LINKED ISSUE #1577: Desktop launcher is untrusted in Qubes 4.3

## Description

When you try to click on the desktop icon, it warns you that the application is untrusted and you can either: 1) Launch anyways, 2) Mark executable as trusted or 3) Cancel.

This appears to be a new XFCE "security" feature per https://forum.xfce.org/viewtopic.php?id=16357

We already set mode +x, but then:
> it needs to have an xfce-exe-checksum metadata entry created for the file:
> `f=FILE; gio set -t string $f metadata::xfce-exe-checksum "$(sha256sum $f | awk '{print $1}')"`

Alternatively per https://forum.xfce.org/viewtopic.php?id=16407 if we stick the file in a $XDG_DATA_DIRS (`/usr/share`) and then symlink it onto the desktop, then we don't need to deal with the special checksum trust stuff.

legoktm:
XFCE has a new "security" feature that requires us to set a checksum of the .desktop file in the gio database before it can be used.

A slightly overkill script sets the value and also offers a --check mode so that way salt can skip setting it when not needed.

Fixes #1577.

## Test plan

* [ ] Run `make dev` with this PR, verify you can click on the launcher and it doesn't have a prompt about it being untrusted.

## Checklist

This change accounts for:

### DIFF
--- a/files/update-xfce-settings
+++ b/files/update-xfce-settings
@@ -57,6 +57,14 @@ elif [[ $TASK == "adjust-icon-size" ]]; then
   echo "update-xfce-settings: Adjusting icon size for user $USER to $ICONSIZE px"
   xfconf-query -c xfce4-desktop -np '/desktop-icons/icon-size' -t 'int' -s $ICONSIZE
 
+elif [[ $TASK == "dismiss-desktop-icon-prompt" ]]; then
+  # Dismiss "security warning" when opening any desktop icons
+  # as it is not a true security mitigation in Qubes. See #1582
+  for f in "/home/${USER:?}/Desktop/"*.desktop; do
+    chmod +x "$f";
+    gio set -t string "$f" metadata::xfce-exe-checksum "$(sha256sum "$f" | awk '{print $1}')";
+  done
+
 elif [[ $TASK == "reset-power-management" ]]; then
   echo "update-xfce-settings: Resetting power management options for user $USER"
 
@@ -86,8 +94,9 @@ else
   echo
   echo "Task must be one of:"
   echo
-  echo " disable-unsafe-power-management   Disable suspend and hibernation buttons"
   echo " adjust-icon-size                  Increase the default desktop icon size"
-  echo " reset-power-management            Reset power management settings to default"
+  echo " disable-unsafe-power-management   Disable suspend and hibernation buttons"
+  echo " dismiss-desktop-icon-prompt       Mark the desktop icon for SDW as trusted"
   echo " reset-icon-size                   Reset icon size to default"
+  echo " reset-power-management            Reset power management settings to default"
 fi
--- a/securedrop_salt/sd-dom0-files.sls
+++ b/securedrop_salt/sd-dom0-files.sls
@@ -48,6 +48,17 @@ dom0-securedrop-launcher-desktop-shortcut:
     - group: {{ gui_user }}
     - mode: 755
 
+{% if grains['osrelease'] != '4.2' %}
+# Dismiss "security warning" when opening any desktop icons
+# as it is not a true security mitigation in Qubes. See #1582
+dom0-dismiss-desktop-icon-prompt:
+  cmd.run:
+    - name: /usr/bin/securedrop/update-xfce-settings dismiss-desktop-icon-prompt
+    - runas: {{ gui_user }}
+    - require:
+      - file: dom0-securedrop-launcher-desktop-shortcut
+{% endif %}
+
 dom0-environment-directory:
   file.directory:
     - name: /var/lib/securedrop-workstation/
--- a/tests/test_dom0_salt_config.py
+++ b/tests/test_dom0_salt_config.py
@@ -1,6 +1,17 @@
+import hashlib
+import os
 import subprocess
+from pathlib import Path
 
 import pytest
+
+from tests.conftest import skip_on_qubes_4_2
+
+DESKTOP_FILE_NAME = "press.freedom.SecureDropUpdater.desktop"
+
+
+def checksum(path: Path) -> str:
+    return hashlib.sha256(path.read_bytes()).hexdigest()
 
 
 def test_is_topfile_enabled():
@@ -13,3 +24,25 @@ def test_is_topfile_enabled():
 
     except subprocess.CalledProcessError:
         pytest.fail("Error checking topfiles")
+
+
+@skip_on_qubes_4_2
+@pytest.mark.provisioning
+def test_trust_desktop_launcher() -> None:
+    desktop_file = Path.home() / "Desktop" / DESKTOP_FILE_NAME
+
+    # Duplicate the dbus session bootstrap logic from `../files/update-xfce-settings`,
+    # so the test works under invocations that don't inherit a session bus,
+    # such as OpenQA's `su user -c`.
+    env = {**os.environ, "DBUS_SESSION_BUS_ADDRESS": f"unix:path=/run/user/{os.getuid()}/bus"}
+
+    # salt should have already provisioned the correct checksum
+    res = subprocess.run(
+        ["gio", "info", "--attributes", "metadata::xfce-exe-checksum", desktop_file],
+        text=True,
+        check=False,
+        capture_output=True,
+        env=env,
+    )
+    assert res.returncode == 0, "desktop launcher does not match"
+    assert checksum(desktop_file) in res.stdout


deeplow (dismissed):
A bit of a temporary overkill, but it works. I think the "overkillness" of it is proportional to the "overkillness" of this "security" feature. LGTM.

Fun fact: while reviewing this I noticed that `sd-dom0-files.sls` is actually running twice (I submitted a fix [here](https://github.com/freedomofpress/securedrop-workstation/pull/1586)). In this case the first time it would actually set up the file first and the second it would run with `--check`, thus testing the full script unintentionally 🙂.

Let's wait on OpenQA, but from the review side, all is good.

deeplow:
> Hmm, I didn't realize OpenQA was running pytest as root. Let me see, I suppose we can have the test conditionally run it as gui_user.

This was not intentionally, I'd prefer if we fixed it OpenQA code as running tests as root is likely to have other unintended consequences. I have just opened [a draft PR for that](https://github.com/freedomofpress/openqa-tests-qubesos/pull/14).

deeplow (commented):
When I agreed with this direction, I did not expect it to be this complex, TBH.  I'm a little hesitant about the approach due to the unnecessary complexity and I was about to suggest reverting course to the symlink approach, but apparently [that doesn't work](https://github.com/freedomofpress/securedrop-workstation/issues/1577) in dom0's XFCE version.

**This means that we're stuck with this for the entirety of Qubes 4.3**, so we should have a more maintainable solution. Alternatively we could send a fix / workaround to Qubes itself.

General feedback:
1. After https://github.com/freedomofpress/openqa-tests-qubesos/pull/14 gets merged, the last commit should be reverted
1. If not too much trouble, what do you think about doing this in  `update-xfce-settings` (in bash)? It's one less file to manage and this fits under the theme. TBH, I think just making all files as trusted [like someone suggested](https://forum.xfce.org/viewtopic.php?pid=70923#p70923):
    
    ```bash
    for f in ~/Desktop/*.desktop; do
      chmod +x "$f";
      gio set -t string "$f" metadata::xfce-exe-checksum "$(sha256sum "$f" | awk '{print $1}')";
    done
    ```

   This would save the need to figure out what broke, if we ever change the file's name.
3. If possible removing DBUS_SESSION by just inheriting the environment when calling the script.
4. Not sure it needs the `--check` because it's only really useful in the test, in which case, we might as we just put the `gio info...` direct in the test itself.

---

An alternative approach could be to have a `systemd --user` path+service, which is triggered when anything in the desktop file is changed. The path portion of it would be as simple as:

```ini
[Unit]
Description=Watch ~/Desktop for changes
[Path]
PathChanged=%h/Desktop
[Install]
WantedBy=default.target
```

And then the user service would simply run the "trust all desktop files".

conorsch:
I've fixed the shellcheck lint, and now have a 4.3 machine to test this on interactively. Will proceed with final, functional review.

conorsch:
Tested this on a recently-upgraded-to-4.3 machine, and it worked well. Specifically, I installed SDW via `make clone && make dev` from the `main` branch, and confirmed that double-clicking the SDW launcher icon on the Xfce desktop generated the scary message documented in #1577. Then I switched to this branch, reran `make clone && make dev`, and tried again, and confirmed that the problem was resolved: the launcher launched, prompting for updates, with no interstitial message about untrusted apps.

That sounds great, and validates the primary purpose of this PR: make sure the scary warning doesn't appear for users on QubesOS 4.3.

However, there are [CI failures in the OpenQA suite](https://openqa.qubes-os.org/tests/175209/logfile?filename=serial_terminal.txt), one of which lists the checksum under test here: 

```
    @skip_on_qubes_4_2
    @pytest.mark.provisioning
    def test_trust_desktop_launcher():
        desktop_file = Path.home() / "Desktop" / DESKTOP_FILE_NAME
    
        # salt should have already provisioned the correct checksum
        res = subprocess.run(
            ["gio", "info", "--attributes", "metadata::xfce-exe-checksum", desktop_file],
            text=True,
            check=False,
            capture_output=True,
        )
        assert res.returncode == 0, "desktop launcher does not match"
>       assert checksum(desktop_file) in res.stdout
E       AssertionError: assert 'a9b06f96fba5b80da74cc51e067b56f6cc88d694039d99259ed4e0d175cee762' in 'uri: file:///home/user/Desktop/press.freedom.SecureDropUpdater.desktop\nlocal path: /home/user/Desktop/press.freedom.SecureDropUpdater.desktop\nunix mount: /dev/mapper/qubes_dom0-root / ext4 rw,relatime,discard,stripe=16\nattributes:\n'
E        +  where 'a9b06f96fba5b80da74cc51e067b56f6cc88d694039d99259ed4e0d175cee762' = checksum(PosixPath('/home/user/Desktop/press.freedom.SecureDropUpdater.desktop'))
E        +  and   'uri: file:///home/user/Desktop/press.freedom.SecureDropUpdater.desktop\nlocal path: /home/user/Desktop/press.freedom.SecureDropUpdater.desktop\nunix mount: /dev/mapper/qubes_dom0-root / ext4 rw,relatime,discard,stripe=16\nattributes:\n' = CompletedProcess(args=['gio', 'info', '--attributes', 'metadata::xfce-exe-checksum', PosixPath('/home/user/Desktop/press.freedom.SecureDropUpdater.desktop')], returncode=0, stdout='uri: file:///home/user/Desktop/press.freedom.SecureDropUpdater.desktop\nlocal path: /home/user/Desktop/press.freedom.SecureDropUpdater.desktop\nunix mount: /dev/mapper/qubes_dom0-root / ext4 rw,relatime,discard,stripe=16\nattributes:\n', stderr='').stdout
tests/test_dom0_salt_config.py:41: AssertionError
```

AFAICT that means that the `update-xfce-settings` additions aren't being executed properly in the OpenQA environment. Presumably that's because we assume, but don't validate, that we have an active `DBUS_SESSION_BUS_ADDRESS` for the script to talk to.

So what should we do about that? The fact that we wrote the test and it failed shows us that our code makes assumptions, which is a really good thing to know. Real-world users, we should note, will almost certainly have a GUI environment that bootstraps a dbus session, so the update-launcher code will work—but we're testing it wrong. I see two options:

1. we figure out how to bootstrap a dbus session in the OpenQA environment, so it's more prod-like; or 
2. we rework the update-launcher checksum logic, e.g. to use systemd services in dom0, which would allow us to declare explicit dependencies on `dbus.socket` or similar

For now, Option 1 would be my preference. @deeplow what do you think of my analysis here? Do you agree about the explanation of the failure of the OpenQA environment? FYI I've rebased yet again, which will ensure another run.

deeplow:
> For now, Option 1 would be my preference. @deeplow what do you think of my analysis here?

I think I know exactly what it is: https://github.com/freedomofpress/securedrop-workstation/issues/1499 :cry: 

Easy fix. But a rather annoying / dangerous bug in SaltStack. Basically in `cmd.script` ignores `args:`.

files/update-xfce-settings
```diff
@@ -57,6 +57,14 @@ elif [[ $TASK == "adjust-icon-size" ]]; then
   echo "update-xfce-settings: Adjusting icon size for user $USER to $ICONSIZE px"
   xfconf-query -c xfce4-desktop -np '/desktop-icons/icon-size' -t 'int' -s $ICONSIZE
 
+elif [[ $TASK == "dismiss-desktop-icon-prompt" ]]; then
```
deeplow, line 60: I also need to add this entry to the instructions (at the bottom of the script).

deeplow:
Apparently that wasn't the only issue. I'm missing a 4.3 system to test this right now. I've added `+x` to the script to see if we get some output on OpenQA. 

**Update** — here is what we got in sdw-admin --apply on [OpenQA test 175786](https://openqa.qubes-os.org/tests/175786#step/install_workstation/61):

```
----------
          ID: dom0-dismiss-desktop-icon-prompt
    Function: cmd.run
        Name: /usr/bin/securedrop/update-xfce-settings dismiss-desktop-icon-prompt
      Result: True
     Comment: Command "/usr/bin/securedrop/update-xfce-settings dismiss-desktop-icon-prompt" run
     Started: 10:32:29.011298
    Duration: 818.252 ms
     Changes:   
              ----------
              pid:
                  8164
              retcode:
                  0
              stderr:
                  + [[ 1000 -eq 0 ]]
                  + TASK=dismiss-desktop-icon-prompt
                  + ICONSIZE=64
                  ++ command -v xfconf-query
                  + '[' -x /usr/bin/xfconf-query ']'
                  ++ id -u user
                  + DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
                  + export DBUS_SESSION_BUS_ADDRESS
                  + [[ dismiss-desktop-icon-prompt == \d\i\s\a\b\l\e\-\u\n\s\a\f\e\-\p\o\w\e\r\-\m\a\n\a\g\e\m\e\n\t ]]
                  + [[ dismiss-desktop-icon-prompt == \a\d\j\u\s\t\-\i\c\o\n\-\s\i\z\e ]]
                  + [[ dismiss-desktop-icon-prompt == \d\i\s\m\i\s\s\-\d\e\s\k\t\o\p\-\i\c\o\n\-\p\r\o\m\p\t ]]
                  + for f in "/home/${USER:?}/Desktop/"*.desktop
                  + chmod +x /home/user/Desktop/press.freedom.SecureDropUpdater.desktop
                  ++ awk '{print $1}'
                  ++ sha256sum /home/user/Desktop/press.freedom.SecureDropUpdater.desktop
                  + gio set -t string /home/user/Desktop/press.freedom.SecureDropUpdater.desktop metadata::xfce-exe-checksum a9b06f96fba5b80da74cc51e067b56f6cc88d694039d99259ed4e0d175cee762
              stdout:
----------
```

deeplow:
> AFAICT that means that the `update-xfce-settings` additions aren't being executed properly in the OpenQA environment. Presumably that's because we assume, but don't validate, that we have an active `DBUS_SESSION_BUS_ADDRESS` for the script to talk to.

I think I know what it was and you were almost there. I believe they are being properly set, but the issue is that the test can't access it. Here's an experiment to demonstrate that:


As you can see if the testing doesn't have the same DBUS Session, then it can't access these metadata properties:

```
[root@dom0 ~]# gio info --attributes metadata::xfce-exe-checksum /home/user/Desktop/press.freedom.SecureDropUpdater3.desktop 
uri: file:///home/user/Desktop/press.freedom.SecureDropUpdater3.desktop
local path: /home/user/Desktop/press.freedom.SecureDropUpdater3.desktop
unix mount: /dev/mapper/qubes_dom0-root / ext4 rw,relatime,discard,stripe=16
attributes:

[root@dom0 ~]#   # COMMENT: as you can see, above it can't really find the attributes. But then I exited the root shell and (Ctrl+D)
logout
[user@dom0 securedrop-workstation]$ gio info --attributes metadata::xfce-exe-checksum ~/Desktop/press.freedom.SecureDropUpdater3.desktop 
uri: file:///home/user/Desktop/press.freedom.SecureDropUpdater3.desktop
local path: /home/user/Desktop/press.freedom.SecureDropUpdater3.desktop
unix mount: /dev/mapper/qubes_dom0-root / ext4 rw,relatime,discard,stripe=16
attributes:
  metadata::xfce-exe-checksum: 386ec254f715d3d9c4745e2aa98b8c4b3134b14092e437f33c5604e80f0df290
```
And then it worked.

Let's see this applied to tests:

```
# All of this was done on a root shell already, because on OpenQA we're running from a root shell like so:
#      su user -c "env XAUTHORITY=/run/lightdm/user/xauthority DISPLAY=:0.0 CI=true make -C /home/user/securedrop-workstation test"

# The following passes
sudo -u user DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus python3 -m pytest tests/test_dom0_salt_config.py::test_trust_desktop_launcher

# The following FAILS
# sudo -u user python3 -m pytest tests/test_dom0_salt_config.py::test_trust_desktop_launcher
```

deeplow:
I don't think I'll be able to wrap this up. But I think we're on the right track.

conorsch:
Added some more work, waiting for results in OpenQA before writing summary and deciding on next steps, if any.

conorsch (approved):

deeplow:
Both of [4.2](https://openqa.qubes-os.org/tests/177118) and [4.3](https://openqa.qubes-os.org/tests/177121)'s `test_dom0` failed with xenlight:

I have restarted just the `test_dom0` portion of the tests:
- 4.2: https://openqa.qubes-os.org/tests/177225
- 4.3: https://openqa.qubes-os.org/tests/177226

The same happened in another PR and restarting the test solved it. https://github.com/freedomofpress/securedrop-workstation/pull/1638#issuecomment-4387924911

deeplow:
The relevant portion passed. Thanks for the test fixup. Previously I had tried the same thing with a env-related fixture pytest has and it didn't work. But yours did.

deeplow:
Just force-pushed to sign the code.

Issue #1573, opened 2026-02-17
Installation fails and shows no output in step: Configure all SecureDrop Workstation VMs


deeplow:
I have seen some tests where salt just fails and no output is shown. What's going on in these cases? Why is there no output?

This [OpenQA test](https://openqa.qubes-os.org/tests/167460#step/install_workstation/56) is an example:

<img width="1287" height="650" alt="Image" src="https://github.com/user-attachments/assets/19edafcc-7081-4ec3-bea6-b9f975ee10dd" />

The last line from stdout is (the stderr was not captured textually but it's in the image above)

```
..........................................................................
Configure all SecureDrop Workstation VMs with service-specific configs
Running "sudo qubesctl --show-output --skip-dom0 --max-concurrency 2 --targets sd-app,sd-base-bookworm-template,sd-devices,sd-devices-dvm,sd-gpg,sd-large-bookworm-template,sd-log,sd-proxy,sd-proxy-dvm,sd-small-bookworm-template,sd-viewer state.highstate"
```

Here are some files for preservation:
- [install_workstation-xl-dmesg(1).log](https://github.com/user-attachments/files/25373611/install_workstation-xl-dmesg.1.log)
- [install_workstation-var_log.tar.gz](https://github.com/user-attachments/files/25373612/install_workstation-var_log.tar.gz)
- [install_workstation-sdw-admin-apply(16).log](https://github.com/user-attachments/files/25373614/install_workstation-sdw-admin-apply.16.log)
- [install_workstation-libxl-driver.log](https://github.com/user-attachments/files/25373613/install_workstation-libxl-driver.log)
- [install_workstation-journalctl.log](https://github.com/user-attachments/files/25373615/install_workstation-journalctl.log)

deeplow:
I have observed this once again in https://github.com/freedomofpress/securedrop-workstation/issues/1644.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/795b6a4c-1029-48f5-b838-8785002d1fbd" />

Here are some logs:
- [install_workstation-journalctl(3).log](https://github.com/user-attachments/files/27392686/install_workstation-journalctl.3.log)
- [install_workstation-var_log(3).tar.gz](https://github.com/user-attachments/files/27392696/install_workstation-var_log.3.tar.gz)
- [install_workstation-sdw-admin-apply(28).log](https://github.com/user-attachments/files/27392711/install_workstation-sdw-admin-apply.28.log)

[...]

Issue #1358, opened 2025-07-04
Optimize OpenQA Run Time


deeplow:
The current [OpenQA tests](https://github.com/QubesOS/openqa-tests-qubesos/tree/f2dcbfd/tests/securedrop) take around 3h to run. Yes, slow nested virtualization is a factor, but we can't really change that. What we can do it optimize the script in various ways. The goal should be to have the base test run in about 1h.

A good thing to do is to look at the actual execussion time to see if there is anything we can save

Possible optimizations:
- [ ] pre-download `debian-12-minimal` <strike>and `fedora-42-xfce` (if not there by default)</strike>
  - [in progress] Marek can will set up a pre-setup qubes disk where debian-12-minimal is present. Currently waiting on PR to be finished https://github.com/freedomofpress/openqa-tests-qubesos/pull/4. Ping Marek afterwards to create the pre-caching job.
- [ ] Move all `assert_script_run` calls to actual scripts or to `root_console` so that no time is spent by the runner typing actual text key-by-key
- [ ] Remove whonix setup parts after #456
- [ ] Look for potentially unecessary steps
- [ ] move to an "upgrade scenario" - start with a pre-deployed "last-release" workstation and just run `sdw-admin --apply` with the new code

legoktm:
I suspect this is because by default podman is rootless, which means the image is stored in ~/.local (persistent), while docker is rootful and stores its images in `/var/lib/docker`, which isn't saved.

Since we only do the build once, a `podman rmi <image name>` would avoid needing to save it.

legoktm:
Yes, that was my proposal in https://github.com/freedomofpress/securedrop-workstation/issues/1328. I can put together a PoC...

deeplow:
> * Update all templates (Use `DO_UPDATE=1` in OpenQA parameters)

I have now added a step to also update because the template we're using comes pretty bare and without updates after the template is installed (to confirm, look at the OpenQA job that creates `install_minimal_unecrypted`; see [example](https://openqa.qubes-os.org/tests/175497)).

PR #1583, opened 2026-02-27
Set fail-fast: false on OpenQA jobs

### LINKED ISSUE #1581: Restarting 4.2 OpenQA job, when 4.3 already failed, fails immediately
Both `OpenQA 4.3` and `OpenQA 4.2` jobs had failed, but I only wanted to restart the 4.2 one. However, when doing so, it immediately failed with the message: `The strategy configuration was canceled because "openqa._4_3" failed`.

I believe the issue here is that we need `continue-on-error: true` when calling `openqa:` on `openqa-pull-request.yml`.

legoktm:
If e.g. 4.2 fails, we don't want to automatically kill the 4.3 job.

Fixes #1581.

## Test plan

Could test by intentionally breaking 4.2 and seeing that the 4.3 job continues?
## Checklist

This change accounts for:

### DIFF
--- a/.github/workflows/nightlies.yml
+++ b/.github/workflows/nightlies.yml
@@ -119,6 +119,7 @@ jobs:
     secrets: inherit
     needs: ["get-main-commit"]
     strategy:
+      fail-fast: false
       matrix:
         qubes_version: ["4.2", "4.3"]
     with:
--- a/.github/workflows/openqa-pull-request.yml
+++ b/.github/workflows/openqa-pull-request.yml
@@ -62,6 +62,7 @@ jobs:
     needs: [check]
     if: ${{ needs.check.outputs.qubes-versions != '' }}
     strategy:
+      fail-fast: false
       matrix:
         qubes_version: ${{ fromJSON(needs.check.outputs.qubes-versions) }}
     uses: ./.github/workflows/openqa.yml
--- a/.github/workflows/openqa-pull-request.yml
+++ b/.github/workflows/openqa-pull-request.yml
@@ -5,9 +5,16 @@ run-name: "[OpenQA] ${{ github.event.action == 'synchronize' && 'synchronize' ||
 on:
   pull_request:
     types:
-      - labeled    # Trigger immediately when label added
-      - unlabeled  # Detect label removal in order to cancel previous runs
-      - synchronize
+      - labeled     # Trigger immediately when label added
+      - unlabeled   # Detect label removal in order to cancel previous runs
+      - synchronize # Re-run against new HEAD when commits are pushed to the PR
+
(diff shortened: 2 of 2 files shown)


conorsch:
I resurrected this PR, and appended a small fix, adding a `concurrency` config block to the CI workflows for the PR variant of the OpenQA jobs. I'm pretty sure that'll resolve the slightly haunted behavior that @deeplow mentioned in https://github.com/freedomofpress/securedrop-workstation/pull/1583#issuecomment-3984467798. 

Note that the OpenQA jobs are currently red, pending known needed fixes like the Fedora 42 EOL (refs #1638).

conorsch (approved):

deeplow:
Thanks for looking at this one. This looks good to me and the other day I caught a test #1638 where this was a problem and the original commit would have solved.

> Note that the OpenQA jobs are currently red, pending known needed fixes like the Fedora 42 EOL (refs https://github.com/freedomofpress/securedrop-workstation/pull/1638).

While this is partly true, there was something odd with the results. If you look at the 4.3 section the `securedrop_install` also failed, which means something else was going on:

Screenshot 2026-05-05 at 09-20-53 Set fail-fast false on OpenQA jobs by legoktm · Pull Request #1583 · freedomofpress_securedrop-workstation

In the process of debugging, I filed / updated a couple of issues:
- https://github.com/freedomofpress/securedrop-workstation/issues/1644
- https://github.com/freedomofpress/securedrop-workstation/issues/1573
   :warning: this is a rather bad one. I couldn't get to the bottom of this because of the lack of output. All seemed good with the VMs.
- https://github.com/freedomofpress/securedrop-workstation/issues/1358#issuecomment-4377961227

deeplow (approved):
I can't really confirm the result because now the tests are failing, but the last changes look visually OK and not merging means sometimes the need to re-run things. Therefore I'm approving (echoing @conorsch's approval) and merging it.

deeplow:
(force-pushed to re-sign code)

deeplow:
The [last nightlies](https://github.com/freedomofpress/securedrop-workstation/actions/runs/25361918579/job/74363281915) actually failed like this as well. So we really need this one now that the CI is in a degraded state.

deeplow:
I'll just wait for these two to confirm if one no longer cancels the other (since both are likely to fail).

Screenshot 2026-05-05 at 10-42-24 Set fail-fast false on OpenQA jobs by legoktm · Pull Request #1583 · freedomofpress_securedrop-workstation

**Update**: this was taking too long and would be helpful already in https://github.com/freedomofpress/securedrop-workstation/pull/1638. So I'll merge it and we revert if we have other issues.

PR #1586, opened 2026-03-02
Salt: avoid running sd-dom0-files.sls twice

### LINKED ISSUE #1585: `sd-dom0-files.sls` is ran twice
`sd-dom0-files.sls` is ran twice:
 1. [when provisioning](https://github.com/freedomofpress/securedrop-workstation/blob/8627a29/files/sdw-admin.py#L99) the base template
 2. [when provisioning everything else](https://github.com/freedomofpress/securedrop-workstation/blob/8627a29/files/sdw-admin.py#L101) (via `state.highstate`)

This is not really critical because SaltStack is idempotent, but it is not necessary.

The reason this happens is because [sd-base-template.sls](https://github.com/freedomofpress/securedrop-workstation/blob/main/securedrop_salt/sd-base-template.sls#L7-L8) includes the dom0 configuration formula, just [to have the debian-12 template installed](https://github.com/freedomofpress/securedrop-workstation/blob/00b0c88/securedrop_salt/sd-dom0-files.sls#L8C1-L12C30). We should move those two together.

deeplow:
Since we're running the salt formulas in bits, SaltStack cannot guarantee that formulas run only once. In this case installing the template was in sd-dom0-files whereas in practice it should be out of that file, since it's only really ever used when creating the sd-base template.

Fixes #1585

## Test plan

- [ ] Start from a clean state
  - [ ] run `sdw-admin --uninstall`
  - [ ] run `qvm-template remove debian-12-minimal`
- [ ] Run `sdw-admin --apply` and look in the output for repeated salt states from `sd-dom0-files.sls` (each should only be run once)
- [ ] `sd-base-bookworm-template` is provisioned correctly.

## Checklist

This change accounts for:

### DIFF
--- a/securedrop_salt/sd-base-template.sls
+++ b/securedrop_salt/sd-base-template.sls
@@ -4,8 +4,10 @@
 # Imports "sdvars" for environment config
 {% from 'securedrop_salt/sd-default-config.sls' import sdvars with context %}
 
-include:
-  - securedrop_salt.sd-dom0-files
+# Ensure debian-12-minimal is present for use as base template
+dom0-install-debian-minimal-template:
+  qvm.template_installed:
+    - name: debian-12-minimal
 
 # Clones a base templateVM from debian-12-minimal
 sd-base-template:
--- a/securedrop_salt/sd-dom0-files.sls
+++ b/securedrop_salt/sd-dom0-files.sls
@@ -6,11 +6,6 @@
 # over time. These scripts should be ported to an RPM package.
 ##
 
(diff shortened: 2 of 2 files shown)


conorsch (approved):
Works great. Tested locally. Let's get it in!

conorsch:
Rebased, to include merge of #1638 , so we should expect CI to pass fully now.

conorsch:
> we should expect CI to pass fully now.

CI passing!

PR #1612, opened 2026-04-09
SecureDrop Workstation 1.6.0-rc2

legoktm:
## Test plan

* [x] base is release/1.6.0
* [x] changelog looks good

## Checklist

This change accounts for:


changelog.md
```diff
@@ -1,5 +1,11 @@
 # changelog
 
+## 1.6.0rc2
+
+* Split 'sd-devices' functionality into 'sd-printers'
```
deeplow:
```suggestion
```

We can remove this line, since we backported the 1.5.2 changelog as well. See line [38 of the changelog](https://github.com/freedomofpress/securedrop-workstation/pull/1612/changes#diff-3bd14d078188074c410028847113ceae68865d0ad5b844a27183ef87fbe2fcc3R33-R38).

changelog.md
```diff
@@ -1,5 +1,11 @@
...
+* Always provision the Inbox
```
deeplow, line 10: If I understand correctly, this will be enabled by default now. Should we make ti a bit more explicit that this is the case here already? I know we'll compact the RCs into a final changelog, but then we may overlook this detail.

deeplow (changes requested):

changelog.md
legoktm: Oops!

changelog.md
legoktm: Sure

deeplow (approved):
Thanks!

PR #1622, opened 2026-04-22
Fix sdw-notify Qt5 compat; always launch Inbox

### LINKED ISSUE #1621: sdw-notify issues for 1.6.0
1. @deeplow noticed that `sdw-notify` is missing the necessary launch_app, so it would instead launch the Client
2. When I started poking at it, I realized that it's been entirely broken in 4.2 because `QMessage.StandardButton.Warning` doesn't exist in Qt5.

legoktm:
The `sdw-notify` background timer that reminds users if updates haven't run recently should always default to launching the Inbox. This is the only remaining place where UpdaterApp was invoked without an explicit `launch_app`.

Additionally, `sdw-notify` was crashing on 4.2 because `QMessageBox.StandardButton.Warning` doesn't exist in Qt5.

Fixes #1621.

## Test plan

* [x] Build RPM and install in Qubes 4.2
* [x] in dom0, `rm ~/.securedrop_updater/sdw-last-updated`
* [x] in dom0, run `sdw-notify`, observe a dialog opens
* [x] when you click "Check for updates", the text should say "...use the SecureDrop **Inbox**" and not Client.
* [ ] (bonus) repeat for Qubes 4.3

## Checklist

This change accounts for:

### DIFF
--- a/files/sdw-notify.py
+++ b/files/sdw-notify.py
@@ -67,7 +67,7 @@ def show_update_warning():
     # has opted to check for updates.
     if result == NotifyApp.NotifyStatus.CHECK_UPDATES:
         log.info("Launching Preflight Updater")
-        updater = UpdaterApp.UpdaterApp()
+        updater = UpdaterApp.UpdaterApp(launch_app=True)
         updater.show()
         sys.exit(app.exec())
     elif result == NotifyApp.NotifyStatus.DEFER_UPDATES:
--- a/sdw_notify/NotifyApp.py
+++ b/sdw_notify/NotifyApp.py
@@ -43,7 +43,7 @@ class NotifyDialog(QMessageBox):
 
     def _ui(self):
         self.setWindowTitle(strings.headline_notify_updates)
-        self.setIcon(QMessageBox.StandardButton.Warning)
+        self.setIcon(QMessageBox.Icon.Warning)
         self.setStandardButtons(QMessageBox.StandardButton.No | QMessageBox.StandardButton.Ok)
         self.setDefaultButton(QMessageBox.StandardButton.Ok)
         self.setEscapeButton(QMessageBox.StandardButton.No)


legoktm:
This is a minimal change to get backported into 1.6.0. I will leave writing a regression test for a follow-up.

deeplow (commented):
LGTM. I only got a chance to test on 4.2.

Btw, Claude also flagged another issue that I think only affects 4.3.

```
File: /home/user/securedrop-workstation/sdw_updater/UpdaterAppUiQt6.py
Line 96: Delete font.setWeight(75)

setBold(True) on line 94 is the Qt-idiomatic way to set bold weight and works identically in both PyQt5 and PyQt6. The setWeight(75) call is redundant and only exists
because pyuic6 6.4.2 emitted both when the .ui file specified <bold>true</bold><weight>75</weight>.
```

The issue is that on 4.3 with PyQt6 the weight 75 is no longer on a scale of 1 to 100, but on a 1000 scale. So it'll probably be ridiculously thin. (see [PyQt5 docs](https://www.riverbankcomputing.com/static/Docs/PyQt5/api/qtgui/qfont.html#Weight) on `Weight` table versus [PyQt6's](https://www.riverbankcomputing.com/static/Docs/PyQt6/api/qtgui/qfont.html#Weight)).

But since we'll have to do more 4.3 changes, I think this is ready to do as far as 4.2 is concerned.

deeplow (approved):

deeplow:
> Btw, Claude also flagged another issue that I think only affects 4.3.

I have now filed a separate issue to track this: https://github.com/freedomofpress/securedrop-workstation/issues/1628


> The `sdw-notify` background timer that reminds users if updates haven't run recently should always default to launching the Inbox. This is the only remaining place where UpdaterApp was invoked without an explicit `launch_app`.

Just to spell out explicitly the behavior we chose in this PR, I filed an issue that we probably won't fix https://github.com/freedomofpress/securedrop-workstation/issues/1629.


> This is a minimal change to get backported into 1.6.0. I will leave writing a regression test for a follow-up.

Filed as https://github.com/freedomofpress/securedrop-workstation/issues/1630.

Issue #1618, opened 2026-04-21
bug: SDW Updater mishandles strings


conorsch:
## Description

While reviewing #1616, I encountered problems running the SDW GUI Updater normally, as a baseline. 

<details>
<summary>
error log from sdw-updater
</summary>

```
[conor@dom0 ~]$ sdw-updater
Using sys-firewall as UpdateVM for Dom0
Checking for dom0 updates...
Updating and loading repositories:
 SecureDrop Workstation Qubes dom0 repo 100% |   2.0 KiB/s |   1.5 KiB |  00m01s
 Qubes Host Repository (updates)        100% | 596.0   B/s |   2.7 KiB |  00m05s
 Qubes OS Contrib Repository for dom0 ( 100% |   2.0 KiB/s |   1.5 KiB |  00m01s
 Fedora 37 - x86_64                     100% |   1.6 KiB/s |   3.6 KiB |  00m02s
 Fedora 37 - x86_64 - Updates           100% |   2.5 KiB/s |   3.6 KiB |  00m01s
Repositories loaded.
No new updates available
No dom0 updates available
/usr/lib/python3.11/site-packages/salt/utils/odict.py:28: DeprecationWarning: This module is deprecated. Use the standard library's collections.OrderedDict or salt.utils.datastructures instead.
  warn_until(
/usr/lib/python3.11/site-packages/salt/utils/odict.py:28: DeprecationWarning: This module is deprecated. Use the standard library's collections.OrderedDict or salt.utils.datastructures instead.
  warn_until(
/usr/lib/python3.11/site-packages/salt/utils/odict.py:28: DeprecationWarning: This module is deprecated. Use the standard library's collections.OrderedDict or salt.utils.datastructures instead.
  warn_until(
/usr/lib/python3.11/site-packages/salt/utils/odict.py:28: DeprecationWarning: This module is deprecated. Use the standard library's collections.OrderedDict or salt.utils.datastructures instead.
  warn_until(
/usr/lib/python3.11/site-packages/salt/utils/odict.py:28: DeprecationWarning: This module is deprecated. Use the standard library's collections.OrderedDict or salt.utils.datastructures instead.
  warn_until(
/usr/lib/python3.11/site-packages/salt/utils/odict.py:28: DeprecationWarning: This module is deprecated. Use the standard library's collections.OrderedDict or salt.utils.datastructures instead.
  warn_until(
[...]

PR #1625, opened 2026-04-23
fix: add type annotations for python code

### LINKED ISSUE #1620: add mypy type annotations to all python code
Based on the report in #1618, I suggest we bite the bullet, and add mypy type annotations throughout the Python code in this repo. We'd get a lot more confidence that patches are working as expected. 

One probable exception is that we could continue to exclude the "tests/" directory; I assume that the pytest mocks would be problematic to type accurately. Then again, the tests dir is about a third of the Python code we've written, so warrants scrutiny:

```
❯ tokei -t Python
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 Language              Files        Lines         Code     Comments       Blanks
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 Python                   50         6738         5198          370         1170
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 Total                    50         6738         5198          370         1170
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

❯ tokei -t Python ./tests/
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 Language              Files        Lines         Code     Comments       Blanks
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 Python                   18         2060         1515          141          404
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 Total                    18         2060         1515          141          404
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```

I'll take a pass through the code to add typing, and investigate whether strong typing would have caught the specific bug in #1618.

conorsch:
Adds type annotations throughout the python codebase, and updates the linter config so that annotations are required for all python code going forward, with no exceptions. Fixes #1620.

In the course of adding annotations, a few bugs were revealed, and they're fixed here:

- sdw_updater/UpdaterApp.py:209 - referenced strings.headline_error_reboot,
  which doesn't exist; should be strings.headline_status_error_reboot.
  Would raise AttributeError when executed.
- sdw_notify/NotifyApp.py:46 - QMessageBox.StandardButton.Warning doesn't exist;
  should be QMessageBox.Icon.Warning.
  Would raise AttributeError and prevent display of the notify dialog.
- scripts/try-client-pr.py:158 - subprocess.check_output().splitlines()
  was mixing list-of-bytes and list-of-strings.
- tests/conftest.py:68 - was using non-existent "pytest.warn()" in the
  dom0 fixture; another AttributeError.
- tests/conftest.py:47 - in mock_block_device, more "decode" attempts on str via QubeWrapper.run()

Includes a stub for the "qubesadmin" package, so functions calling its subtypes can be annotated explicitly. Made a light refactor on the Updater calls, trimming an optional timestamp parameter, because it wasn't used.

Notably all checks pass via "make lint". Better, all changes going forward will be required to use type annotations, lest CI fail.

## Testing and review

This branch contains fixes also present in #1622  and therefore #1622 should be reviewed and merged first.

This branch is based on #1624, so that should be reviewed and merged first. Then, after rebasing here, you should _revert_ the [fix commit](https://github.com/freedomofpress/securedrop-workstation/pull/1624/changes/479390763cec8a6b947779cac0368369cdbc2176) locally and ensure that `make lint` catches the string-handling error initially reported in #1618. 

Make sure CI is green across the board. Also, consider whether we really want to require typing in all Python code in the repo.

## Checklist

This change accounts for:


sdw_updater/Updater.py
legoktm: Instead of needing the overload, I would just split the function into two

files/destroy-vm.py
```diff
@@ -34,7 +35,7 @@ def parse_args():
     return args
 
 
-def destroy_vm(vm):
+def destroy_vm(vm: Any) -> None:
```
legoktm: Let's create a stub for qubesadmin like we did for qubesdb in https://github.com/freedomofpress/securedrop-client/blob/main/client/stubs/qubesdb.pyi

legoktm (commented):
This is really great, just some suggestions

files/destroy-vm.py
conorsch: Great pointer! I'd thought the next step would be to use a [mypy Protocol](https://mypy.readthedocs.io/en/stable/protocols.html), which felt heavy, but I'm pleased to learn that stubs are a thing.

conorsch: Added a stub for `qubesadmin`, which strengthens the associated typing.

sdw_updater/Updater.py
conorsch: I ended up removing the `with_timestamps` param altogether, which required a light refactor in associated callsites. This is probably the most substantive part of the diff, so please scrutinize the change to ` Updater.read_dom0_update_flag_from_disk()`.

conorsch:
The stub logic needs more work: I'm seeing import errors in dom0. Moving back to in-progress for now.

conorsch:
Ready for review. @legoktm overall the type annotations are unobjectionable; however, there's a second commit that adds TypedDict for the dom0 config JSON parsing. I think that addition is worthwhile to include here; but it's worth noting that I didn't go all the way and enforce "parse, don't validate" which is out of scope for this change set. Flagging for your special attention in case you have strong feelings either way. I made sure to keep that change in a discrete commit, so it's easy to drop.

N.B. the [OpenQA 4.2 suite is failing](https://openqa.qubes-os.org/tests/175632/logfile?filename=serial_terminal.txt), but the only error listed in there is fedora-42-eol, which will shortly be resolved via #1638.

files/validate_config.py
```diff
@@ -26,58 +29,61 @@ class ValidationError(Exception):
...
-        self.config = self.read_config_file()
+        # Validators below operate on the raw, untrusted dict view.
+        self._raw_config: dict[str, Any] = self.read_config_file()
         self.confirm_onion_config_valid()
         self.confirm_submission_privkey_file()
         self.confirm_submission_privkey_fingerprint()
         self.confirm_environment_valid()
         self.validate_existing_size()
+        # All validators have passed; expose the typed view to consumers.
+        self.config: Dom0Config = cast(Dom0Config, self._raw_config)
```
legoktm: Does anything actually use this? I couldn't immediately find anything. If not, we can remove it.

I was going to leave a longer message about how this is definitely an improvement but now there's a question of when to use self.raw_config vs self.config leading to potential mistakes. But if nothing uses self.config anymore it doesn't matter as much?

Personally I would pass in raw_config as an argument to each of the validating functions instead of as a member variable where it theoretically could be reached even in failure, but realistically I don't think it matters that much.

sdw_updater/Updater.py
```diff
@@ -509,14 +520,14 @@ def should_launch_updater(interval):
     return True
 
 
-def _valid_status(status):
+def _valid_status(status: dict[str, Any] | None) -> TypeGuard[dict[str, Any]]:
```
legoktm, line 524: nice!

stubs/qubesadmin/__init__.pyi
```diff
@@ -0,0 +1,14 @@
+# Type stubs for the `qubesadmin` package, covering the surface area used by
```
legoktm, line 1: Did you write these by hand or automatically? A very very brief README might be nice for when we eventually need to update it.

legoktm (changes requested):
This is really great, thank you. My inline comments are pretty minor.

The TypedDict is nice, I do think that now that we have it, the next logical follow-up is to move to a dataclass, so we get runtime typechecking too.

files/validate_config.py
conorsch: Added a new commit that implements "parse don't validate" as requested. It adds a lot to the diff, so I'd appreciate a careful eye. Also tacked on a bunch of tests that exercise the validation logic against intentionally malformed JSON fixtures for bad dom0 configs. Perhaps overkill, but I appreciate the regression guards.

stubs/qubesadmin/__init__.pyi
conorsch: Automatically, although it took several passes and a lot of nudging to map out the necessary boundaries. Added a README that points to the upstream `qubesadmin` src, so that maintainers can easily look up the relevant type signatures if and when they change.

files/validate_config.py
```diff
@@ -10,74 +14,42 @@
...
 
 # CONFIG_FILEPATH = "/srv/salt/securedrop_salt/config.json"
 CONFIG_FILEPATH = "config.json"
 SECRET_KEY_FILEPATH = "sd-journalist.sec"
 
-
-class ValidationError(Exception):
-    pass
+# Re-export for callers that historically imported ValidationError from this module.
+__all__ = ["SDWConfigValidator", "ValidationError"]
```
legoktm: In a follow-up, we should just update the callers

tests/test_dom0_validate.py
```diff
@@ -49,11 +49,22 @@ def test_config_malformed_onion_json(test_resources_dir, tmpdir):
...
 
     with pytest.raises(ValidationError) as exc_info:
         SDWConfigValidator(tmpdir)
 
     assert "Invalid PGP key fingerprint specified" in exc_info.exconly()
+
+
+def test_config_mismatched_fpr(test_resources_dir: Path, tmpdir: Path) -> None:
+    """A well-formed but wrong fingerprint must be rejected against the on-disk key."""
+    shutil.copy(f"{test_resources_dir}/testconfig.json.mismatched_fpr", f"{tmpdir}/config.json")
```
legoktm: Do we need to undo this?

legoktm (commented):
This is really great, just one question. Otherwise consider this approved!

files/validate_config.py
conorsch: The commit to update the call-sites is surprisingly small, so I've appended here.

tests/test_dom0_validate.py
conorsch: No, we don't need to handle tmpdir cleanup, because the `tmpdir` in the args is truly a temporary directory and therefore will get cleaned up automatically by the test harness. However, Copilot rightly [points out below](https://github.com/freedomofpress/securedrop-workstation/pull/1625#discussion_r3197136422) that the type for "tmpdir" is not actually `Path` and recommends the [tmp_path fixture](https://docs.pytest.org/en/6.2.x/tmpdir.html) instead. This is a good catch.

conorsch:
Rebased to sign everything, to satisfy the merge requirement. We should expect CI to pass.

tests/test_dom0_validate.py
legoktm: Right :)

legoktm (dismissed):
Thanks for sticking with it, this is going to be really great :)

Approving for whenever CI passes

conorsch:
Rebased one more time, to include merge of #1638, so we can expect green CI across the board.

conorsch:
Needs a rebase due to tiny change in #1656; plus, I'm on quest to get the 4.2 suite passing here, so will keep nudging until I see green, then merge.

conorsch:
ci-let-there-be-green

It's beautiful. :cry:

conorsch:
> Needs a rebase due to tiny change in https://github.com/freedomofpress/securedrop-workstation/pull/1656

Looks like my rebase cleared the approval, and thus the PR is stuck in the queue. Whenever you've got a moment, @legoktm, another stamp should get this landed.

legoktm (approved):

PR #1637, opened 2026-04-29
Stop writing update flags to sd-app

### LINKED ISSUE #1632: Stop writing update status to sd-app's ~/.securedrop_client
Currently the updater writes a status flag to sd-app's ~/.securedrop_client. This folder is going away so it would need to be updated, but AFAICT nothing in the legacy Client actually reads from this flag anyways that is worth porting to Inbox, so we should just get rid of it.

legoktm:
These are entirely unused, and ~/.securedrop_client will go away soon. Plus if we did still want this feature we would implement it using a custom RPC rather than having the updater directly write into a VM with qvm-run.

Fixes #1632.

## Test plan

* [ ] visual review is sufficient?
## Checklist

This change accounts for:

### DIFF
--- a/launcher/tests/test_updater.py
+++ b/launcher/tests/test_updater.py
@@ -143,21 +143,13 @@ def test_apply_templates(
 
 
 @pytest.mark.parametrize("status", UpdateStatus)
-@mock.patch("subprocess.check_call")
 @mock.patch("sdw_updater.Updater.sdlog.error")
 @mock.patch("sdw_updater.Updater.sdlog.info")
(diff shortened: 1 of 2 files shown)


legoktm:
No urgency on this, it's wrapped in a try/except that fails without blocking anything.

conorsch:
Looks good. Tested locally, and OpenQA passes, too. But as the OP says: a simple grep shows that this was dead code. 

Merge might introduce small conflicts with #1625, but those'll be easy to address in a rebase.

conorsch (approved):

PR #1640, opened 2026-04-30
Add semgrep for SaltStack regression tests

### LINKED ISSUE #1499: SaltStack `cmd.run` doesn't get `args` passed along (Qubes 4.3)

deeplow:
Adds semgrep to the repo and a base test to check for #1499 regressions. This was motivated by a lack of tooling in the repo to lint saltstack and a regression introduced in https://github.com/freedomofpress/securedrop-workstation/pull/1582#issuecomment-4351485347.

We are already using semgrep in the client repo, so I figured it would make sense to add it here without much need for discussion.

## Test plan

- [ ] Semgrep addition:
  - [ ] poetry-related files make sense (only semgrep added to the dev group — as well as its dependencies
  - [ ] `make semgrep` checks for issues and exists with 1 if it issue detected (tip: temporarily remove the commit that fixes #1499)
  - [ ] `make lint` calls semgrep target
- [ ] CI passes in last commit, but fails on the commit without the fix: 207cfd6.
- [ ] Regression test for #1499
  - [ ] fault injection on the test and the code to see if it detects the issue properly

## Checklist

This change accounts for:


.semgrep/saltstack-rules.yml
```diff
@@ -0,0 +1,16 @@
+rules:
+  - id: saltstack-cmd-script-args-qubes-4-3
+    patterns:
+      - pattern: |
+          cmd.script:
+            ...
+            - args: ...
+    message: |
+      SaltStack cmd.script silently ignores args
```
legoktm: Let's do something like:

```suggestion
      SaltStack cmd.script silently ignores args, use cmd.run instead
```

.semgrep/saltstack-rules.yml
```diff
@@ -0,0 +1,16 @@
...
+      (Regression test for issue 1499)
```
legoktm: I don't think we need this in the message, it could just be a YAML comment.

legoktm (changes requested):
This is really great, just one small comment about the wording.

legoktm (approved):
Great :)

PR #1654, opened 2026-05-11
Adjust headline font weight for Qt6

### LINKED ISSUE #1628: SDWNotify: Thin text was supposed to be bold
Claude also flagged another issue with the PyQt5 to PyQt6 porting done earlier that I think only affects 4.3. (Note: I have not visually confirmed this yet, but the reasoning makes sense).

> ```
> File: /home/user/securedrop-workstation/sdw_updater/UpdaterAppUiQt6.py
> Line 96: Delete font.setWeight(75)
>
> setBold(True) on line 94 is the Qt-idiomatic way to set bold weight and works identically in both PyQt5 and PyQt6. The setWeight(75) call is redundant and only exists
> because pyuic6 6.4.2 emitted both when the .ui file specified <bold>true</bold><weight>75</weight>.
> ```
> 
> The issue is that on 4.3 with PyQt6 the weight 75 is no longer on a scale of 1 to 100, but on a 1000 scale. So it'll probably be ridiculously thin. (see [PyQt5 docs](https://www.riverbankcomputing.com/static/Docs/PyQt5/api/qtgui/qfont.html#Weight) on `Weight` table versus [PyQt6's](https://www.riverbankcomputing.com/static/Docs/PyQt6/api/qtgui/qfont.html#Weight)).
> 
> 
> — _Originally posted by @deeplow in https://github.com/freedomofpress/securedrop-workstation/pull/1622#pullrequestreview-4162005068_

legoktm:
In Qt6, font weights are now out of 1000, so 75 is actually quite thin instead of being bold.

Based on testing, DemiBold (weight=600) is closest to what we had under PyQt5/Qubes 4.2 so that's what I've gone with here.

The .ui file is updated to use 600 to be forward looking, but the Python code will still handle PyQt5 correctly.

Fixes #1628.

## Test plan

* [ ] Review screenshots in https://github.com/freedomofpress/securedrop-workstation/issues/1628#issuecomment-4424839012 and agree with me that demibold is the best looking one
* [ ] try it in Qubes 4.3, verify it looks good
## Checklist

This change accounts for:


conorsch:
This was bothering me, even though I couldn't put my finger on it. Tested locally; results:

### Before

sdw-updater-before

### After

sdw-updater-after

The revised version is definitely more pleasing to my eye.

conorsch (approved):

PR #1656, opened 2026-05-11
Have mypy ignore rpm-build/

legoktm:
After a `make rpm-build` there might be duplicate Python files there that conflict with the main copies, causing duplicate module failures.

## Test plan

* [ ] `make rpm-build && make mypy` doesn't fail with duplicate module issues
## Checklist

This change accounts for:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,6 +122,7 @@ files = [
 ]
 exclude = [
     "launcher/", # Moving to sd-updater
+    "rpm-build/",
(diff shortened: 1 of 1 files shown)


conorsch:
Ah-ha! I've seen this, but hadn't tracked down the error, due to switching branches so much. I'd thought it was related to #1625. Thanks for fixing!

conorsch (approved):

Issue #1641, opened 2026-04-30
Newly installed Fedora template missed updates (but got them after next updater run)


deeplow:
## Description

The salt state `update-fedora-template-if-new` in [`sd-sys-vms.sls`](https://github.com/freedomofpress/securedrop-workstation/blob/1.6.0/securedrop_salt/sd-sys-vms.sls#L38) was silently being ignored, even though it looked like it was running just fine:

```
----------
     ID: update-fedora-template-if-new
Function: cmd.wait
   Name: qubes-vm-update --quiet --force-update --targets fedora-41-xfce
 Result: True
Comment: 
Started: 16:52:35.772452
Duration: 0.838 ms
Changes:   
 ----------
```

> NOTE: on a real command execution, you'll see a comment and an action process ID:
> ```
> Comment: Command "qubes-vm-update --quiet --force-update --targets fedora-43-xfce" run
> ```

This bug was introduced in https://github.com/freedomofpress/securedrop-workstation/pull/1280 (released in 1.4.0). Longer explanation in the `comment` section.

#### Impact

- **Clean installs** ever since 1.4.0 were affected in the first installation. There, it was silently configuring qubes with an outdated management dvm. After rebooting the system and running the updater, it is fully patched.

## Steps to Reproduce

See testing instructions in https://github.com/freedomofpress/securedrop-workstation/pull/1638. The fact that [an already-fixed bug](https://github.com/freedomofpress/securedrop-workstation/pull/1638#issuecomment-4352988535) was hit, made it evident that `qubes-vm-update` had not run. You can also look at the vm startup notifications and see that when you run `sudo qubesctl --show-output state.apply securedrop_salt.sd-sys-vms` on a system without the latest fedora, it doesn't start fedora-43-dvm during the update step.

## Expected Behavior
Updater runs

## Actual Behavior
Updater does not run (even though it looks like it does).

## Comments

\<rant\>
 Turns out previously we had `cmd.wait`, which was reacting (via `watches:`) to the template installation (which was a `cmd.run`). Then, when we decided to switch to the native `qvm.template_installed` we swapped the `watches` with onchanges because the `qvm.template_installed` implementation missed a [`mod_watch`](https://github.com/freedomofpress/securedrop-workstation/pull/1280/changes/73db2d101eb6837795560ba84f9072d13ffd33d5), which made it silently fail. This got fixed back then.

[...]

Issue #758, opened 2022-03-31
Updater should refresh list of target VMs after dom0 update


eloquence:
Currently, the list of VMs the updater will target is hardcoded here:
https://github.com/freedomofpress/securedrop-workstation/blob/356b289390005d59eaa0efd2cd7647c4f75e3b9d/launcher/sdw_updater_gui/Updater.py#L40-L49

As a result, if _during an update_, the list changes, the updater will happily continue on its run using the old list. For example, I just updated on old system that hadn't received a Whonix template update yet. The updater downloaded Whonix 16, but continued to update Whonix 15 instead. Of course, this will be cleared up on the next run, but it would seem to be preferable to refresh the list after the `dom0` update (e.g., via `qvm-prefs` approach as suggested in the code comment).

zenmonkeykstop:
Flagging for @deeplow's attention - if possible we should update the list of non-dom0 target VMs after any potential dom0 chnages, maybe using VM tags instead of an explicit list.

PR #1666, opened 2026-05-14
Semgrep: detect salt pitfalls (watch, cmd.wait)

deeplow:
Detect tricky salt commands that even the documentation [warns about](https://docs.saltproject.io/en/latest/ref/states/requisites.html#watch). Helps detect regressions similar to #1641.

## Test plan

- [ ] `make semgrep` exits `0`
- [ ] Apply the following patch
    ```patch
    diff --git a/securedrop_salt/sd-sys-vms.sls b/securedrop_salt/sd-sys-vms.sls
    index a1298f8..d058f05 100644
    --- a/securedrop_salt/sd-sys-vms.sls
    +++ b/securedrop_salt/sd-sys-vms.sls
    @@ -40,7 +40,7 @@ set-fedora-template-as-default-mgmt-dvm:
     #     running updater has a stale list
     #     (see https://github.com/freedomofpress/securedrop-workstation/issues/758)
     update-fedora-template-if-new:
    -  cmd.run:
    +  cmd.wait:
         - name: qubes-vm-update --quiet --force-update --targets {{ sd_fedora_base_template }}
         - runas: {{ gui_user }}
         - require:
    @@ -59,6 +59,8 @@ set-fedora-default-template-version:
         - require:
           - qvm: dom0-install-fedora-template
           - sls: qvm.default-dispvm
    +    - watch:
    +      - cmd: update-fedora-template-if-new
     
     # On 4.1, several sys qubes are disposable by default - since we also want to
     # upgrade the templates for those, we need to ensure that the respective dvms
    ```
- [ ] `make semgrep` exits `1` and accuses the errors.
 
## Checklist

This change accounts for:

### DIFF
--- a/.semgrep/saltstack-rules.yml
+++ b/.semgrep/saltstack-rules.yml
@@ -12,3 +12,29 @@ rules:
     paths:
       include:
         - "*.sls"
+
+  - id: saltstack-cmd-wait-discouraged # "Regression test" for issue #1641
+    patterns:
+      - pattern: |
+          cmd.wait:
+            ...
+    message: SaltStack's cmd.wait is particularly tricky and the docs discourage it
+    languages:
+      - generic
+    severity: ERROR
+    paths:
+      include:
+        - "*.sls"
+  - id: saltstack-watch-requires-mod-watch # "Regression test" for issue #1641
+    patterns:
+      - pattern: |
+          ...:
+            ...
+            - watch: ...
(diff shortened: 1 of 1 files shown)


legoktm (changes requested):
The thing copilot flagged about duplicate IDs needs to be fixed. Otherwise it's good!

deeplow:
Whoops. Done.

legoktm (approved):
Thanks!

PR #1683, opened 2026-05-18
Stop installing securedrop-client and hide it

legoktm:
Remove the "SecureDrop Client (legacy)" Qubes menu entry as a minimal way to remove the Client; the package itself will be removed as part of the next Inbox release.
    
And stop installing the securedrop-client package on fresh installs.

Refs https://github.com/freedomofpress/securedrop-client/issues/3411.

## Test plan

* [ ] Do a fresh install (`--uninstall` is fine), verify the `securedrop-client` package is not installed in the small template
* [ ] smoke test that Inbox runs fine, i.e. wasn't relying on anything the `securedrop-client` package pulled in to the sd-app VM.
* [ ] Open the Qubes menu, "SecureDrop Client (legacy)" is no longer listed
## Checklist

This change accounts for:

### DIFF
--- a/tests/test_vm_sd_app.py
+++ b/tests/test_vm_sd_app.py
@@ -26,7 +26,6 @@ def qube() -> QubeWrapper:
             "SD_SUBMISSION_KEY_FPR",
             "SD_MIME_HANDLING",
         },
-        enforced_apparmor_profiles=set(),
         mime_types_handling=True,
     )
 
--- a/tests/test_vm_sd_app.py
+++ b/tests/test_vm_sd_app.py
@@ -26,9 +26,7 @@ def qube() -> QubeWrapper:
             "SD_SUBMISSION_KEY_FPR",
             "SD_MIME_HANDLING",
         },
-        enforced_apparmor_profiles={
-            "/usr/bin/securedrop-client",
-        },
+        enforced_apparmor_profiles=set(),
         mime_types_handling=True,
     )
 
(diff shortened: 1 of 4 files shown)


legoktm:
This is nearly ready for review; I need to test to see if we need salt to explicitly uninstall the client package or if we can do that entirely in the inbox side of things.

tests/test_vm_sd_app.py
```diff
@@ -26,9 +26,7 @@ def qube() -> QubeWrapper:
             "SD_SUBMISSION_KEY_FPR",
             "SD_MIME_HANDLING",
         },
-        enforced_apparmor_profiles={
-            "/usr/bin/securedrop-client",
-        },
+        enforced_apparmor_profiles=set(),
```
deeplow: Nit: Do we want this there to make it more explicit that this does not have an apparmor profile?

If not, I'd suggest we just remove it, since it's the default:

```suggestion
```

deeplow:
> ## Test plan

I have tweaked this slightly to also include an "update scenario"
- [x] Confirm presence of legacy menu item
- [x] update via RPM installation (no need for `sdw-admin --apply`)
- [x] Confirm absence of legacy menu item

The following is the original one, which I'll use for the clean install scenario.
>  * [x]  Do a fresh install (`--uninstall` is fine), verify the `securedrop-client` package is not installed in the small template
> * [x]  smoke test that Inbox runs fine, i.e. wasn't relying on anything the `securedrop-client` package pulled in to the sd-app VM.
> * [x]  Open the Qubes menu, "SecureDrop Client (legacy)" is no longer listed

On top of this I also logged out, to make sure the menu has had the opportunity to fully refresh.

deeplow (changes requested):
Minor nitpick, but otherwise LGTM code-wise (still testing).

tests/test_vm_sd_app.py
legoktm: No, I just was lazy and assumed it was a required parameter, we can remove it.

deeplow (approved):

PR #1728, opened 2026-06-17
Fix `make staging` in Qubes 4.3 

micahflee:
I was getting a crash when running `make staging`, complaining that `/var/lib/qubes/updates/repodata/repomd.xml` did not exist.

With the help of Claude Sonnet 4.6, it turns out this is because `qubes-dom0-update --clean` deletes `/var/lib/qubes/updates/`, and it only recreates it when a package needs to be installed.

This fixes it by skipping running `qubes-dom0-update` when the package (like `securedrop-workstation-keyring-staging`) is already installed, which prevents it from deleting `repomd.xml`, allowing the installation to proceed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,11 +63,11 @@ install-rpm: assert-dom0 ## Install locally-built rpm (dev) or download publishe
 	@rpm -q securedrop-workstation-keyring || sudo qubes-dom0-update -y --clean securedrop-workstation-keyring
 ifeq ($(RPM_INSTALL_STRATEGY),dev)
 	@echo "Install dependencies and locally-built rpm"
-	@sudo qubes-dom0-update --clean -y grub2-xen-pvh
+	@rpm -q grub2-xen-pvh || sudo qubes-dom0-update --clean -y grub2-xen-pvh
 	@./scripts/prep-dev
 else
 	@echo "Install published rpm"
-	@sudo qubes-dom0-update -y securedrop-workstation-dom0-config
+	@rpm -q securedrop-workstation-dom0-config || sudo qubes-dom0-update -y securedrop-workstation-dom0-config
 endif
 	@echo "Provide instance-specific configuration and run sdw-admin --apply."
 
--- a/scripts/bootstrap-keyring.py
+++ b/scripts/bootstrap-keyring.py
@@ -42,7 +42,7 @@ def rpm_import(key_file: Path) -> None:
 def is_key_imported(rpm_id: str) -> bool:
     """Check rpmdb for key with givem rpm_id."""
     try:
-        subprocess.check_call(["rpm", "-q", rpm_id])
+        subprocess.check_call(["rpm", "-q", "--quiet", rpm_id])
         return True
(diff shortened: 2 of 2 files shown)


Makefile
deeplow: For completion sake we might as well do this here as well:
```suggestion
	@rpm -q grub2-xen-pvh || sudo qubes-dom0-update --clean -y grub2-xen-pvh
```

scripts/bootstrap-keyring.py
```diff
@@ -50,13 +50,26 @@ def is_key_imported(rpm_id: str) -> bool:
...
+    try:
+        subprocess.check_call(
+            ["rpm", "-q", package_name],
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+        )
+        print(f"Package {package_name} is already installed.")
+        return
+    except subprocess.CalledProcessError:
+        pass
```
deeplow: We already have a function for this. So maybe we can re-use it. I was tempted to suggest a rename like `is_package_installed`, but `is_key_imported` kind of also fits the meaning, even though we're querying for packages and not RPM key ids. But it works  ̄\_ (ツ)_/ ̄

```suggestion
    if is_key_imported(package_name):
        print(f"Package {package_name} is already installed.")
        return
```

I've also removed the long comment, because I think it's now self-explanatory.

scripts/bootstrap-keyring.py
```diff
@@ -50,13 +50,26 @@ def is_key_imported(rpm_id: str) -> bool:
...
+    # Skip if already installed. On Qubes 4.3, qubes-dom0-update --clean wipes
+    # the local package cache; if no packages are then downloaded (already installed),
+    # the dom0-local DNF phase fails because it can't read the now-empty
+    # qubes-dom0-cached repo. This matches the pattern used in the Makefile.
+    try:
+        subprocess.check_call(
+            ["rpm", "-q", package_name],
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+        )
```
deeplow: The `--quiet` avoids the need for stdout and stderr as DEVNULL:

```suggestion
        subprocess.check_call( "-q", "--quiet", package_name]```
```

But if we go with my suggestion, this should just be added in `is_key_imported()` instead. This would also avoid the annoying key ID from being printed when it checks for those:

```
[user@dom0 securedrop-workstation]$ ./scripts/bootstrap-keyring.py --env staging
gpg-pubkey-3fab65ab-660f2beb
gpg-pubkey-3fab65ab-660f2beb
[...]
```

deeplow (commented):
Thanks for catching this one. I suggested some simplifications, just to keep this script small.

deeplow (approved):
Thanks for addressing these.

If possible would you be able to either squash the commits or amend the first one's message? It had a very long commit message that makes it hard to read, especially on GitHub — what I generally do is a short 50 chars subject, then a space and then the body.

deeplow (approved):

Issue #1722, opened 2026-06-09
Changing template names: `small` and `large` are unclear


deeplow:
## Description
The current templates are named `sd-small-bookworm-template` and (`sd-large-...`). Large and small do not sufficiently describe their differences and expectations.

> **NOTE:** The ongoing work to update to Debian 13 in https://github.com/freedomofpress/securedrop-workstation/pull/1719 is the perfect timing for this rename, without any additional work.

### Background
There used to be multiple templates, but then they [got consolidated into the current two](https://github.com/freedomofpress/securedrop-workstation/pull/619/changes). Stated reasoning for separate templates is related to to reducing supply chain risks:

> I think this makes sense as a major simplification, faster provisioning time, and for the sped up update time. One thought is instead of having a single `securedrop-workstation-buster` template or template per-VM instead we could have two kinds of templates that are based on `securedrop-workstation-buster`:
> 
> * **Lower security template** - We install printer drivers (eventually installing potentially shady drivers will be required for export code as we expand printer support), and potentially an ever expanding suite of programs for opening a broader set of different files. We'd use this for `sd-devices` and `sd-viewer`.
> 
> * **Higher security** - we use this for `sd-app`, `sd-log`, `sd-proxy` (which e.g. has the journalist interface URL currently in `/etc` in the template (we could also change this but regardless I think we'd want to have two types of templates given the considerations in the above bullet)), and `sd-gpg`. We are more judicious about what we install here since it will be shared between multiple VMs. 
>
> — _Originally posted by @redshiftzero in [#471](https://github.com/freedomofpress/securedrop-workstation/issues/471#issuecomment-595465108)_

> #### Number of Templates:
> 
> - We have the minimal templates listed as a mitigation (MITIG24 for: Vulnerability introduced via Template Dependency and Malicious (backdoored) VM Template: A reduced package list will reduce the likelihood of compromise, but also potentially the impact (if it affects a sensitive VM such as `sd-gpg` or `sd-client`, the impact is higher than for the `sd-viewer`). In this case, we rely on lack of network and Xen to prevent exfiltration.
> 
> —  _Originally posted by @emkll in [#471](https://github.com/freedomofpress/securedrop-workstation/issues/471#issuecomment-654418868)_

### Proposal: indicate trust level

My suggestion would be to reflect the trust expectations in the template's name:

- `small` → `trusted`
- `large` → `untrusted`

*Any other ideas?*

legoktm:
[...]

Issue #1712, opened 2026-06-01
Have salt provision a trixie template and migrate VMs


legoktm:
Update our salt code to provision a new trixie template and migrate all of our VMs to use it.

deeplow:
Some implementation options I am considering proposing in a PR:
- I briefly considered `debian-13` as base, since `sd-admin`'s implementation uses [this one as well](https://github.com/freedomofpress/securedrop-workstation/pull/1572/changes#diff-6b8338d5b45342a15d61bba3d1457f45279f405619ff2adb6104fd15861e0b05R5-R8) (although I known few systems will be both admin and workstations). But this seems to have way too much software. So I'll probably keep the minimal, even though the [docs](https://doc.qubes-os.org/en/latest/user/templates/minimal-templates.html) specifically point out that:
       ```
       You could, of course, use `qubes-vm-recommended` to automatically install many of these, but in that case you are well on the way to a standard Debian template.
       ```

- **Dropping "base" sd template** — Looking at the code, I don't see a reason to keep the base template. At the end of the day it only helps not downloading core shared packages the first time, at the expense of never being useful ever again and yet:
  - Still needing a dedicated configuration via salt on every `sdw-admin --apply`
  - Extra storage for storing this template

- **Centralizing the template names**— it's really tempting to do #1158 because we have so many references to "bookworm" and "debian-12" and sooner or later this tech debt may cause issues. I can see if we can do a small to review, but sane improvement here.

legoktm:
I guess the base template helps avoid needing to download the kernel twice, but it's only a one-time optimization because all future updates have to download it twice for small and large, so I'm +1 on getting rid of it. That said, if it ends up being more complicated than needed, we can punt on this because we can always remove the base template later, right?

For centralizing template names it would be nice but if it's not straightforward I don't think you need to go out of your way.

deeplow:
> I guess the base template helps avoid needing to download the kernel twice

I also thought about that. Then I took a look the actual thing, and it's just 24.0MB.

> but it's only a one-time optimization because all future updates have to download it twice for small and large, so I'm +1 on getting rid of it.

👍 

> That said, if it ends up being more complicated than needed, we can punt on this because we can always remove the base template later, right?

I guess we can, but it'll be annoying because we'll need to keep the removal. But we totally do it.


> For centralizing template names it would be nice but if it's not straightforward I don't think you need to go out of your way.

OK. I'll see what I come up with.

PR #1737, opened 2026-06-25
Rename templates: `sd-small`/`sd-large` → `sd-inbox`/`sd-viewer`

deeplow:
Completes the template rename discussed in #1722.

**NOTE:** I decided to split this one off from #1712 because it's a large one, but pretty repetitive. I'll have another one to tackle in #1712. Upgrade is still expected to fail in this PR. (CC @legoktm)

## Test plan
- [ ] look in the code if there are no more missing `small` / `large` references

Post-merge:
  - [ ] docs changes to fix old template name references

## Checklist

This change accounts for:

### DIFF
--- a/securedrop_salt/securedrop-handle-upgrade
+++ b/securedrop_salt/securedrop-handle-upgrade
@@ -83,7 +83,7 @@ elif [[ $TASK == "remove" ]]; then
     # before deleting it.
     # TODO: clean this up, we don't have separate templates anymore and nobody
     # will be upgrading from the original setup
-    for template in sd-small-bookworm-template sd-large-bookworm-template
+    for template in sd-inbox-bookworm-template sd-viewer-bookworm-template
     do
         if qvm-check "${template}" --quiet; then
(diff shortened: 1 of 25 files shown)


securedrop_salt/securedrop-handle-upgrade
legoktm, line 69: Missed this and the one for sd-log

legoktm (changes requested):
Everything seems to be working, just found a spot in the handle-upgrade script that still refers to the old names, unless that was intentional.

securedrop_salt/securedrop-handle-upgrade
deeplow: That was intentional. I meant to have it as a separate PR, because the upgrade script also had other edge-cases that were causing issues (although it's very tedious to test it).

legoktm: Got it

legoktm (approved):
LGTM then

PR #1739, opened 2026-06-25
Update test_mime_types for switch to gnome-text-editor

legoktm:
Refs <https://github.com/freedomofpress/securedrop-client/pull/3524>.

## Test plan

* [ ] test passes in OpenQA
## Checklist

This change accounts for:


deeplow (approved):
Thanks. We should actually move these tests to the client and then have the OpenQA somehow call them. Otherwise we'll need to go back there to notice the issue.

Issue #1680, opened 2026-05-18
validate_config.py can no longer be run standalone


nathandyer:
During a pristine install on Qubes 2.4.2, I cannot complete a successful `make staging` run. What I did was:

1. Install Qubes according to https://workstation.securedrop.org/en/stable/admin/install/prepare.html#install-qubes-os-estimated-wait-time-30-45-minutes (no Whonix)
2. Run all updates
3. Clone the securedrop-workstation repo in the work VM, then send to dom0
4. Add config.json and sd-journalist.sec to the securedrop-workstation directory in dom0
5. `cd securedrop-workstation && make staging`

It fails at:

```
Traceback (most recent call last):
  File "/home/user/securedrop-workstation/./files/validate_config.py", line 21, in <module>
    from sdw_util.config_types import Dom0Config, ValidationError
ModuleNotFoundError: No module named 'sdw_util.config_types'
```

Full output:

<details>

```
[user@dom0 securedrop-workstation]$ make staging
gpg-pubkey-3fab65ab-660f2beb
gpg-pubkey-3fab65ab-660f2beb
Using sys-firewall as UpdateVM for Dom0
Downloading packages. This may take a while...
Cache directory "/var/lib/qubes/dom0-updates/var/cache/dnf" does not exist. Nothing to clean.
Updating and loading repositories:
 Fedora 37 - x86_64 ???% |   0.0   B???% |   0.0   B
 Qubes Host Repository (updates) ???% |   0.0   BA
 Fedora 37 - x86_64 100% |   4.0 KiB
 Qubes Host Repository (updates) ???% |   0.0   BA
 Fedora 37 - x86_64 100% |   4.0 KiB
 Qubes Host Repository (updates) 100% |   3.4 KiBA
 Fedora 37 - x86_64 100% |   4.0 KiB
 SecureDrop Workstation Keyring (staging) ???% |   0.0   B
 Qubes Host Repository (updates) 100% |   3.4 KiB
 Fedora 37 - x86_64 100% |  10.2 KiB
 SecureDrop Workstation Keyring (staging) ???% |   0.0   B
 Qubes Host Repository (updates) 100% |   5.3 KiB
 Fedora 37 - x86_64 100% |  10.2 KiB
 SecureDrop Workstation Keyring (staging) 100% |   1.5 KiB
 Qubes Host Repository (updates) 100% |   5.3 KiB
 Fedora 37 - x86_64 100% |  10.2 KiB
[...]

PR #1744, opened 2026-06-29
Replace handle-upgrade script with generic logic

### LINKED ISSUE #1225: Discussion: can we deprecate or simplify `securedrop-handle-upgrade`?

### LINKED ISSUE #1712: Have salt provision a trixie template and migrate VMs

deeplow:
The bash script had gone stale as many things changed since it was last used for a major debian upgrade. There were several problems motivating this change:
  - awkard salt->bash script call
  - hard-coded vm names that needed some updating
  - remove qubes where in reality it only needed to shut them down

The man reason why the script existed was to shut down the old templates prior to the switch. If not done, this could make the switch fail even if even a dispvm is indirectly based on it.

The solution was to write it in sdw-admin as a sort of salt wrapper. Moving this to salt was a work in progress, but was turning out more complex than necessary, especially due to the state ordering needs.

Introduces two new mechanics:
1. **Disable preloaded disposables** (during entire provisioning and configuration)
   - saves time as preloaded disposables were re starting up as the templates got changed
   - frees up RAM, which makes out of memory failures more rare during apply (and we know the system won't be used during this time)
   - critically, it prevents some `sdw-admin --apply` failures where template switching couldn't be completed because some qube based on the old template was still running. I believe this was the case with preloaded disposables. The second such scenario was where those preloaded disposables would call the RPC `securedrop.Log`, thus making `sd-log` start up continuously and resulting in the same kind of failure.

2. **Shuts down and prevents startup during provisioning** — in provisioning (i.e. creating and changing VMs) we can get away with this as no qubes are expected to start as a result. Here's how that works: 
    1. detects need to change template via template feature 'os-version' shipped with every default template since Qubes 4.2.
    2. shuts down all sdw qubes
      **NOTE:** selecting just the template-derived via qube.derived_vms would have been appropriate, but since when debian changes everything changes, a shortcut was made.
    3. Prohibits startup of every qube during provisioning to account for startups via qrexec calls such as securedrop.Log or securedrop.Proxy, which would create race conditions. This makes use of a Qubes 4.3+ feature.
    4. Performs provisioning where template switching can occur safely
    5. Re-enables starting various qubes.

Fixes #1712 
Fixes #1225

## Test plan

- [x] install prod 1.7.1
- [x] make dev and ensure:
  - [x] no warnings show up about preloaded disposables attempting to start but failing (feel free to see `journalctl -u qubesd` instead)
  - [x] No major errors / failures (please restart entire test process if something occurs)
  - [x] At the end no `sd-*-bookworm-template` exist
  - [x] Proper templates are set
  - [x] sd-viewer is still set as default disposable
  - [x] preloaded disposables enabled at the end
  - [x] smoke test inbox
- [x] repeat installation but inject failure — e.g. by shutting down a qube mid-provisioning
  - [x] ensure a subsequent `sdw-admin --apply` works
  - [x] preloaded disposables are in the background 
- [x] `sdw-admin --uninstall` is able to remove all sdw qubes
- [ ] Confirm upgrade OpenQA testing passed: [job 186347](https://openqa.qubes-os.org/tests/186347/) (logs failed to upload, but the [video is available](https://openqa.qubes-os.org/tests/186347/video?filename=video.webm&t=320.54,320.58)). This was run on commit `801c775`. Please also check that the failed tests in the [respective dom0 tests](https://openqa.qubes-os.org/tests/186348#dependencies) are the expected ones and nothing else.
  - This ran with https://github.com/freedomofpress/openqa-tests-qubesos/pull/23

## Checklist

This change accounts for:


deeplow:
I'm running an upgrade scenario at https://openqa.qubes-os.org/tests/186217#live. [Basically installs prod first](https://github.com/freedomofpress/openqa-tests-qubesos/compare/main...2026-06-upgrade-testing) (with `SECUREDROP_UPGRADE=1`) and then `dev` from this branch (or whatever is set by `SECUREDROP_ENV`).

deeplow:
Unfortunately the logs uploading failed in that [upgrade testing job](https://openqa.qubes-os.org/tests/186217/) and the video jumps directly to the end without the ability to see what happened:

video

However, running locally it failed in the same step because it coudn't start sd-viewer-debian-13:

Screenshot_2026-06-30_20-45-55

So I suspect there's some issue in the logic of removing `prohibit-startup`

deeplow:
Now I know why: `sd-base-debian-13` is set with `prohibit-start`, but only qubes previously set with with that get it removed. The thing is that `sd-viewer-debian-13` and `sd-inbox-debian-13` are cloned from it. Therefore, they inherit `prohibit-start` but don't have it removed. This should be an easy fix.

files/sdw-admin.py
```diff
@@ -136,20 +146,124 @@ def run_cmd(args: list[str]) -> None:
...
+    try:
+        yield
+    finally:
+        print("[info] Re-enabling preloaded disposables")
+
+        # Reset to original settings
+        dom0.features["preload-dispvm-max"] = original_preload_dispvm_max
+
+
+class template_upgrade_handler(ContextDecorator):
```
deeplow, line 171: NOTE: It'd like to make some tests for this (separately from the release — can go on another PR), but the fact that `sdw-admin.py` is a standalone python file rather than an proper Debian package has made that a bit awkward.

Shall we turn it into a python package (or include it in `sdw_updater`), for example?

legoktm: Yes, I think we should. IIRC we were kinda moving in this direction anyways with https://github.com/freedomofpress/securedrop-workstation/issues/1680 (one other script that can no longer be standalone).

I would suggest for tests though that you just do an importlib hack/workaround for now, and once we have tests then we can rework sdw-admin :)

files/sdw-admin.py
```diff
@@ -136,20 +147,127 @@ def run_cmd(args: list[str]) -> None:
...
+            q
+            for q in self.app.domains
+            if ("sd-workstation" in q.tags and q not in self.excluded and is_managed(q))
+        ]
+
+    def template_upgrades_needed(self) -> bool:
+        # NOTE: a more clever detection is warranted, but inspecting what salt is going
+        # to do is not a particular thing salt is good at. This is left here for when
+        # re-implemented with another IaC tool that doesn't need wrapper scripts like these.
+        templ_current_version_checks = [
```
legoktm, line 242: Does this also need to force a refresh?

files/sdw-admin.py
```diff
@@ -280,15 +398,17 @@ def perform_uninstall() -> None:
     )
 
 
-def is_managed(qube_name: str) -> bool:
+def is_managed(qube: str | QubesVM) -> bool:
```
legoktm, line 400: nit: can we just enforce this is called with QubesVM? are the remaining callers non-trivial to migrate?

files/sdw-admin.py
```diff
@@ -136,20 +147,127 @@ def run_cmd(args: list[str]) -> None:
...
+    dom0 = Qubes().domains["dom0"]
+    original_preload_dispvm_max = dom0.features.get("preload-dispvm-max", "0")
+
+    # Disable preloaded disposables
+    dom0.features["preload-dispvm-max"] = "0"
+
+    try:
+        yield
+    finally:
+        print("[info] Re-enabling preloaded disposables")
```
legoktm, line 165: This ends up showing up *after* the "Provisioning complete. Please reboot to complete the installation."

which means that provisioning wasn't actually complete. I think we may need to adjust that print statement to run later? outside of this context handler

legoktm (changes requested):
Still going through the test plan, main thing I've noticed so far is that "Provisioning complete" print happens before it actually finishes.

files/sdw-admin.py
```diff
@@ -136,20 +147,127 @@ def run_cmd(args: list[str]) -> None:
...
+
+    def template_upgrades_needed(self) -> bool:
+        # NOTE: a more clever detection is warranted, but inspecting what salt is going
+        # to do is not a particular thing salt is good at. This is left here for when
+        # re-implemented with another IaC tool that doesn't need wrapper scripts like these.
+        templ_current_version_checks = [
+            q.features["os-version"] == DEBIAN_VERSION
+            for q in Qubes().domains
+            if ("sd-workstation" in q.tags and q.klass == "TemplateVM")
+        ]
```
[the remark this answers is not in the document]
legoktm: No, this is fine as is.

files/sdw-admin.py
```diff
@@ -136,20 +147,127 @@ def run_cmd(args: list[str]) -> None:
...
+
+    def affected_qubes(self) -> list[QubesVM]:
+        # IMPORTANT: List of qubes may have changed
+        self.app.domains.refresh_cache(force=True)
+
+        return [
+            q
+            for q in self.app.domains
+            if ("sd-workstation" in q.tags and q not in self.excluded and is_managed(q))
+        ]
```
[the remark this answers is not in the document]
legoktm: fine as-is

securedrop_salt/sd-remove-unused-qubes.sls
```diff
@@ -5,8 +5,15 @@
...
 # This is because the workstation may have been offline for a while
 # and skipped some salt updates.
-{% for qube_name in ["sd-retain-logvm", "sd-whonix"] %}
+{% set qubes_for_removal = [
+  "sd-small-bookworm-template",
+  "sd-large-bookworm-template",
+  "sd-base-bookworm-template",
+  "sd-retain-logvm",
+  "sd-whonix",
+] %}
```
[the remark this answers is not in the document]
legoktm: fine as-is IMO

files/sdw-admin.py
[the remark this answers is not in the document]
deeplow: No idea what's going on here. Might be AI slop.

files/sdw-admin.py
```diff
@@ -136,20 +147,127 @@ def run_cmd(args: list[str]) -> None:
...
+            return False
+
+        print("[info] Re-enabling startup for managed qubes.")
+
+        # Obtain the list again since:
+        #  - some qubes may have been removed (e.g. old templates)
+        #  - somes cloned qubes may have inherited prohibit-startup
+        for qube in self.affected_qubes():
+            del qube.features["prohibit-start"]
+
```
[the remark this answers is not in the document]
legoktm: I believe the Qubes API already correctly handles this, right @deeplow ? it shouldn't raise KeyErrors if the key isn't present.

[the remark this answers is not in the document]
deeplow: In theory we're not removing prohibit start, but this could be a good catch in case a new qube is created during provisioning.

files/sdw-admin.py
deeplow: No because this runs right after instantiating a `qubesadmin.Qubes()` but that is assuming when such an object is instantiated, it get fresh state. Do you want me to add that just in case?

files/sdw-admin.py
deeplow: I considered that but it would require touching other things, and I decided against. But I don't particularly like these dynamic types. Basically in `configure()` the names are passed as literal strings (a retro-fit to what we had in the past).

If you'd like to, we could do that, but probably best if we did it as a follow-up fix, not necessarily part of the release, if that's OK with you.

securedrop_salt/sd-remove-unused-qubes.sls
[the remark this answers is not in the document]
deeplow: Here the `order: last` is what does the trick. If it partly failed, next time it runs, it's going to have to finish the "partial" setup. So ti should be fine.

deeplow:
I've pushed some fixes and I'm running the upgrade scenario on top of these: https://openqa.qubes-os.org/tests/186609

files/sdw-admin.py
deeplow: Good catch. How about 1169010? (no letters, heh). I'd rather avoid moving that out of the context manager and move the "Provisioning complete" print to later. Does that work?

legoktm: Perfect.

files/sdw-admin.py
legoktm: Yep

legoktm (approved):
This is good to go, but I'll wait for OpenQA to finish before merging

deeplow:
> I've pushed some fixes and I'm running the upgrade scenario on top of these: https://openqa.qubes-os.org/tests/186609

it hit #1735. I've restarted it as: https://openqa.qubes-os.org/tests/186612

deeplow:
The last [upgrade test](https://openqa.qubes-os.org/tests/187471#live) ran into https://github.com/freedomofpress/securedrop-workstation/issues/1735. I have restarted it as https://openqa.qubes-os.org/tests/187471#live.

deeplow:
The update test is now passing (on the expected tests) https://openqa.qubes-os.org/tests/187472#.

PR #1749, opened 2026-07-02
Rename sd-admin's template use `debian-13` in name

### LINKED ISSUE #1722: Changing template names: `small` and `large` are unclear

deeplow:
Follow the convention established in #1722:
- old: `sd-admin-trixie-template`
- new: `sd-admin-debian-13`

This way it's consistent with the other templates: sd-viewer-debian-13 and sd-inbox-debian-13

Fixes #1722

## Test plan

Prep:
1. sd-dev: `make build-admin-rpm`
2. dom0: 
   - `make clone-norpm`
   - `install install-admin-rpm`

Testing:
 - [ ]  in dom0: `make sd-admin` runs successfully
 - [ ] `sd-admin` is based on `sd-admin-debian-13` 

## Checklist

This change accounts for:
- [x] any necessary RPM packaging updates (e.g., added/removed files, see `rpm-build/SPECS/securedrop-workstation-dom0-config.spec`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,8 +128,8 @@ sd-admin: assert-dom0 ## Provision sd-admin VM and install securedrop-admin
 	@echo "Creating sd-admin template and AppVM..."
 	sudo qubesctl --show-output -- state.sls admin_salt.sd-admin
 	@echo "Installing packages inside sd-admin-debian-13..."
-	sudo qubesctl --show-output --skip-dom0 --targets sd-admin-trixie-template -- state.sls admin_salt.sd-admin-packages
-	qvm-shutdown --wait -- sd-admin-trixie-template
+	sudo qubesctl --show-output --skip-dom0 --targets sd-admin-debian-13 -- state.sls admin_salt.sd-admin-packages
+	qvm-shutdown --wait -- sd-admin-debian-13
 
 clone: assert-dom0 ## Builds rpm && pulls the latest repo from work VM to dom0
 	@./scripts/clone-to-dom0
--- a/admin_salt/sd-admin-packages.sls
+++ b/admin_salt/sd-admin-packages.sls
@@ -3,7 +3,7 @@
 
 ##
 # Configures the FPF apt repository and installs securedrop-admin
-# inside the sd-admin-trixie-template VM.
+# inside the sd-admin-debian-13 VM.
 #
 # Mirrors the approach in securedrop_salt/fpf-apt-repo.sls and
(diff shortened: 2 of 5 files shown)


zenmonkeykstop (approved):
LGTM based on visual review.

Issue #1539, opened 2026-01-16
test-created disposable causing side effects other tests when ran in parallel


deeplow:
I've observed [some flakiness](https://openqa.qubes-os.org/tests/164960#step/test_vm_sd_log/7) in other tests due to the intermediate (`sd-viewer-dvm`-based) disposable qube combined with the introduction of [parallel tests](https://github.com/freedomofpress/securedrop-workstation/pull/1512).

We may need another strategy like running that test separately or changing the test so it doesn't have side-effects.

<details><summary>See specific error</summary>

```
qube = <tests.base.QubeWrapper object at 0x739391f38ec0>
sdw_tagged_vms = <filter object at 0x739391f69990>

    def test_logs_are_flowing(qube, sdw_tagged_vms):
        """
        To test that logs work, we run a unique command in each VM we care
        about that gets logged, and then check for that string in the logs.
        """
        # Random string, to avoid collisions with other test runs
        token = "".join(secrets.choice(string.ascii_uppercase) for _ in range(10))
    
        # base template doesn't have sd-log configured
        # TODO: test a sd-viewer based dispVM
        skip = [f"sd-base-{DEBIAN_VERSION}-template", "sd-viewer"]
        # VMs we expect logs will not go to
        no_log_vms = ["sd-gpg", "sd-log"]
    
        # We first run the command in each VM, and then do a second loop to
        # look for the token in the log entry, so there's enough time for the
        # log entry to get written.
>       for vm in sdw_tagged_vms:

tests/test_vm_sd_log.py:71: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/base.py:70: in is_workstation_qube
    return is_managed_qube(qube) and SD_TAG in qube.tags
/usr/lib/python3.13/site-packages/qubesadmin/tags.py:65: in __contains__
    response = self.vm.qubesd_call(self.vm.name, 'admin.vm.tag.Get', elem)
/usr/lib/python3.13/site-packages/qubesadmin/base.py:76: in qubesd_call
    return self.app.qubesd_call(dest, method, arg, payload,
/usr/lib/python3.13/site-packages/qubesadmin/app.py:876: in qubesd_call
    return self._parse_qubesd_response(return_data)
[...]

PR #1754, opened 2026-07-07
Template upgrade unit tests

### LINKED ISSUE #1862: Test fixture `all_vms`: not updated during test execution
## Description

Some tests in https://github.com/freedomofpress/securedrop-workstation/pull/1754 surfaced an issue where the list of vms as presented by [`all_vms` fixture were outdated](https://openqa.qubes-os.org/tests/195054/modules/TestTemplateUpgradesAvailable/steps/1). Thus was due to a `session`-scoped fixture.

In theory this should not be a problem under the assumption that tests are self-contained and don't leave significant side-effects. However, preloaded disposables are a factor in the tests being introduced in https://github.com/freedomofpress/securedrop-workstation/pull/1754 and the dispXXXX is different at the end of the test than before.

## Steps to Reproduce

Run `CI=true make test` as of [`59c84e2`](https://github.com/freedomofpress/securedrop-workstation/tree/59c84e2/). There's a certain chance you'll hit the issue.

## Actual Behavior

```
E           qubesadmin.exc.QubesVMNotFoundError: No such domain: 'disp5649'
```

<details>

```
[...]
tests/test_sdw_admin.py::test_suppress_preloaded_disposables PASSED                                                                                                          [ 17%]
tests/test_sdw_admin.py::TestTemplateUpgradesAvailable::test_template_upgrade_handler ERROR                                                                                  [ 18%]
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

all_vms = <qubesadmin.app.VMCollection object at 0x736a60102a90>

    @pytest.fixture(scope="function")
    def sdw_tagged_vms(all_vms: VMCollection) -> list[QubesVM]:
        """Obtain all SecureDrop Workstation-exclusive qubes"""
>       return list(filter(is_workstation_qube, all_vms))

tests/conftest.py:129: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/base.py:99: in is_workstation_qube
    return is_managed_qube(qube) and SD_TAG in qube.tags
/usr/lib/python3.13/site-packages/qubesadmin/tags.py:64: in __contains__
    response = self.vm.qubesd_call(self.vm.name, 'admin.vm.tag.Get', elem)
/usr/lib/python3.13/site-packages/qubesadmin/base.py:96: in qubesd_call
    return self.app.qubesd_call(dest, method, arg, payload,
/usr/lib/python3.13/site-packages/qubesadmin/app.py:877: in qubesd_call
    return self._parse_qubesd_response(return_data)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

response_data = b"2\x00QubesVMNotFoundError\x00\x00No such domain: 'disp5649'\x00"

    @staticmethod
    def _parse_qubesd_response(response_data):
        '''Parse response from qubesd.
    
        In case of success, return actual data. In case of error,
        raise appropriate exception.
        '''
    
        if response_data == b'':
            raise qubesadmin.exc.QubesDaemonAccessError(
                'Got empty response from qubesd. See journalctl in dom0 for '
                'details.')
    
        if response_data[0:2] == b'\x30\x00':
            return response_data[2:]
        if response_data[0:2] == b'\x32\x00':
            (_, exc_type, _traceback, format_string, args) = \
                response_data.split(b'\x00', 4)
            # drop last field because of terminating '\x00'
            args = [arg.decode() for arg in args.split(b'\x00')[:-1]]
            format_string = format_string.decode('utf-8')
            exc_type = exc_type.decode('ascii')
            try:
                exc_class = getattr(qubesadmin.exc, exc_type)
            except AttributeError:
                if exc_type.endswith('Error'):
                    exc_class = __builtins__.get(exc_type,
                        qubesadmin.exc.QubesException)
                else:
                    exc_class = qubesadmin.exc.QubesException
            # TODO: handle traceback if given
>           raise exc_class(format_string, *args)
E           qubesadmin.exc.QubesVMNotFoundError: No such domain: 'disp5649'

/usr/lib/python3.13/site-packages/qubesadmin/base.py:130: QubesVMNotFoundError
```
</details>

## Comments

We need to make sure this is a function-scoped fixture. In the past session-scoped fixture could have been fined as tests were simply testing the deployed state of the system, but later tests were added to be able to test disposables (e.g. sd-viewer-based disposable), and those changed the picture.

deeplow:
Follow-up to https://github.com/freedomofpress/securedrop-workstation/pull/1744/changes

## Test plan

- [x] visual review
- [ ] failure injection
- [ ] tests pass

## Checklist

This change accounts for:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -36,10 +36,10 @@ def load_non_standard_module() -> Callable[[os.PathLike[Any]], ModuleType]:
     """
     Fixture factory for loading a non-standard python modules
 
-    This is necessary as a workaround due to the fact that some files not
-    following the standard python naming inventions, in particular:
-      1. Files ending in '.py'
-      2. No dashes ('-') in file names
+     This is necessary as a workaround for files that do not follow standard Python
+     module naming conventions, in particular:
+       1. Files NOT ending in '.py'
+       2. Dashes ('-') in file names
 
     Example:
 
@@ -47,7 +47,7 @@ def load_non_standard_module() -> Callable[[os.PathLike[Any]], ModuleType]:
 
         @pytest.fixture()
         def custom_module(load_non_standard_module):
-            return load_non_standard_module("custom_module", "/usr/bin/custom-module")
+            return load_non_standard_module("/usr/bin/custom-module")
 
         def test_foo(custom_module):
             custom_module.custom_fn()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -32,7 +32,7 @@ def proj_root() -> os.PathLike[Any]:
 
 
 @pytest.fixture(autouse=True)
-def load_non_standard_module() -> Callable:
+def load_non_standard_module() -> Callable[[os.PathLike[Any]], ModuleType]:
     """
     Fixture factory for loading a non-standard python modules
 
--- a/tests/test_sdw_admin.py
+++ b/tests/test_sdw_admin.py
@@ -1,8 +1,18 @@
-from collections.abc import Callable
+import os
+from collections.abc import Callable, Generator
 from pathlib import Path
 from types import ModuleType
+from typing import TYPE_CHECKING, Any
 
 import pytest
+import qubesadmin
+from qubesadmin.tests.mock_app import MockQube, QubesTestWrapper
+
+from tests.base import SD_TAG
+
+if TYPE_CHECKING:
+    from qubesadmin.app import VMCollection
+    from qubesadmin.vm import QubesVM
 
 
 @pytest.fixture
@@ -22,5 +32,218 @@ def sdw_admin(
     return load_non_standard_module(proj_root / "files" / "sdw-admin.py")
 
 
+@pytest.fixture
+def template_upgrades_available(sdw_admin: Any, mocker: Any) -> None:
+    """
+    Pretend that there are template upgrades available
+    """
+    mock_func = mocker.MagicMock()
+    mock_func.return_value = True
+    sdw_admin.template_upgrade_handler.template_upgrades_needed = mock_func
+
+
+@pytest.fixture
+def cleanup_prohibit_start(sdw_tagged_vms: "VMCollection") -> Generator:
+    """
+    Ensure cleanup in case of a test failure when testing with prohibit-start feature
+    """
+    yield
+    for qube in sdw_tagged_vms:
+        if qube.features.get("prohibit-start") == "disabled during set up":
+            del qube.features["prohibit-start"]
+
+
+@pytest.fixture
+def suppress_policies() -> Generator:
+    """
+    Temporarily suppress SDW RPC services to prevent qube startup
+
+    Qrexec services have the ability to start a qube. This may create
+    race-conditions during a VM startup.
+
+    NOTE: this might be https://github.com/freedomofpress/securedrop-workstation/issues/1751
+    """
+
+    policy_path = "/run/qubes/policy.d/20-sdw-override.policy"
+
+    with open(policy_path, "w") as policy_f:
+        policy_f.write(
+            "securedrop.Log           *  @tag:sd-workstation   sd-log    deny        notify=no\n"
+            "securedrop.GetSecretKeys *  sd-gpg                dom0      deny        notify=no\n"
+            "securedrop.Proxy         *  sd-app                sd-proxy  deny        notify=no\n"
+        )
+    yield
+    os.unlink(policy_path)
+
+
+@pytest.fixture
+def mock_qubes_app(sdw_admin: Any, mocker: Any) -> QubesTestWrapper:
+    """
+    Simulate a qubesadmin.Qubes() object called by sdw_admin
+    """
+
+    class MockQubesWorkstation(QubesTestWrapper):
+        def __init__(self) -> None:
+            super().__init__()
+
+            # 1. A few SD-specific qubes
+            sd_templates = [
+                MockQube(
+                    name="sd-debian-template-1",
+                    qapp=self,
+                    klass="TemplateVM",
+                    netvm="",
+                ),
+                MockQube(
+                    name="sd-debian-template-2",
+                    qapp=self,
+                    klass="TemplateVM",
+                    netvm="",
+                ),
+            ]
+            sd_app_qubes = [
+                MockQube("sd-qube-1", self, template=sd_templates[0]),
+                MockQube("sd-qube-2", self, template=sd_templates[1]),
+            ]
+
+            sd_disposables = [
+                MockQube("sd-disp-1", self, klass="DisposableVM", template=sd_app_qubes[0]),
+                MockQube("sd-disp-2", self, klass="DisposableVM", template=sd_app_qubes[1]),
+            ]
+
+            # Some default system qubes
+            default_template = MockQube(
+                name="default-template",
+                qapp=self,
+                klass="TemplateVM",
+                netvm="",
+            )
+            default_app_qube = MockQube(
+                name="default-qube",
+                qapp=self,
+                template=default_template,
+                netvm="sys-firewall",
+            )
+            default_disp = MockQube(
+                name="default-disp",
+                qapp=self,
+                template=default_app_qube,
+                netvm="",
+            )
+
+            self.update_vm_calls()
+
+            # Populate some expected qubesd queries
+
+            # Report these with the SD tag
+            for qube in sd_templates + sd_app_qubes + sd_disposables:
+                self.expected_calls[(qube.name, "admin.vm.tag.Get", SD_TAG, None)] = b"0\x001"
+
+            # Report these without the SD tag (if asked)
+            for qube in [self.domains["dom0"], default_template, default_app_qube, default_disp]:
+                self.expected_calls[(qube.name, "admin.vm.tag.Get", SD_TAG, None)] = b"0\x000"
+
+    mock_qubes_app = MockQubesWorkstation()
+
+    # Patch "Qubes()" to allow tests to run on this fake mock
+    qubes_mock = mocker.MagicMock()
+    qubes_mock.return_value = mock_qubes_app
+    sdw_admin.Qubes = qubes_mock
+
+    # yield the mock to allow for further modifications in tests
+    return mock_qubes_app
+
+
 def test_is_managed(sdw_admin: ModuleType) -> None:
     assert sdw_admin.is_managed("sd-app")
+
+
+@pytest.mark.provisioning
+def test_suppress_preloaded_disposables(sdw_admin: Any) -> None:
+    def get_preloaded_qubes() -> list["QubesVM"]:
+        return list(filter(lambda q: getattr(q, "is_preload", False), app.domains))
+
+    app = qubesadmin.Qubes()
+
+    # Save num of preloaded disposables
+    old_preload_dispvm_max = int(app.domains["dom0"].features["preload-dispvm-max"])
+    old_preload_disposables = get_preloaded_qubes()
+    assert old_preload_dispvm_max == len(old_preload_disposables) != 0
+
+    with sdw_admin.suppress_preloaded_disposables():
+        app.domains.refresh_cache(force=True)
+
+        # Ensure set back to 0 during contextual execution
+        assert int(app.domains["dom0"].features["preload-dispvm-max"]) == 0
+
+        # No preloaded disposables remain
+        assert len(get_preloaded_qubes()) == 0
+
+    # Value is set back to previous one
+    app.domains.refresh_cache(force=True)
+    new_preload_dispvm_max = int(app.domains["dom0"].features["preload-dispvm-max"])
+    new_preload_disposables = get_preloaded_qubes()
+    assert new_preload_disposables != old_preload_disposables
+    assert new_preload_dispvm_max == int(app.domains["dom0"].features["preload-dispvm-max"])
+
+
+class TestTemplateUpgradesAvailable:
+    def test_template_upgrade_handler(
+        self,
+        sdw_admin: Any,
+        template_upgrades_available: None,
+        suppress_policies: None,
+        cleanup_prohibit_start: None,
+    ) -> None:
+        # Start with an SDW qube
+        app = qubesadmin.Qubes()
+        sd_proxy = app.domains["sd-proxy"]
+        if sd_proxy.is_halted():
+            sd_proxy.start()
+
+        with sdw_admin.template_upgrade_handler():
+            # SDW qubes should have all be shut down
+            assert sd_proxy.is_halted()
+
+            # And they can no longer be started
+            with pytest.raises(qubesadmin.exc.QubesException) as exc_info:
+                sd_proxy.start()
+            assert "Qube start is prohibited" in str(exc_info.value)
+
+        # sd-proxy should start just fine (startup has been re-enabled)
+        sd_proxy.start()
+
+        # Shut down sd-proxy so it doesn't start sd-log through 'securedrop.Log'
+        # after test finishes
+        sd_proxy.shutdown()
+
+    @pytest.mark.parametrize(
+        ("template_ver", "expected_ver", "upgrades_needed"),
+        [
+            ("11", "13", True),  # Version jump
+            ("12", "13", True),  # Regular version bump
+            ("13", "13", False),  # Same version; no upgrade needed
+        ],
+    )
+    def test_template_upgrades_needed(
+        self,
+        template_ver: str,
+        expected_ver: str,
+        upgrades_needed: bool,
+        sdw_admin: Any,
+        mock_qubes_app: QubesTestWrapper,
+    ) -> None:
+        # Patch sdw-admin to expect a certain Debian version
+        sdw_admin.DEBIAN_VERSION = expected_ver
+
+        # Make templates return report a specific 'os-version' without
+        # actually messing with the system
+        for qube in mock_qubes_app.domains:
+            # Just tell is tagged as 'sd-workstation'
+            if qube.klass == "TemplateVM" and SD_TAG in qube.tags:
+                mock_qubes_app.expected_calls[
+                    (qube.name, "admin.vm.feature.Get", "os-version", None)
+                ] = b"0\x00" + template_ver.encode()
+
+        upgrade_handler = sdw_admin.template_upgrade_handler()
+        assert upgrade_handler.template_upgrades_needed() == upgrades_needed
--- a/tests/test_sdw_admin.py
+++ b/tests/test_sdw_admin.py
@@ -38,8 +38,8 @@ def template_upgrades_available(sdw_admin: Any, mocker: Any) -> None:
     Pretend that there are template upgrades available
     """
     mock_func = mocker.MagicMock()
-    mock_func.return_value = True
-    sdw_admin.template_upgrade_handler.template_upgrades_needed = mock_func
+    mock_func.return_value = False
+    sdw_admin.template_upgrade_handler.template_upgrades_skipped = mock_func
 
 
 @pytest.fixture
@@ -218,18 +218,18 @@ class TestTemplateUpgradesAvailable:
         sd_proxy.shutdown()
 
     @pytest.mark.parametrize(
-        ("template_ver", "expected_ver", "upgrades_needed"),
+        ("template_ver", "expected_ver", "should_upgrades_be_skipped"),
         [
-            ("11", "13", True),  # Version jump
-            ("12", "13", True),  # Regular version bump
-            ("13", "13", False),  # Same version; no upgrade needed
+            ("11", "13", False),  # Version jump
+            ("12", "13", False),  # Regular version bump
+            ("13", "13", True),  # Same version; no upgrade needed
         ],
     )
-    def test_template_upgrades_needed(
+    def test_template_upgrades_skipped(
         self,
         template_ver: str,
         expected_ver: str,
-        upgrades_needed: bool,
+        should_upgrades_be_skipped: bool,
         sdw_admin: Any,
         mock_qubes_app: QubesTestWrapper,
     ) -> None:
@@ -246,4 +246,4 @@ class TestTemplateUpgradesAvailable:
                 ] = b"0\x00" + template_ver.encode()
 
         upgrade_handler = sdw_admin.template_upgrade_handler()
-        assert upgrade_handler.template_upgrades_needed() == upgrades_needed
+        assert upgrade_handler.template_upgrades_skipped() == should_upgrades_be_skipped
--- a/tests/test_sdw_admin.py
+++ b/tests/test_sdw_admin.py
@@ -27,7 +27,7 @@ def sdw_admin(
     name and also not currently being in its own python module.
     """
 
-    # FIXME this is a workaroud. A better approach is to have sdw-admin in
+    # FIXME this is a workaround. A better approach is to have sdw-admin in
     # a proper python module, trivially importable in tests. See #1750.
     return load_non_standard_module(proj_root / "files" / "sdw-admin.py")
 
--- a/files/sdw-admin.py
+++ b/files/sdw-admin.py
@@ -184,7 +184,7 @@ class template_upgrade_handler(ContextDecorator):
     def __enter__(self) -> Callable[..., ContextDecorator]:
         self.app = Qubes()
 
-        self.skip_upgrade_handler = self.template_upgrades_needed()
+        self.skip_upgrade_handler = self.template_upgrades_skipped()
         if self.skip_upgrade_handler:
             return self
 
@@ -236,7 +236,7 @@ class template_upgrade_handler(ContextDecorator):
             if ("sd-workstation" in q.tags and q not in self.excluded and is_managed(q))
         ]
 
-    def template_upgrades_needed(self) -> bool:
+    def template_upgrades_skipped(self) -> bool:
         # NOTE: a more clever detection is warranted, but inspecting what salt is going
         # to do is not a particular thing salt is good at. This is left here for when
         # re-implemented with another IaC tool that doesn't need wrapper scripts like these.
--- a/stubs/qubesadmin/exc.pyi
+++ b/stubs/qubesadmin/exc.pyi
@@ -1,0 +1,3 @@
+# Type stubs for the `qubesadmin.exc` package.
+
+class QubesException(Exception): ...


legoktm:
Given #1762 I'm going to assume this will need to be reworked

tests/test_sdw_admin.py
```diff
@@ -0,0 +1,249 @@
...
+    policy_path = "/run/qubes/policy.d/20-sdw-override.policy"
+
+    with open(policy_path, "w") as policy_f:
+        policy_f.write(
+            "securedrop.Log           *  @tag:sd-workstation   sd-log    deny        notify=no\n"
+            "securedrop.GetSecretKeys *  sd-gpg                dom0      deny        notify=no\n"
+            "securedrop.Proxy         *  sd-app                sd-proxy  deny        notify=no\n"
+        )
+    yield
+    os.unlink(policy_path)
```
legoktm, line 76: I think this should be in a try/finally

legoktm (commented):
Nice. One pretty minor thing I noticed

deeplow:
I didn't actually have this assigned to me. That's why I had missed this for some time.

tests/test_sdw_admin.py
deeplow: Not sure I understand your suggestion. According to the [pytest docs](https://docs.pytest.org/en/stable/how-to/fixtures.html#teardown-cleanup-aka-fixture-finalization):

> Any teardown code for that fixture is placed after the yield.

So I'd expect it to honor the cleanup just as it would honor a `finally` statement.

deeplow: Oh. I see what you mean now with the context of the LLM comment bellow. We could do that. Please ignore my comment.

deeplow: The thing is that concurrency really screws this test up (and it might screw up other tests). Not sure there's much we can do here. If we we add a try: finally, it might remove the policy suppression set by another test in parallel.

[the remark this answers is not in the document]
deeplow: I don't think we can run tests concurrently without creating a per-test qubesd instance (which is what Qubes does). But they don't use pytest. Just unittest. Not sure what can be done here. In the short term a way to exclude tests from running in parallel would be good, but I don't know of an elegant solution for that.

deeplow:
Did an overdue rebase on this one

tests/conftest.py
```diff
@@ -21,7 +25,50 @@
...
-PROJ_ROOT = Path(__file__).parent.parent
+
+@pytest.fixture
+def proj_root() -> os.PathLike[Any]:
+    return Path(__file__).parent.parent
+
+
+@pytest.fixture(autouse=True)
+def load_non_standard_module() -> Callable:
+    """
```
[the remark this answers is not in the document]
deeplow: Done in 29c452c89

tests/conftest.py
```diff
@@ -21,7 +25,50 @@
...
+
+    .. code-block:: python
+
+        @pytest.fixture()
+        def custom_module(load_non_standard_module):
+            return load_non_standard_module("custom_module", "/usr/bin/custom-module")
+
+        def test_foo(custom_module):
+            custom_module.custom_fn()
+    """
```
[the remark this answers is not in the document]
deeplow: Fixed in c68de8a

tests/test_sdw_admin.py
```diff
@@ -0,0 +1,249 @@
...
+) -> ModuleType:
+    """
+    Equivalent to 'import sdw_admin', except as a pytest fixture.
+
+    Workaround needed due to 'sdw-admin.py' having a non-pythonic '-' in its
+    name and also not currently being in its own python module.
+    """
+
+    # FIXME this is a workaroud. A better approach is to have sdw-admin in
+    # a proper python module, trivially importable in tests. See #1750.
```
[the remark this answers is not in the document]
deeplow: Fixed in d630df2

tests/test_sdw_admin.py
```diff
@@ -0,0 +1,249 @@
...
+        # after test finishes
+        sd_proxy.shutdown()
+
+    @pytest.mark.parametrize(
+        ("template_ver", "expected_ver", "upgrades_needed"),
+        [
+            ("11", "13", True),  # Version jump
+            ("12", "13", True),  # Regular version bump
+            ("13", "13", False),  # Same version; no upgrade needed
+        ],
```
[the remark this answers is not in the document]
deeplow: This was indeed a problem. The main issue was with the function in sdw-admin itself. It implied it was checking if upgrades were available where in practice it did the opposite. Despite this, the implementation seems to do the correct thing. I have now fixed the name and updated the tests.

tests/test_sdw_admin.py
```diff
@@ -0,0 +1,249 @@
...
+
+@pytest.fixture
+def template_upgrades_available(sdw_admin: Any, mocker: Any) -> None:
+    """
+    Pretend that there are template upgrades available
+    """
+    mock_func = mocker.MagicMock()
+    mock_func.return_value = True
+    sdw_admin.template_upgrade_handler.template_upgrades_needed = mock_func
+
```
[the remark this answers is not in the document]
deeplow: See https://github.com/freedomofpress/securedrop-workstation/pull/1754/changes#r3721610511

tests/test_sdw_admin.py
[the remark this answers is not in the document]
legoktm: I think https://pytest-xdist.readthedocs.io/en/stable/distribution.html#:~:text=Tests%20are%20grouped%20by%20the%20xdist%5Fgroup%20mark does what we want. We can mark each test that access `qubesd` as in the same xdist group, and then xdist will schedule all those tests on the same worker, and within that worker, they'll run serially.

tests/conftest.py
```diff
@@ -21,7 +25,50 @@
     is_workstation_qube,
 )
 
-PROJ_ROOT = Path(__file__).parent.parent
+
+@pytest.fixture
```
legoktm: `ScopeMismatch: You tried to access the function scoped fixture proj_root with a session scoped request object. Requesting fixture stack:`

Think this just needs:

```suggestion
@pytest.fixture(scope="session")
```

tests/test_sdw_admin.py
```diff
@@ -0,0 +1,249 @@
...
+            # SDW qubes should have all be shut down
+            assert sd_proxy.is_halted()
+
+            # And they can no longer be started
+            with pytest.raises(qubesadmin.exc.QubesException) as exc_info:
+                sd_proxy.start()
+            assert "Qube start is prohibited" in str(exc_info.value)
+
+        # sd-proxy should start just fine (startup has been re-enabled)
+        sd_proxy.start()
```
legoktm, line 214: I don't understand why but openQA is complaining that sd-proxy no longer exists??

legoktm (changes requested):

tests/test_sdw_admin.py
deeplow: I'm very confused as well. We need to investigate this. For future reference, this happened in https://openqa.qubes-os.org/tests/191408 and here are the logs:

- [test_dom0-journalctl.log](https://github.com/user-attachments/files/30946548/test_dom0-journalctl.3.log)
- [test_dom0-var_log.tar.gz](https://github.com/user-attachments/files/30946550/test_dom0-var_log.1.tar.gz) (look under `var/xen/console/`)

deeplow: It doesn't explain the behavior but sd-proxy failed to boot due to some potential grsec issues.

```
[2026-08-05 12:49:18] Debian GNU/Linux 13 sd-proxy hvc0
[2026-08-05 12:49:18] 
[2026-08-05 12:49:18] sd-proxy login: [   28.118001] kauditd_printk_skb: 94 callbacks suppressed
[2026-08-05 12:49:22] [   28.118005] audit: type=1400 audit(1785948561.248:106): apparmor="ALLOWED" operation="open" class="file" profile="Xorg" name="/dev/xen/gntalloc" pid=899 comm="Xorg" requested_mask="wr" denied_mask="wr" fsuid=1000 ouid=0
[2026-08-05 12:49:22] [   28.659132] grsec: denied kernel module auto-load of snd_seq by uid 1000
[2026-08-05 12:49:22] [   28.723907] grsec: denied kernel module auto-load of snd_seq by uid 1000
         Stopping [0;1;39msession-c2.scope[0m - Session c2 of User user...
```


Also some other odd coincidences:
- qrexec timeout of a disposable right before sd-proxy is removed

tests/test_sdw_admin.py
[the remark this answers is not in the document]
deeplow: We actually had considered some of these options in https://github.com/freedomofpress/securedrop-workstation/issues/1539. If we always run through make targets, something like that could work.

tests/test_sdw_admin.py
deeplow: I have now restarted the tests to see if there's any difference.

deeplow:
Had forgotten to push that fix.

deeplow:
I think a significant amount of errors like mimetypes not working or even `/home/user/.mailcap` not existing is indicative of on-boot configuration failing. Things aren't working because they were not placed right locations. That doesn't explain the absence of `sd-proxy`.

But these "haunted tests" deserve a closer look.

legoktm:
Well I ran it locally and didn't disable xdist and something really did delete sd-proxy. I'm reprovisioning and then will try again with concurrency=1.

legoktm:
I'm still working through what exactly is causing the sd-proxy disappearance but can we split 58a19301583f1115106f9059b7e18cfe7ddd565f + e296159c351f61d7d74e073356e9a0a019336ea6 to their own PR just to shrink this one a bit?

deeplow:
> I'm still working through what exactly is causing the sd-proxy disappearance but can we split https://github.com/freedomofpress/securedrop-workstation/commit/58a19301583f1115106f9059b7e18cfe7ddd565f + https://github.com/freedomofpress/securedrop-workstation/commit/e296159c351f61d7d74e073356e9a0a019336ea6 to their own PR just to shrink this one a bit?

Done! https://github.com/freedomofpress/securedrop-workstation/pull/1839

legoktm:
Claude tracked down the failure to an upstream bug in which a named disposable w/ prohibit-start would fail to start, triggering the cleanup routine, which deletes the entire VM. Turns out this was already filed as https://github.com/QubesOS/qubes-issues/issues/11042 and Marek put in a revert ~20 hours ago.

deeplow:
<strike>I'll switch base now.</strike> turns out this is an automatic thing now. For reference the previous tests [were these](https://openqa.qubes-os.org/tests/194523). But checking on the [latest base Qubes image build](https://openqa.qubes-os.org/tests/192670#settings) it won't have the fix in it. We may need to force an update before the test runs.

deeplow:
I'm running doing an OpenQA run (https://openqa.qubes-os.org/tests/194556) where updates should be applied at the beginning. It should upload the package list on the first job and we can double-check that it has core-admin v4.3.51.

I'll just wait to confirm that and then pytest parallel runs, as we had discussed.

legoktm (dismissed):
Approved pending a good OpenQA run (which I just triggered); the upstream fix is in the stable repos now so we should be good to go...

deeplow:
Turns out `all_vms` was a session-scoped fixture, which was making the list of qubes diverge from the expected one, particularly for disposables. The issue is that this is a race condition and doesn't always show up locally. Let's see how it is in OpenQA. Even if it succeeds we could re-run just the `test_dom0` job to confirm it's not a fluke.

deeplow:
I had not seen it failing to install the keyring. We're not capturing the stderr, I'm afraid and I couldn't dig out anything in the logs.

Screenshot 2026-09-15 at 13-29-28 Qubes OS openQA Video


This was in https://openqa.qubes-os.org/tests/196324/

I'll just re-run.

deeplow:
Still failing on [`test_upgrade_handler`](https://openqa.qubes-os.org/tests/196330#step/TestTemplateUpgradesAvailable/1) but because it's not a normal python module, it's a bit hard to debug on the exact reason.


<details>

```
self = <tests.test_sdw_admin.TestTemplateUpgradesAvailable object at 0x70f20402dbd0>, sdw_admin = <module 'sdw_admin' from '/home/user/securedrop-workstation/files/sdw-admin.py'>, template_upgrades_available = None, suppress_policies = None, cleanup_prohibit_start = None

    def test_template_upgrade_handler(
        self,
        sdw_admin: Any,
        template_upgrades_available: None,
        suppress_policies: None,
        cleanup_prohibit_start: None,
    ) -> None:
        # Start with an SDW qube
        app = qubesadmin.Qubes()
        sd_proxy = app.domains["sd-proxy"]
        if sd_proxy.is_halted():
            sd_proxy.start()
    
>       with sdw_admin.template_upgrade_handler():

tests/test_sdw_admin.py:205: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
files/sdw-admin.py:204: in __enter__
    qube.features["prohibit-start"] = "disabled during set up"
/usr/lib/python3.13/site-packages/qubesadmin/features.py:49: in __setitem__
    self.vm.qubesd_call(self.vm.name, 'admin.vm.feature.Set', key,
/usr/lib/python3.13/site-packages/qubesadmin/base.py:96: in qubesd_call
    return self.app.qubesd_call(dest, method, arg, payload,
/usr/lib/python3.13/site-packages/qubesadmin/app.py:877: in qubesd_call
    return self._parse_qubesd_response(return_data)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

response_data = b"2\x00QubesVMNotFoundError\x00\x00No such domain: 'disp882'\x00"

    @staticmethod
    def _parse_qubesd_response(response_data):
        '''Parse response from qubesd.
    
        In case of success, return actual data. In case of error,
        raise appropriate exception.
        '''
    
        if response_data == b'':
            raise qubesadmin.exc.QubesDaemonAccessError(
                'Got empty response from qubesd. See journalctl in dom0 for '
                'details.')
    
        if response_data[0:2] == b'\x30\x00':
            return response_data[2:]
        if response_data[0:2] == b'\x32\x00':
            (_, exc_type, _traceback, format_string, args) = \
                response_data.split(b'\x00', 4)
            # drop last field because of terminating '\x00'
            args = [arg.decode() for arg in args.split(b'\x00')[:-1]]
            format_string = format_string.decode('utf-8')
            exc_type = exc_type.decode('ascii')
            try:
                exc_class = getattr(qubesadmin.exc, exc_type)
            except AttributeError:
                if exc_type.endswith('Error'):
                    exc_class = __builtins__.get(exc_type,
                        qubesadmin.exc.QubesException)
                else:
                    exc_class = qubesadmin.exc.QubesException
            # TODO: handle traceback if given
>           raise exc_class(format_string, *args)
E           qubesadmin.exc.QubesVMNotFoundError: No such domain: 'disp882'

/usr/lib/python3.13/site-packages/qubesadmin/base.py:130: QubesVMNotFoundError

```
</details>

deeplow:
Definitely some interesting edge-cases with preloaded disposables. I thought the issue with the race condition was that we were not waiting for them to start and shut down (after messing with `qvm-features dom0 preload-dispvm-max`), but there must be something else....

In testing locally, I finished the test with two disposables that are just hanging around

is_preload_issue


And then they mess up some tests

<details>

```
FAILED tests/test_vms_exist.py::test_all_sdw_vms_present - AssertionError: assert {'sd-devices', 'sd-devices-dvm', 'sd-viewer', 'sd-app', 'sd-viewer-debian-13', 'disp1731', 'sd-gpg', 'sd-inbox-debian-13', 'sd-base-debian-13', 'sd-proxy-dvm', 'sd-printers', 'sd-log', 'sd-proxy', 'disp1626'} == {'sd-viewer', 'sd-devices', 'sd-devices-dvm', 'sd-app', 'sd-viewer-debian-13', 'sd-gpg', 'sd-inbox-debian-13', 'sd-base-debian-13', 'sd-proxy-dvm', 'sd-printers', 'sd-proxy', 'sd-log'}
  
  Extra items in the left set:
  'disp1731'
  'disp1626'
  
  Full diff:
    {
  +     'disp1626',
  +     'disp1731',
        'sd-app',
        'sd-base-debian-13',
        'sd-devices',
        'sd-devices-dvm',
        'sd-gpg',
        'sd-inbox-debian-13',
        'sd-log',
        'sd-printers',
        'sd-proxy',
        'sd-proxy-dvm',
        'sd-viewer',
        'sd-viewer-debian-13',
    }
============================================================== 2 failed, 268 passed, 29 skipped in 716.94s (0:11:56) ===============================================================
make: *** [Makefile:170: test] Error 1
```
</details>

I think I have narrowed it down and filed as qubes-issues#11129. Oddly enough, on a local system this wouldn't happen if I only ran `test_sdw_admin.py`, making diagnosis harder. It would instead trigger sometimes when I ran all the tests. In the end it turned out the way to reproduce this is by running `test_suppress_preloaded_disposables` twice in a row.

I'll push a fix for this. Hopefully this is the last one, but since it's hard to reproduce the things that happen in OpenQA, I have no guarantee this will fix it.

deeplow:
It's [failing](https://openqa.qubes-os.org/tests/196566#step/test_sdw_admin/2) now in the new fixture, but it could be just a timing issue where in OpenQA it takes much longer for disposables to start and shut down. I'll increase the timeout.

deeplow:
Looks like [it passed](https://openqa.qubes-os.org/tests/196635). I'm just re-running the tests [here](https://openqa.qubes-os.org/tests/196920#dependencies) to confirm it's not a fluke.

legoktm:
Second run had one failure, which I think is the flaky(??) GetSecretKeys

```
# test_sd_proxy_services
# failure: 

# timestamp None
qube = <tests.base.QubeWrapper object at 0x706d285be9e0>

    @pytest.mark.configuration
    def test_sd_proxy_services(qube: QubeWrapper) -> None:
>       assert qube.service_is_active("securedrop-get-secret-keys")
E       AssertionError: assert False
E        +  where False = service_is_active('securedrop-get-secret-keys')
E        +    where service_is_active = <tests.base.QubeWrapper object at 0x706d285be9e0>.service_is_active

tests/test_vm_sd_gpg.py:124: AssertionError
```

tests/test_sdw_admin.py
```diff
@@ -17,10 +28,264 @@ def sdw_admin(
...
+def template_upgrades_available(sdw_admin: Any, mocker: Any) -> None:
+    """
+    Pretend that there are template upgrades available
+    """
+    mock_func = mocker.MagicMock()
+    mock_func.return_value = False
+    sdw_admin.template_upgrade_handler.template_upgrades_skipped = mock_func
+
+
+@pytest.fixture
```
legoktm, line 46: Maybe this should be `autouse=True` so it always gets cleaned up?

legoktm (approved):
Thanks, LGTM, hopefully some of the upstream issues get fixed so we can drop these workarounds

Issue #1763, opened 2026-07-09
updater: If full migration is needed, we shouldn't run dom0 states first


legoktm:
Our updater logic is basically like:
```
apply_dom0()
if migration_needed:
    apply_everything()
```

Which means that dom0 state always needs to be workable standalone (cf #1762) and when the full migration is needed, we end up running the dom0 states twice.

The logic really should just be:
```
if migration_needed:
    apply_everything()
else:
    apply_dom0()
```

deeplow:
I found another edge-case with this. When running dom0 highstate first and that includes a template upgrade, then `sd-viewer-debian-13` and `sd-inbox-debian-13` both fail to start. In short, with provisioning only, the qube's settings are set to boot with an in-vm kernel, which doesn't exist because the base template was never provisioned with it.

This was probably a situation we had never ran into because AFAIK before SDW 1.0.0 we stopped shipping our own templates (which come already with the kernel). See https://github.com/freedomofpress/securedrop-workstation/issues/969 for context.

Specifically, the templates (and all derived qubes) are failing with:

```
[user@dom0 securedrop-workstation]$ tail -f /var/log/xen/console/guest-sd-viewer-debian-13.log
[2026-07-10 19:30:10] /boot/grub2/grub.cfg.
[2026-07-10 19:30:10] .[H.[J.[1;1H.[2;32HGRUB version 2.12
[2026-07-10 19:30:10] 
[2026-07-10 19:30:10] .[?25h   Minimal BASH-like line editing is supported. For the first word, TAB   
[2026-07-10 19:30:10]    lists possible command completions. Anywhere else TAB lists possible   
[2026-07-10 19:30:10]    device or file completions. To enable less(1)-like paging, "set        
[2026-07-10 19:30:10]    pager=1".                                                              
[2026-07-10 19:30:10] 
[2026-07-10 19:30:10] 
[2026-07-10 19:30:10] grub> Logfile Opened
[2026-07-10 19:31:13] .[30m.[47mWelcome to GRUB!
[2026-07-10 19:31:13] 
[2026-07-10 19:31:13] .[37m.[40m.[37m.[40m.[37m.[40m.[3;34H      [ grub-xen.cfg  424B  100%  12.94KiB/s ].[3;1Herror: ../../grub-core/commands/search.c:472:no such device:
[2026-07-10 19:31:14] /boot/xen/pvboot-x86_64.elf.
[...]

PR #1764, opened 2026-07-10
Move upgrade script back into salt

deeplow:
Necessary due to updater bug https://github.com/freedomofpress/securedrop-workstation/issues/1763. In short, dom0 highstate needs to be fully independent until we fix it.

Implementation:
- have a template upgrade script (like the old bash one), but with a slightly different trick for precedence — use of `order:first` and `order: last`.
- Uses the python logic to shut down all qubes and prevent startup during provisioning.
  - no longer context-manager-based (script needs to be called twice: the start portion and finish portion)
  - Finish portion expected to always pass

## Test plan
- [ ] start with prod 1.7.1
- [ ] then set it to staging, but we cannot running `sdw-admin --apply`, so we need a hack (otherwise configuration will fail due missing packages)
  - `echo "copy_config()" | (cd /usr/bin/ && python3 -i sdw-admin --validate)`
- [ ] install local RPM from this branch by running `make clone && ./scripts/prep-dev && sudo qubesctl state.highstate`
  - This simulates what the updater does when applying the dom0 state 
- [ ] preloaded disposables work after running
- [ ] starting sd-qubes works (NOTE: we may not actually be able to do this due [to a bug](https://github.com/freedomofpress/securedrop-workstation/issues/1763#issuecomment-4938440282))
- [ ] Inject fault mid-provisiong (`qubesctl state.highstate`) and ensure the salt state `run-post-upgrade-scripts` still runs, regardless (therefore, re-enabling the templates and preloaded disposables)
 
## Checklist

This change accounts for:

### DIFF
--- a/securedrop_salt/template_upgrade_helper.py
+++ b/securedrop_salt/template_upgrade_helper.py
@@ -33,6 +33,16 @@ app = qubesadmin.Qubes()
 
 
 def start_template_upgrade() -> None:
+    """
+    Temporarily prevents startup of managed qubes
+
+    Necessary during provisioning, particularly in template changes, where
+    all qubes dependent on a template (including disposables only
+    indirectly based on it) need to be shut down, otherwise provisioning fails.
+
+    NOTE: deferred template changes may make this redundant
+    https://github.com/qubesos/qubes-issues/issues/8070
+    """
     # Exclude:
     #   - the ones already with prohibit-start for unrelated reason
     #   - preloaded disposables
--- a/securedrop_salt/template_upgrade_helper.py
+++ b/securedrop_salt/template_upgrade_helper.py
@@ -24,7 +24,7 @@ from sdw_updater.Updater import DEBIAN_VERSION
 # Logging set up
 log = logging.getLogger(Path(__file__).name)
 
-PROHIBIT_START_REASON = "disabled during set up"
+PROHIBIT_START_REASON = "SDW: disabled during set up"
 PRELOAD_MAX_TMP_FEAT_NAME = "preload-dispvm-max-saved"
 
 app = qubesadmin.Qubes()
--- a/securedrop_salt/template_upgrade_helper.py
+++ b/securedrop_salt/template_upgrade_helper.py
@@ -26,6 +26,12 @@ log = logging.getLogger(Path(__file__).name)
 
 PROHIBIT_START_REASON = "SDW: disabled during set up"
 
+# If the script gets abruptly executed at the right point we may loose
+# the actual number of disposable VMs. Either that or it was never set
+# for some unexplicable circumstance. Since 32GB is the recommended RAM
+# having 2 as the default (in case of breakage) it should be acceptable.
+SANE_N_DISPVMS = "2"
+
 # qubes-feature name to temporarily hold the number of preloeaded disposables
 PRELOAD_MAX_TMP_FEAT_NAME = "preload-dispvm-max-saved"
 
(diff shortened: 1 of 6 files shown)


securedrop_salt/template_upgrade_helper.py
```diff
@@ -0,0 +1,160 @@
...
+from pathlib import Path
+
+import qubesadmin
+
+from sdw_updater.Updater import DEBIAN_VERSION
+
+# Logging set up
+log = logging.getLogger(Path(__file__).name)
+
+PROHIBIT_START_REASON = "disabled during set up"
```
conorsch: Nit: let's prefix the `PROHIBIT_START_REASON` with a namespace, e.g. "SDW: temporarily disabled during upgrade".

securedrop_salt/template_upgrade_helper.py
```diff
@@ -0,0 +1,160 @@
...
+PRELOAD_MAX_TMP_FEAT_NAME = "preload-dispvm-max-saved"
```
conorsch, line 36: I had some trouble understanding the utility of `PRELOAD_MAX_TMP_FEAT_NAME`, even after having read the commit messages. Can we add an explanatory comment to assist future maintainers? 😇 There's a brief comment on L107 which I think better resides up here, at the declaration site.

securedrop_salt/template_upgrade_helper.py
```diff
@@ -0,0 +1,160 @@
...
+
+    NOTE: this must be safe to run even if start did not run (in case of partial failure)
+    """
+    log.info("Re-enabling preloaded disposables")
+    dom0 = app.domains["dom0"]
+
+    # Reset to original settings
+    dom0.features["preload-dispvm-max"] = dom0.features.get(PRELOAD_MAX_TMP_FEAT_NAME, "2")
+    if dom0.features.get(PRELOAD_MAX_TMP_FEAT_NAME) is None:
+        del dom0.features[PRELOAD_MAX_TMP_FEAT_NAME]
```
conorsch, line 156: Related to my comment above about not understanding the utility of "PRELOAD_MAX_TMP_FEAT_NAME", the logic in L132-134 is hard for me to understand: 

1. The value of "2" feels magical, and should be a constant with a comment. (Presumably the "2" value is a pragma carried over from other parts of the code with which I'm unfamiliar.)
2. The value of `dom0.features["preload-dispvm-max"]` is variously set to "2" or "0" throughout this script; shouldn't we standardize on one of those?
3. The if-clause is performing cleanup via `del` only if the temporary feature we set is unset... which it won't be, because we set it. 

Perhaps I'm reading the logic wrong, but that's a narration of me trying to follow the execution flow. Even if I misunderstood the code, let's comment it to explain the rationale better, which will make it easier to find bugs, both now and later.

conorsch (commented):
Jotted some notes during read-through. Seems like the challenge will be exercising the full upgrade logic, rather than testing the easy way, via clean install in a dev environment. Let's write down the manual testing steps to ground functional review.

securedrop_salt/template_upgrade_helper.py
deeplow: This used to be a [context_manager](https://github.com/freedomofpress/securedrop-workstation/pull/1764/changes#diff-6104c690fa1d1c1b216b58c8744593b8b49a9ec12c8c7eb4b8d101524cb02451L148-L169), which made much more sense than a bash script that's called twice. This was the reason why I needed to use some tricks to save the value in-between runs (like qubes-features).

In any case, I have added a comment now.

securedrop_salt/template_upgrade_helper.py
deeplow: I've added an explanation comment on where the 2 actually came from. This is just in to set a sane default dispvm number. Otherwise the script could fail because it reads `None` and explodes.

But you're right. This is harder to follow (see comment above about [the context manager](https://github.com/freedomofpress/securedrop-workstation/pull/1764#discussion_r3572390008)).

Let me know if you feel now the comments are a bit more clear.

deeplow:
Moved back to draft as it may not need at all if https://github.com/freedomofpress/securedrop-workstation/pull/1769 works.

deeplow:
Let's drop this one, given that the 1.8.0 release moved in a different direction (#1769).

Issue #1158, opened 2024-08-09
Canonical way of tracking SDW-managed system VMs


deeplow:
* [ ] I have searched for duplicates or related issues

## Description

Some default system app qubes are needed for a functional SecureDrop Workstation (e.g. `sys-net`, `sys-firewall`, `sys-usb`) and even some templates (`whonix-workstation-17`). What these qubes also have in common is that we don't have a canonical way to refer to them. Therefore we don't have a way to refer to them all.

This leads to awkward situations where we need to explicitly list them https://github.com/freedomofpress/securedrop-workstation/issues/758.

My proposal is to keep track of these qubes with a tag like `sd-workstation-required` so that we can then make tests that don't need to [call them out explicitly](https://github.com/freedomofpress/securedrop-workstation/blob/114d8ee18fe31a29a2960e9dcd8c6dacaf3f2339/tests/test_qubes_vms.py#L24) or even to have an updater knowing from the tag which [qubes also need ensured updates](https://github.com/freedomofpress/securedrop-workstation/issues/758).

## How will this impact [SecureDrop/SecureDrop Workstation users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?

No impact. Tags are not user-facing.

## How would this affect the SecureDrop Workstation [threat model](https://github.com/freedomofpress/securedrop-workstation/#threat-model)?

It shouldn't.

## User Stories

As a developer, I'd like not to have to rely on manual lists of VMs in various parts of the code. The least qube name references, the better for maintainability and security (to avoid forgetting one).

legoktm:
Not sure how this fits in exactly, but I want to propose a `sd-updater` tag, which is only applied to templateVMs that the updater should update. This would subsume #1598, as instead of looking for `sd-extras`, we would instruct people to tag their extra VMs with the updater tag.

deeplow:
That makes sense. And that works better since that unblocks (for the time being) the need to discover which MultiVM applications exist in the system. I have updated the proposal to reflect that.

PR #1769, opened 2026-07-13
Disable .top in postinst and re-enable in sdw-admin

### LINKED ISSUE #1762: Template switch needs to happen in dom0 salt state
> We'll need an rc3. I didn't assume the updater [would apply the dom0 before applying a migration](https://github.com/freedomofpress/securedrop-workstation/blob/1.7.1/sdw_updater/UpdaterApp.py#L283-L288)... With the solution implemented in #1744 the dom0 state can only be applied via `sdw-admin --apply`... This was only caught now because we cannot test via the update earlier than the release process.
> 
> The challenge here is that we can't really modify the updater as a fix, because it'll be running the 1.7.1 updater...
>  

 _Originally posted by @deeplow in [#1758](https://github.com/freedomofpress/securedrop-workstation/issues/1758#issuecomment-4927864270)_

Discussed and we'll just have to move this back into the dom0 state, in a Python script.

deeplow:
Work around issue #1763 with the simple approach of disabling the top file. The issue was that a dom0 highstate would always run before a migration update and this was blocking a release with template changes (they can't just run fully with a highstate due to the need to install the base template -> configure it to install the kernel -> clone as individual templates). Otherwise the qubes don't boot as they don't have a kernel installed.

This workaround simply disables the sd-workstation.top, to bypass the problematic issue with the updater and then re-enables it in sdw-admin apply.

Fixes #1762 (although not as the issue originally phrases it)

## Test plan

* [ ] Install a 1.7.1 prod system, but set config.json set to "staging" (so trixie debs are available)
* [ ] Fetch this branch, run `make clone` so it's in dom0 and there's a built RPM
* [ ] install the `rpm` and `createrepo-c` packages in your sd-dev VM (if fedora based, `rpm-sign` + `createrepo_c`)
* [ ] check out the "2026-07-yum-qa-local-upgrade" branch in this repo and get it in dom0 with `make clone-norpm` (don't clobber the previously built one)
* [ ] run the new `./scripts/local-rpm-server.sh` which should create a signing key and inject a local RPM repo
* [ ] run `sudo qubes-dom0-update --check-only` and see that it hits our local server and shows an update to 1.8.0-rc1
* [ ] launch the updater with `sdw-updater --skip-delta 0` and start it
* [ ] wait ...
* [ ] upgrade should finish successfully and all templates should be debian-13 based w/ bookworm templates gone 

## Checklist

This change accounts for:

### DIFF
--- a/files/sdw-admin.py
+++ b/files/sdw-admin.py
@@ -108,6 +108,12 @@ def provision_and_configure() -> None:
     """
     Applies the salt state.highstate on dom0 and all VMs
     """
+
+    # HACK: enable top as a workaround for #1763. In release 1.8.0 post-inst disabled
+    # the top file and it gets re-enabled in sdw-admin. This may be removed after the
+    # release.
+    run_cmd(["sudo", "qubesctl", "top.enable", "securedrop_salt.sd-workstation"])
+
     provision("Provisioning Fedora-based system VMs", "securedrop_salt.sd-sys-vms")
     provision("Provisioning base template", "securedrop_salt.sd-base-template")
     configure("Configuring base template", ["sd-base-debian-13"])
--- a/rpm-build/SPECS/securedrop-workstation-dom0-config.spec
+++ b/rpm-build/SPECS/securedrop-workstation-dom0-config.spec
@@ -191,7 +191,7 @@ install -m 644 securedrop_salt/apt-test_freedom_press.sources.j2 %{buildroot}/sr
 # Update Salt Configuration
 qubesctl saltutil.clear_cache -l quiet --out quiet > /dev/null || true
 qubesctl saltutil.sync_all refresh=true -l quiet --out quiet > /dev/null || true
-qubesctl top.enable securedrop_salt.sd-workstation > /dev/null ||:
+qubesctl top.disable securedrop_salt.sd-workstation > /dev/null ||:
 
 # Force full run of all Salt states - uncomment in release branch
 mkdir -p /tmp/sdw-migrations
--- a/securedrop_salt/sd-remove-unused-qubes.sls
+++ b/securedrop_salt/sd-remove-unused-qubes.sls
@@ -5,9 +5,9 @@
 # WARNING: only remove when complete reinstall is assumed (e.g. 1.0.0 release)
 # This is because the workstation may have been offline for a while
 # and skipped some salt updates.
+# FIXME: sd-small-bookworm-template and sd-large-bookworm-template are not removed
+# yet because the old updater still expects them to be around (#1771)
 {% set qubes_for_removal = [
-  "sd-small-bookworm-template",
-  "sd-large-bookworm-template",
   "sd-base-bookworm-template",
   "sd-retain-logvm",
   "sd-whonix",


deeplow:
Not sure what happened [in OpenQA](https://openqa.qubes-os.org/tests/188680): 
```
sd-base-debian-13: ERROR (exception No such domain: 'disp-mgmt-sd-base-debian-13')
```

I'll re-reun. I may dig into this later.

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -191,11 +191,11 @@ install -m 644 securedrop_salt/apt-test_freedom_press.sources.j2 %{buildroot}/sr
 # Update Salt Configuration
 qubesctl saltutil.clear_cache -l quiet --out quiet > /dev/null || true
 qubesctl saltutil.sync_all refresh=true -l quiet --out quiet > /dev/null || true
-qubesctl top.enable securedrop_salt.sd-workstation > /dev/null ||:
+qubesctl top.disable securedrop_salt.sd-workstation > /dev/null ||:
```
conorsch, line 194: I gather it's a temporary solution to disable the top file, so please add a comment in here explaining that we're disabling the top file in postinst for SDW Updater compatibility, as part of the 4.2 -> 4.3 migration.

files/sdw-admin.py
```diff
@@ -108,6 +108,12 @@ def provision_and_configure() -> None:
     """
     Applies the salt state.highstate on dom0 and all VMs
     """
+
+    # HACK: enable top as a workaround for #1763. In release 1.8.0 post-inst disabled
+    # the top file and it gets re-enabled in sdw-admin. This may be removed after the
+    # release.
+    run_cmd(["sudo", "qubesctl", "top.enable", "securedrop_salt.sd-workstation"])
```
conorsch, line 115: Force-enabling the top file in `sdw-admin` makes sense. Based on visual review, this looks like it'll behave how we want. We'll see want interactive testing shows, though: I'll set up a prod Workstation install, and stand by for testing.

conorsch (commented):
Looking forward to testing, once we have a test plan written. 

Also, this PR probably supersedes other potential solutions, like #1764, right?

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
deeplow: My idea would be just not back-porting this at all. https://github.com/freedomofpress/securedrop-workstation/issues/1763.

legoktm:
I've rebased this on top of the release/1.8.0 branch so it includes the migration flag bump that was missing on main.

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
legoktm: I think we have to keep it until we're confident that people are no longer running 1.7.1 right?

deeplow:
Makes sense. If you agree with not backporting then we just merge it straight into the release branch, then.

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
deeplow: Fair point. Then we should probably move it to the updater (right after dom0 apply). Otherwise we'd loose the dom0 highstates of every update.

deeplow: Never mind. The updater is not a good place either because it'll be running the outdated version. Then this probably means we need to keep this there AND make every release with the migration flag until then.

legoktm: When we modify the updater, we can change the location of where the migration flag is dropped, and then always stick it in the old location, so if you're on the old updater, it always kicks off the migration, but if you're on the fixed/new updater, it gets ignored.

But we can figure that out later...

legoktm:
I've written up the test plan that I'm following thanks to @deeplow's new local-rpm-server.sh script. I'm still mid-update though, it hasn't failed yet 🤞🏾 :)

legoktm:
New issue I think: the updater fails on the template upgrade step. I think that's because, per updater-detail.log, it's using the old list of templates, and is trying to update `... --targets sd-large-bookworm-template,sd-small-bookworm-template,fedora-43-xfce`.

legoktm:
Surely we had some way to handle that in the past?

deeplow:
OK. At least we're past the previous issue. For this one I have an issue as well :wink: (I mean, it's part of the issue): https://github.com/freedomofpress/securedrop-workstation/issues/1158#issuecomment-2769079036

deeplow:
> Surely we had some way to handle that in the past?

Probably. We good dig into git archaeology, but this seems to be in `sdw-admin`, so it should be an easy fix.

deeplow:
Let me see if I can get a fix in still today.

legoktm:
> but this seems to be in `sdw-admin`, so it should be an easy fix.

After looking at logs we realized this was in the updater, not sdw-admin so there isn't an easy fix. We're going to just not remove the bookworm templates for now, and will get rid of them later. We were concerned if this meant the inbox/viewer templates would not be fully updated, however the fpf-apt-repo.sls `upgrade-all-packages` step takes care of that, so we can be confident that the freshly provisioned templates are fully up to date already. And of course, the next updater run will have the new list of packages, so it'll be fine.

conorsch:
Test results in: looking good! Specifically, I used a QubesOS 4.3 [sic] machine, cleaning out a previously configured dev env, then followed the wonderful test plan below. 

* [x] Install a 1.7.1 prod system, but set config.json set to "staging" (so trixie debs are available)
* [x] Fetch this branch, run `make clone` so it's in dom0 and there's a built RPM
* [x] install the `rpm` and `createrepo-c` packages in your sd-dev VM (if fedora based, `rpm-sign` + `createrepo_c`)
* [x] check out the "2026-07-yum-qa-local-upgrade" branch in this repo and get it in dom0 with `make clone-norpm` (don't clobber the previously built one)
* [x] run the new `./scripts/local-rpm-server.sh` which should create a signing key and inject a local RPM repo
* [x] run `sudo qubes-dom0-update --check-only` and see that it hits our local server and shows an update to 1.8.0-rc1
* [x] launch the updater with `sdw-updater --skip-delta 0` and start it
* [x] wait ... 
* :x: upgrade should finish successfully and all templates should be debian-13 based w/ bookworm templates gone 
  * the bookworm templates are still present, but that's expected, given @legoktm's addition here. So I'm flagging this as a deviation of expectations according to the test plan, but still technically correct behavior, and therefore not a merge blocker.

As additional info I include what the SDW VMs on my system were, before and after. 

<details>
<summary>
`qvm-ls --tags sd-workstation` before update
</summary>

```
NAME                        STATE    CLASS       LABEL   TEMPLATE                    NETVM
disp5260                    Paused   DispVM      green   sd-viewer                   -
disp7788                    Paused   DispVM      green   sd-viewer                   -
sd-app                      Halted   AppVM       yellow  sd-small-bookworm-template  -
sd-base-bookworm-template   Halted   TemplateVM  black   -                           -
sd-devices                  Halted   DispVM      red     sd-devices-dvm              -
sd-devices-dvm              Halted   AppVM       red     sd-large-bookworm-template  -
sd-gpg                      Running  AppVM       purple  sd-small-bookworm-template  -
sd-large-bookworm-template  Halted   TemplateVM  black   -                           -
sd-log                      Running  AppVM       red     sd-small-bookworm-template  -
sd-printers                 Halted   DispVM      red     sd-devices-dvm              -
sd-proxy                    Running  DispVM      blue    sd-proxy-dvm                sys-firewall
sd-proxy-dvm                Halted   AppVM       blue    sd-small-bookworm-template  sys-firewall
sd-small-bookworm-template  Halted   TemplateVM  black   -                           -
sd-viewer                   Halted   AppVM       green   sd-large-bookworm-template  -

```
</details>

<details>
<summary>
`qvm-ls --tags sd-workstation` after update
</summary>

```
NAME                        STATE    CLASS       LABEL   TEMPLATE             NETVM
disp6890                    Paused   DispVM      green   sd-viewer            -
disp7000                    Paused   DispVM      green   sd-viewer            -
sd-app                      Running  AppVM       yellow  sd-inbox-debian-13   -
sd-base-debian-13           Halted   TemplateVM  black   -                    -
sd-devices                  Halted   DispVM      red     sd-devices-dvm       -
sd-devices-dvm              Halted   AppVM       red     sd-viewer-debian-13  -
sd-gpg                      Running  AppVM       purple  sd-inbox-debian-13   -
sd-inbox-debian-13          Halted   TemplateVM  black   -                    -
sd-large-bookworm-template  Halted   TemplateVM  black   -                    -
sd-log                      Running  AppVM       red     sd-inbox-debian-13   -
sd-printers                 Halted   DispVM      red     sd-devices-dvm       -
sd-proxy                    Running  DispVM      blue    sd-proxy-dvm         sys-firewall
sd-proxy-dvm                Halted   AppVM       blue    sd-inbox-debian-13   sys-firewall
sd-small-bookworm-template  Halted   TemplateVM  black   -                    -
sd-viewer                   Halted   AppVM       green   sd-viewer-debian-13  -
sd-viewer-debian-13         Halted   TemplateVM  black   -                    -

```
</details>

Additionally, I made sure that post update and post reboot I was able to log into the Inbox and connect to a server, validating the bare minimum of connectivity throughout the SDW VMs.

Nice work on finding such a tiny intervention to solve this problem, @deeplow!

conorsch (approved):
Approving based on functional review (see test plan above). 

I've not investigated the OpenQA failure. Happy to test again if more changes are added before the final state. 

Ten thousand thanks for the super clear test plan. :pray:

conorsch:
> run the new `./scripts/local-rpm-server.sh` which should create a signing key and inject a local RPM repo

Absolutely bonkers how much of a timesaver this script was. :metal:

legoktm:
> the bookworm templates are still present, but that's expected, given @legoktm's addition here. So I'm flagging this as a deviation of expectations according to the test plan, but still technically correct behavior, and therefore not a merge blocker.

Yep sorry I didn't update the test plan after we made that switch.

legoktm (approved):
Thanks Conor for testing; my upgrade was also successful so let's move forward here and onto rc2

PR #1774, opened 2026-07-15
Skip test_mock_device_attach_deny for now because it's broken

legoktm:
It's obscuring other failures that we should be paying more attention to.

## Checklist

This change accounts for:

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -359,7 +359,7 @@ class Test_SD_VM_Common:
                 actual_app == expected_app
             ), f"MIME type {mime_type}: expected {expected_app}, got {actual_app}"
 
-    @pytest.mark.skip(reason="broken")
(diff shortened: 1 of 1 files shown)


tests/base.py
```diff
@@ -359,6 +359,7 @@ def test_mime_types(self, qube: QubeWrapper) -> None:
                 actual_app == expected_app
             ), f"MIME type {mime_type}: expected {expected_app}, got {actual_app}"
 
+    @pytest.mark.skip(reason="broken")
```
deeplow: Nit:

```suggestion
    @pytest.mark.skip(reason="Broken in CI. Tracked as #1579")
```

deeplow: (Note: this technically runs in local machines)

deeplow (approved):

PR #1777, opened 2026-07-16
SecureDrop Workstation 1.8.0

legoktm:
Fixes #

## Test plan

* [ ] changelog is updated
* [ ] version is correct
## Checklist

This change accounts for:


vickiniu (approved):

changelog.md
```diff
@@ -1,17 +1,12 @@
...
-
-## 1.8.0-rc1
+## 1.8.0
 
 This release focused on upgrading SecureDrop Workstation qubes from Debian 12 (Bookworm)
 to Debian 13 (Trixie). Some document viewer applications have been replaced and others
 updated with major visual changes.
 
-* Update SDW templates to Debian 13 (Trixie) (#1719, #1739)
+* Update SDW templates to Debian 13 (Trixie) (#1719, #1739, #1759)
```
deeplow, line 9: (just catching up on this now, since I was already out)

Nit: #1759 was the migration flag and normally we don't include these. But I don't think it's critical to change it here.

If anything, having it here, may be a good thing as we may need to "backport" this flag as we may need to keep this in for a few releases until we're confident all instances have migrated to 1.8.0. To quote myself in https://github.com/freedomofpress/securedrop-workstation/pull/1772#issue-4892880260:

> * This also means that we may need to make every workstation release a "migration" release. Otherwise `sdw-admin.py` does not run and that is what re-enables the top file again ⚠️. Does it make sense to through it in here? Or as a separate PR. If separate, CI may fail.

rpm-build/SPECS/securedrop-workstation-dom0-config.spec
deeplow, line 231: We missed these. We may want to remove these two, at least in `main`, when backporting the changelog.

Issue #1766, opened 2026-07-13
Updater core logic re-design (when moving to ansible)


deeplow:
This is the continuation of past conversations:
- https://github.com/freedomofpress/dangerzone/issues/530#issuecomment-3057794327
- [MultiVM Applications (Qubes Summit 2023 design session)](https://cfp.3mdeb.com/qubes-os-summit-2023/talk/9RYPUA/)
- [MultiVM Application Requirements.pdf](https://github.com/user-attachments/files/29966539/MultiVM.Application.Requirements.pdf) ­— a document I put together in October 2024 about MultiVM Life Cycle management.

## Description
The updater is a mission critical element of the workstation. The current design has had some critical fractures over the years and with the pending move [from salt to ansible](https://github.com/freedomofpress/securedrop-workstation/issues/1354) this is going to be necessary any ways.

### Requirements:

<details><summary>Current requirements (non exhaustive list)</summary>

- Enforce SecureDrop's update policies:
  - Only enable start of SecureDrop inbox after all updates
  - Enforce dom0 restart after dom0 update installed
  - Enforce VM restarts for those whose template received updates
  - Explain why restarts are needed
- Notify users in case of long-running (sdw-notify)  
- Show progress information
</details>

Proposed new requirements:
- [ ] **MultiVM-Agnostic** — We already have two distinct MultiVM applications `sdw-workstation` and [`sdw-admin`](https://github.com/freedomofpress/securedrop-workstation/pull/1572/changes). Dangerzone could also make use of it and overall it would be great if these were upstreamed into the official updater itself (even if we still wrap around it to enforce our more strict update policies)
   - [ ] Ability to detect when a MultiVM has pending architectural updates (without assuming dom0 RPM, necessarily)
   - [ ] Ability to detect MultiVM applications (hard-coding is fine initially)
      - https://github.com/freedomofpress/securedrop-workstation/issues/1598 — or a way (ideally GUI) for users to add other VMs to be included in updates.
      - https://github.com/freedomofpress/securedrop-workstation/issues/1791 — updater handling multiple MultiVM applications (for the concrete case of `sd-admin` and `sd-workstation`). 
- [ ] **Fine-grained progress reporting** — Salt was quite opaque in term of progress (all or nothing) and the project appeared unwilling to [address this](https://github.com/saltstack/salt/issues/46665) and sometimes our current updater may appear [to be stuck](https://github.com/freedomofpress/securedrop-workstation/issues/1394).
[...]

PR #1785, opened 2026-07-20
Track what version of the updater invokes sdw-admin and salt

### LINKED ISSUE #1778: Expose updater version to everything it invokes (salt, sdw-admin, etc.) as a back-compat failsafe

## Description

I propose that for updater expose a `SD_UPDATER_VERSION` environment variable for every command it shells out to (or maybe a CLI arg) as a failsafe for when we need to work around an updater bug. Specifically this would expose the *old* version of the code, even if a new RPM was installed with new sdw-admin code

For example, in https://github.com/freedomofpress/securedrop-workstation/issues/1771 we were unable to remove the bookworm templates because the updater relied on them, so we just had to disable that part.

If we had this environment variable we could guard it on SD_UPDATER_VERSION >= 1.8.0 and if so, do the removal. This primarily allows us to work around the issue where we can't introduce code that relies on the new behavior until we're sure that everyone is off of the buggy version. We don't have any tracking/analytics to verify that, nor do we necessarily want to brick a machine that hasn't been used in a month or two. So instead we can just check if we're past the buggy version, and if so, do the thing we wanted.

legoktm:
The updater now passes its old version to sdw-admin and salt so in case we discover other updater bugs in the future, we can conditionally work around them as needed.

Fixes #1778.

## Test plan

1.7.1 case:
* [ ] Install 1.7.1, run sdw-admin --apply
* [ ] Build this RPM and spin up local-rpm-server
* [ ] Run through the 1.7.1 updater, it should finish successfully and correctly leave the bookworm templates behind

1.8.0 case:
* [ ] Install 1.7.1, run sdw-admin --apply
* [ ] Update to prod 1.8.0, i.e. have 1.8.0 updater plus bookworm templates
* [ ] Build this RPM and spin up local-rpm-server
* [ ] Manually drop the migration flag in /tmp
* [ ] Run through the 1.8.0 updater, it should finish successfully and remove the bookworm templates

## Checklist

This change accounts for:


sdw_updater/Updater.py
```diff
@@ -426,7 +428,7 @@ def apply_dom0_state() -> UpdateStatus:
     cmd = ["sudo", "qubesctl", "--show-output", "state.highstate"]
     cmd_for_log = " ".join(cmd)
     try:
-        output = subprocess.check_output(cmd)
+        output = subprocess.check_output(cmd, env={**os.environ, "SD_UPDATER_VERSION": __version__})
```
legoktm: Claude flagged that this is likely to not work since sudo won't pass the environment variable through. I'm going to switch to a CLI arg.

files/sdw-admin.py
```diff
@@ -88,9 +88,36 @@ def parse_args() -> argparse.Namespace:
...
+    3) An old updater, which conservatively is 1.7.0 as the oldest
+       potential updater we need to support.
+    """
+    if cli_updater_version:
+        return cli_updater_version
+    elif sys.stdin.isatty():
+        # direct sdw-admin invocation
+        return __version__
+    else:
+        return "1.7.0"
```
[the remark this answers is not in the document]
legoktm: Switching to the cmdline approach.

files/sdw-admin.py
```diff
@@ -104,20 +131,39 @@ def copy_config() -> None:
...
+    # preloaded dispvms. Needed due to inclusion of 'qvm.preload-disposables'
+    # indirectly via 'qvm.default-dvm'. Removing this does not mean preloaded
+    # disposables are disabled. Just that they don't get enabled on provisioning.
+    # FIXME: https://github.com/freedomofpress/securedrop-workstation/issues/1523
+    pillar: dict[str, dict] = {
+        "qvm": {"dom0": {"preload": False}},
+        # Pass through the updater's version to salt for any back-compat workarounds
+        "sd": {"updater_version": updater_version},
+    }
+    return pillar
```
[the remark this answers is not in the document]
legoktm: Dropped the type hint entirely

securedrop_salt/sd-remove-unused-qubes.sls
```diff
@@ -5,14 +5,21 @@
...
   "sd-whonix",
 ] %}
 
+# We can only remove these once we're on a >=1.8.0 updater.
+{% set updater_version = salt['pillar.get']('sd:updater_version', None) %}
+{% if updater_version and salt['pkg.version_cmp'](updater_version, '1.8.0') >= 0 %}
+{% do qubes_for_removal.extend([
+  "sd-small-bookworm-template",
+  "sd-large-bookworm-template",
+]) %}
```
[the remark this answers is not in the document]
legoktm: Seems fine as-is.

legoktm:
I need to test the swap to /proc/{id}/cmdline before this is ready for review.

legoktm:
@deeplow just a request for review on the general approach before I spend more time testing / fixing this up.

files/sdw-admin.py
```diff
@@ -88,9 +88,52 @@ def parse_args() -> argparse.Namespace:
...
+    3) A direct sdw-admin invocation (e.g. during fresh install), in which case
+       its own version is the relevant one.
+    """
+    if cli_updater_version:
+        return cli_updater_version
+    elif _invoked_by_updater():
+        return "1.7.0"
+    else:
+        # direct sdw-admin invocation
+        return __version__
```
deeplow, line 134: Not sure if we need to use its own version. As long we cover this test case, not having an updater_version could be helpful as well to know it was not called by the updater. And I think your salt code already guards against this.

sdw_updater/Updater.py
```diff
@@ -423,7 +424,15 @@ def apply_dom0_state() -> UpdateStatus:
...
-    cmd = ["sudo", "qubesctl", "--show-output", "state.highstate"]
+    # Pass our own version so Salt states can work around bugs in old updaters.
+    pillar = {"sd": {"updater_version": __version__}}
+    cmd = [
+        "sudo",
+        "qubesctl",
+        "--show-output",
+        "state.highstate",
+        f"pillar={json.dumps(pillar)}",
+    ]
```
deeplow, line 435: This made me think that maybe we shouldn't have the updater interface with salt at all. What if from here we called `sdw-admin --apply --quick` or something? `sdw-admin` as the sole interface with salt sounds conceptually simpler.

The advantage here would be that you we wouldn't need to also pass pillar data. It's not that big of a difference.

What do you think about this idea? Or do you think this would be a discussion with larger implications?

files/sdw-admin.py
```diff
@@ -88,9 +88,52 @@ def parse_args() -> argparse.Namespace:
         action="store_true",
         help="Configure SecureDrop Workstation",
     )
+    parser.add_argument(
+        "--updater-version",
+        default=None,
+        required=False,
+        help=("Version of the updater invoking sdw-admin. Do not manually set."),
+    )
```
deeplow, line 96: I think there's an option to make this a hidden argument. No need for users to see this option if they call `--help`, I think.

files/sdw-admin.py
```diff
@@ -88,9 +88,52 @@ def parse_args() -> argparse.Namespace:
...
+    """
+    Determine the version of the updater orchestrating this run.
+
+    There are three main scenarios here:
+
+    1) A newer updater, which passed in --updater-version
+    2) An old updater, which shells out to `sdw-admin --apply` without the flag;
+       conservatively assume 1.7.0, the oldest updater we need to support.
+    3) A direct sdw-admin invocation (e.g. during fresh install), in which case
+       its own version is the relevant one.
```
deeplow, line 126: I find the cmdline potentially brittle, but I coudn't think of anything better either.

sdw_updater/Updater.py
legoktm: That's a partial step to the idea I pitched in https://github.com/freedomofpress/securedrop-workstation/issues/1766#issuecomment-5046602740 but I don't think that fully helps us here because even if we had `sdw-admin --apply --quick`, we still need to know which version of the updater invokes it, so we know if we can drop the bookworm templates or not.

deeplow: > I don't think that fully helps us here because even if we had sdw-admin --apply --quick, we still need to know which version of the updater invokes it, so we know if we can drop the bookworm templates or not.

True. I guess my point was that we wouldn't need to repeat the pillar data in two places. Otherwise we're duplicating the pillar data in the updater as well as in `sdw-admin.py`. It's pretty small now and we will move away from salt, but it's still duplicate parameters that have to stay in sync and it would be easier if they're just in one place.

And since we're in agreement with having that, my question is: would it be worth doing as a separate PR? Or shall we not implement such a thing as we'll be reworking this all soon.

PR #1793, opened 2026-07-21
Obtain list of templates to update on the fly instead of hardcoding

### LINKED ISSUE #758: Updater should refresh list of target VMs after dom0 update

### LINKED ISSUE #1771: Updater uses old, hardcoded list of templates even when new templates are provisioned
New issue I think: the updater fails on the template upgrade step. I think that's because, per updater-detail.log, it's using the old list of templates, and is trying to update `... --targets sd-large-bookworm-template,sd-small-bookworm-template,fedora-43-xfce`.

_Originally posted by @legoktm in https://github.com/freedomofpress/securedrop-workstation/issues/1769#issuecomment-4973540583_

legoktm:
If the updater has a hardcoded list of templates, that means you cannot remove the old template until everyone is off the old updater version.

Instead, fetch the list dynamically so it always uses the correct set of templates. We look at everything tagged `sd-workstation`, plus a hardcoded set of sys-* VMs that use Fedora, figure out the underlying template, and update those.

Fixes #758.
Fixes #1771.

## Test plan

* [ ] Install 1.7.1 RPM and `sdw-admin --apply`
* [ ] Build RPM based on this branch and install it into dom0 `sudo dnf install ./whatever.rpm`
* [ ] `sdw-updater --skip-delta 0` -> this is basically simulating the updater as if the fixed updater was being used for trixie switch
* [ ] observe in updater logs, the "Applying all updates to VMs" line only lists the newly created debian-13 templates, not the bookworm ones

## Checklist

This change accounts for:

### DIFF
--- a/launcher/tests/conftest.py
+++ b/launcher/tests/conftest.py
@@ -56,35 +56,79 @@ skip_in_dom0 = pytest.mark.skipif(
 )
 
 
+SD_TAG = "sd-workstation"
+MOCK_FEDORA_TEMPLATE = "fedora-XX-xfce"
+
+# SecureDrop-managed TemplateVMs, all tagged `sd-workstation`.
+MOCK_SDW_TEMPLATES = [
+    "sd-base-debian-XX",
+    "sd-inbox-debian-XX",
+    "sd-viewer-debian-XX",
+]
+
+# SecureDrop-managed AppVMs (tagged) mapped to their template.
+MOCK_SDW_APPVMS = {
+    "sd-app": "sd-inbox-debian-XX",
+    "sd-log": "sd-inbox-debian-XX",
+    "sd-proxy": "sd-inbox-debian-XX",
+    "sd-gpg": "sd-inbox-debian-XX",
+    "sd-viewer": "sd-viewer-debian-XX",
+    "sd-devices": "sd-viewer-debian-XX",
+    "sd-printers": "sd-viewer-debian-XX",
+}
+
+
 @pytest.fixture
 def mocked_qubes_app(mocker):
+    from qubesadmin.tests import mock_app
     from qubesadmin.tests.mock_app import MockQube, QubesTestWrapper
+
+    # The mock only registers `admin.vm.tag.Get` calls for tags it knows about,
+    # so register `sd-workstation` to allow `"sd-workstation" in vm.tags` checks
+    # against untagged VMs (e.g. the Fedora template).
+    if SD_TAG not in mock_app.POSSIBLE_TAGS:
+        mock_app.POSSIBLE_TAGS.append(SD_TAG)
 
     class MockQubesWorkstation(QubesTestWrapper):
         def __init__(self):
             super().__init__()
 
-            # FIXME avoid using module under test. Obtain directly from main tests
-            current_vms = sdw_updater.Updater._get_current_vms()
-
-            # 1. create the templates
-            for template_name in set(current_vms.values()):
+            # 1. Create the SecureDrop templates (tagged `sd-workstation`)
+            for template_name in MOCK_SDW_TEMPLATES:
                 self._qubes[template_name] = MockQube(
                     name=template_name,
                     qapp=self,
                     klass="TemplateVM",
                     netvm="",
+                    tags=[SD_TAG],
                 )
 
-            # 2. create the app qubes
-            for qube_name, template_name in current_vms.items():
+            # 2. Create the Fedora template backing the sys-* VMs (NOT tagged)
+            self._qubes[MOCK_FEDORA_TEMPLATE] = MockQube(
+                name=MOCK_FEDORA_TEMPLATE,
+                qapp=self,
+                klass="TemplateVM",
+                netvm="",
+            )
+
+            # 3. Create the SecureDrop app qubes (tagged `sd-workstation`)
+            for qube_name, template_name in MOCK_SDW_APPVMS.items():
                 MockQube(
                     qube_name,
                     self,
                     template=template_name,
+                    tags=[SD_TAG],
                 )
 
-            # 3. TODO Lastly create the disposables
+            # 4. Create the sys-* VMs (NOT tagged) based on the Fedora template
+            for sys_vm in sdw_updater.Updater.SYSTEM_VMS:
+                MockQube(
+                    sys_vm,
+                    self,
+                    template=MOCK_FEDORA_TEMPLATE,
+                )
+
+            # 5. TODO Lastly create the disposables
 
             self.update_vm_calls()
 
--- a/launcher/tests/test_updater.py
+++ b/launcher/tests/test_updater.py
@@ -12,6 +12,12 @@ import pytest
 from sdw_updater import Updater
 from sdw_updater.Updater import UpdateStatus
 
+skipif_no_qubesadmin = pytest.mark.skipif(
+    importlib.util.find_spec("qubesadmin") is None,
+    reason="qubesadmin module not available (only runs in dom0)",
+)
+
+
 debian_based_vms = [
     "sd-app",
     "sd-log",
@@ -50,12 +56,13 @@ TEST_RESULTS_UPDATES = {
 }
 
 
-def test__get_current_vms():
-    assert len(Updater._get_current_vms()) == 8
-
-
-def test__get_current_templates():
-    assert len(Updater._get_current_templates()) == 3
+@skipif_no_qubesadmin
+def test__get_current_templates(mocked_qubes_app):
+    assert Updater._get_current_templates() == {
+        "sd-inbox-debian-XX",
+        "sd-viewer-debian-XX",
+        "fedora-XX-xfce",
+    }
 
 
 @mock.patch("sdw_updater.Updater._write_updates_status_flag_to_disk")
@@ -356,10 +363,7 @@ def test_read_dom0_update_flag_from_disk_fails(mocked_info, mocked_error, tmp_pa
         ("4.4", "4.3", True),  # System in middle of 4.3 -> 4.4 upgrade
     ],
 )
-@pytest.mark.skipif(
-    importlib.util.find_spec("qubesadmin") is None,
-    reason="qubesadmin module not available (only runs in dom0)",
-)
+@skipif_no_qubesadmin
 def test_is_qubes_mid_upgrade(qubes_ver, agent_ver, is_mid_upgrade, mocker, mocked_qubes_app):
     # Overrding "dnf.rpm.detect_releasever" to simulate being on particular Qubes version
     mocker.patch("dnf.rpm").detect_releasever.return_value = qubes_ver
--- a/sdw_updater/Updater.py
+++ b/sdw_updater/Updater.py
@@ -39,26 +39,50 @@ sdlog = Util.get_logger(module=__name__)
 detail_log = Util.get_logger(prefix=DETAIL_LOGGER_PREFIX, module=__name__)
 
 
-def _get_current_vms() -> dict[str, str]:
-    debian_version = "13"
-
-    # The are the TemplateVMs that require full patch level at boot in order to start the inbox,
-    # as well as their associated TemplateVMs.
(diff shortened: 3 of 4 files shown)


sdw_updater/Updater.py
```diff
@@ -39,26 +42,48 @@
...
+    # Lazy import: qubesadmin is a dom0-only system package, not available in
+    # the CI venv.
+    import qubesadmin
+
+    app = qubesadmin.Qubes()
+
+    templates = set()
+    for vm in app.domains:
+        if vm.klass == "AppVM" and "sd-workstation" in vm.tags:
+            templates.add(_backing_template(vm))
```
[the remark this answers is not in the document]
legoktm: This is an intentional limitation to skip the base template.

launcher/tests/test_updater.py
```diff
@@ -356,10 +366,7 @@ def test_read_dom0_update_flag_from_disk_fails(mocked_info, mocked_error, tmp_pa
         ("4.4", "4.3", True),  # System in middle of 4.3 -> 4.4 upgrade
     ],
 )
-@pytest.mark.skipif(
-    importlib.util.find_spec("qubesadmin") is None,
-    reason="qubesadmin module not available (only runs in dom0)",
-)
+@skipif_no_qubesadmin
 def test_is_qubes_mid_upgrade(qubes_ver, agent_ver, is_mid_upgrade, mocker, mocked_qubes_app):
     # Overrding "dnf.rpm.detect_releasever" to simulate being on particular Qubes version
```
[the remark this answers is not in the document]
legoktm: pre-existing issue

sdw_updater/Updater.py
```diff
@@ -39,26 +42,49 @@
...
+def _backing_template(vm: qubesadmin.vm.QubesVM) -> str:
+    """
+    Return the name of the TemplateVM backing a qube.
+
+    A DispVM's template is itself an AppVM (the DispVM template), so follow
+    `.template` until we reach the backing TemplateVM.
+    """
+    while vm.klass != "TemplateVM":
+        vm = vm.template
+    return vm.name
```
deeplow: Minor suggestion suggestion: we could go the other way around, which has a built-in function:

> |  derived_vms
> |      Return list of all domains based on the current TemplateVM
> |      at any level of inheritance.
 

basically something like this.

```python3
templates = [q for q in app.domains if "sd-workstation" in q.tags and t.klass == "TemplateVM"]
in_use_templates = [t for t in templates if len(t.derived_vms) > 0]
```

This way we don't need to do the traversal ourselves. Not that it's too complicated.

sdw_updater/Updater.py
```diff
@@ -39,26 +42,49 @@
...
+    """
+    # Lazy import: qubesadmin is a dom0-only system package, not available in
+    # the CI venv.
+    import qubesadmin
+
+    app = qubesadmin.Qubes()
+
+    templates = set()
+    for vm in app.domains:
+        if vm.klass == "AppVM" and "sd-workstation" in vm.tags:
```
deeplow: Another thing to check for here is if it doesn't have `prohibit-start`. Not sure what the interaction will be. I don't expect us to hit this situation, but might as well avoid it.

sdw_updater/Updater.py
```diff
@@ -39,26 +42,49 @@
...
 
 
 def _get_current_templates() -> set[str]:
-    return set([val for key, val in _get_current_vms().items() if key != "dom0"])
+    """
+    Return the set of TemplateVMs that the updater should update.
+
+    Iterate over:
+    * AppVMs tagged sd-workstation
+    * Hardcoded SYSTEM_VMs that are untagged
```
deeplow: My suggestion here would be separate this out into its own function and I think it would be reasonable to add a tag to these VMs so we don't have to hardcode them. Maybe a tag like `managed-by-securedrop` (note that this is [a special kind of tag](https://doc.qubes-os.org/en/latest/developer/services/admin-api.html#tags)).

This way we start untangling what is `sd-workstation` and what is simply `managed-by-securedrop`.

launcher/tests/conftest.py
```diff
@@ -56,35 +56,79 @@ def tmpdir():
...
+# SecureDrop-managed AppVMs (tagged) mapped to their template.
+MOCK_SDW_APPVMS = {
+    "sd-app": "sd-inbox-debian-XX",
+    "sd-log": "sd-inbox-debian-XX",
+    "sd-proxy": "sd-inbox-debian-XX",
+    "sd-gpg": "sd-inbox-debian-XX",
+    "sd-viewer": "sd-viewer-debian-XX",
+    "sd-devices": "sd-viewer-debian-XX",
+    "sd-printers": "sd-viewer-debian-XX",
+}
```
deeplow, line 78: Great idea, moving these hardcoded values to tests!

deeplow (commented):
Early feedback, as requested.

Strategy-wise I think this a fine approach but I have a few suggestions.

sdw_updater/Updater.py
legoktm: oooh. Yes. TIL

sdw_updater/Updater.py
deeplow: Looking at this again, I think this merits its own discussion. Fine to keep the hardcoding.

https://github.com/freedomofpress/securedrop-workstation/issues/1817

legoktm:
@deeplow I believe I addressed everything you suggested. (GitHub Actions is borked right now unfortunately)

sdw_updater/Updater.py
```diff
@@ -39,26 +39,43 @@
...
+    # the CI venv.
+    import qubesadmin
+
+    app = qubesadmin.Qubes()
+
+    templates = set()
+    for vm in app.domains:
+        if vm.klass != "TemplateVM":
+            continue
+        if not vm.derived_vms:
```
deeplow, line 69: I was the one suggesting `derived_vms`, but the function is perhaps overkill due to its [recursiveness](https://github.com/QubesOS/qubes-core-admin-client/blob/bb41ca5e0f4b7d58c91caabc84020168231171c1/qubesadmin/vm/__init__.py#L441-L449)  and on some systems can some extra seconds than necessary. There's the [`qube.appvms`](https://doc.qubes-os.org/projects/core-admin-client/en/latest/qubesadmin.vm.html#qubesadmin.vm.QubesVM.appvms), which is basically the same thing, but checks just a single level.

sdw_updater/Updater.py
```diff
@@ -39,26 +39,43 @@
...
 
 
 def _get_current_templates() -> set[str]:
-    return set([val for key, val in _get_current_vms().items() if key != "dom0"])
+    """
+    Return the set of TemplateVMs that the updater should update.
+
+    Find:
+    * TemplateVMs tagged sd-workstation with derived_vms
+    * TemplateVMs that back hardcoded SYSTEM_VMS
```
deeplow, line 52: If feels like these two could be separate, at least separate loops (as you [had before](https://github.com/freedomofpress/securedrop-workstation/commit/1b570a1498e4f568c838d31e9a8ebcf4eec6dd76#diff-c49796294389d55ca4beb4396a52db9fbaee39b87fbd82f3d3ad2b503e8da417R78-R85)). If we do so:
- checking for `sd-workstation` tagged qubes can exit much earlier
- checking for System VMs can be as simple looping through `getattr(vm, "template")` until there's nothing else.

deeplow (changes requested):
I'm a little concerned about the number of cycles introduced here and I have suggested some optimizations.

### Thoughts obtaining SystemVM (without hardcoding) 
The System VMs are the "complicated" case for which hard-coding is the easiest. If we wanted to avoid doing so (and that could be a _separate issue on its own_). For now I find it preferable to have the [solution you had before](https://github.com/freedomofpress/securedrop-workstation/commit/1b570a1498e4f568c838d31e9a8ebcf4eec6dd76#diff-c49796294389d55ca4beb4396a52db9fbaee39b87fbd82f3d3ad2b503e8da417R83-R85).

But for the sake of completeness here are some thoughts on how such a generic function could be achieved. This could be useful in other parts of the code that deal with non SDW qubes — for example doing template upgrades for Fedora qubes (ideally though `qvm-template-upgrade` once it's published).

Conceptually speaking the updater wants to find the **transitive closure** of `sd-workstation` — i.e. all the "reachable" qubes from the ones tagged with `sd-workstation` (implicitly or explicitly). This should even be an upstreamable function. Such a function would need to need to traverse the graph though all the qube properties [that refer to other ones](https://github.com/QubesOS/qubes-core-admin-client/blob/3d421e7/qubesadmin/utils/__init__.py#L156-L163) plus RPC policies.

This last one could re-use `qrexec-policy-graph` (or some of its functions) to find the explicitly allowed policies. This is what we'd use to reach `sys-usb` from `sd-devices`.

### Thoughts on Complexity

I understand the motivation iterating through all qubes in the system, but combined with calculating `derived_vms` it becomes unnecessarily slow, I find.

For context I tested this on a system with a significant number of qubes and it took almost 8 seconds to calculate (`_get_current_templates`).
```
[user@dom0 sdw_updater]$ python3 test.py 
{'fedora-43-xfce', 'sd-viewer-debian-13', 'sd-inbox-debian-13'}
Took 7.739326399983838 seconds
```

Here are some other metrics I took, for comparison:
- a single iteration through all qubes in the system took `0.06` seconds.
- calculating `derived_vms` for all templates took about `3.94` seconds.

And this lines up with the implementation complexity as it calculates `derived_vms` twice.

sdw_updater/Updater.py
legoktm: Claude flags that worst-case derived_vms is O(N^2) but there's an O(N) implementation; I'll see about submitting that upstream.

legoktm:
If the issue is just that derived_vms is slow, I'd rather fix that, we can just have our own function for now and swap back to the upstream implementation once its fixed.

legoktm:
TIL `app.cache_enabled`. That fixes all the performance issues, and is scoped to just our Qubes() instance.

sdw_updater/Updater.py
deeplow: I guess my broader point in the review comment was that the algorithms are conceptually very different. In the system VM case it's just traversing back (as you had done first). The other one is a graph traversal problem. Combining the two feels like mixing together different concepts and makes it a bit harder for me to reason around what it does.

<details><summary>I was drafting  a suggestion that addresses the efficiency and the separation, but my main point is the that they feel conceptually separate (regardless of efficiency)</summary>

```python
# Obtain tagged templates
templates.update(
    vm.name
    for vm in app.domains
    if vm.klass == "TemplateVM"
    and "prohibit-start" not in vm.features  # not disabled
    and "sd-workstation" in vm.tags
    and any(vm.appvms)   # something uses this template
)

# Obtain system templates
for name in SYSTEM_VMS:
      vm = app.domains[name]
      while vm.klass != "TemplateVM":
          vm = vm.template
       templates.add(vm.name)
```

</details>

sdw_updater/Updater.py
legoktm: Done!

deeplow (approved):
Visual review-wise these last changes LGTM. Will wait for OpenQA results prior to merging.

The caching really made a big difference. Thanks for finding that out!

sdw_updater/Updater.py
legoktm: Filed upstream: https://github.com/QubesOS/qubes-issues/issues/11083.

PR #1795, opened 2026-07-22
fix(policy): prevent qubes.USBAttach wildcard from shadowing sd-works…

mausamrijall:
…tation deny rules

The wildcard rule 'qubes.USBAttach * @anyvm @anyvm ask' in 31-securedrop-workstation.policy was evaluated before the deny rules in 32-securedrop-workstation.policy (Qubes first-match-wins evaluation order). This made the deny rules in file 32 unreachable dead code, meaning a USB attach request targeting any @tag:sd-workstation VM would result in an interactive 'ask' prompt instead of being silently denied.

Fix: Add explicit deny rules for @tag:sd-workstation targets directly in 31-securedrop-workstation.policy, immediately before the wildcard ask rule, ensuring they are evaluated first within the same file.

  qubes.USBAttach * @anyvm @tag:sd-workstation deny
  qubes.USBAttach * @tag:sd-workstation @anyvm deny

The sd-export-target allow rule (sys-usb -> @tag:sd-export-target) is intentionally preserved above the new deny rules since that allow is the correct and expected policy for USB export operations.

Add regression test test_usbattach_policy_denies_sd_workstation() to tests/test_rpc_policy.py to verify that qubes.USBAttach from external VMs (sys-net, sys-firewall, sys-usb) to any non-export sd-workstation VM is denied, and that sd-workstation VMs cannot initiate USB attaches to those external VMs either.

Fixes #

## Checklist

This change accounts for:

### DIFF
--- a/files/31-securedrop-workstation.policy
+++ b/files/31-securedrop-workstation.policy
@@ -32,6 +32,11 @@ qubes.GpgImportKey      *           @tag:sd-client sd-gpg allow
 qubes.Gpg2              *           @tag:sd-client sd-gpg allow target=sd-gpg
 
 qubes.USBAttach         *           sys-usb @tag:sd-export-target allow  user=root
+# Deny USB attachment to/from all sd-workstation-tagged VMs explicitly before
+# the wildcard ask rule below, preventing the deny rules in
+# 32-securedrop-workstation.policy from being shadowed (first-match-wins).
+qubes.USBAttach         *           @anyvm @tag:sd-workstation deny
+qubes.USBAttach         *           @tag:sd-workstation @anyvm deny
 qubes.USBAttach         *           @anyvm @anyvm ask
 
 qubes.USB               *           @tag:sd-export-target sys-usb allow
--- a/tests/test_rpc_policy.py
+++ b/tests/test_rpc_policy.py
@@ -101,3 +101,46 @@ def test_policy_from_sdgpg_to_dom0_allowed(sdw_tagged_vms: list[QubesVM], qubes_
             assert allowed
         else:
             assert not allowed
+
+
+@pytest.mark.provisioning
+def test_usbattach_policy_denies_sd_workstation(sdw_tagged_vms: list[QubesVM]) -> None:
+    """
+    Regression test: qubes.USBAttach to or from any @tag:sd-workstation VM must
+    be denied and must NOT resolve to an 'ask' prompt.
+
+    Previously, the wildcard 'qubes.USBAttach * @anyvm @anyvm ask' rule in
+    31-securedrop-workstation.policy was evaluated before the deny rules in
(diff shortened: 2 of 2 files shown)


files/31-securedrop-workstation.policy
```diff
@@ -32,6 +32,11 @@ qubes.GpgImportKey      *           @tag:sd-client sd-gpg allow
 qubes.Gpg2              *           @tag:sd-client sd-gpg allow target=sd-gpg
 
 qubes.USBAttach         *           sys-usb @tag:sd-export-target allow  user=root
+# Deny USB attachment to/from all sd-workstation-tagged VMs explicitly before
+# the wildcard ask rule below, preventing the deny rules in
+# 32-securedrop-workstation.policy from being shadowed (first-match-wins).
+qubes.USBAttach         *           @anyvm @tag:sd-workstation deny
+qubes.USBAttach         *           @tag:sd-workstation @anyvm deny
 qubes.USBAttach         *           @anyvm @anyvm ask
```
deeplow, line 40: Generally our `31-` file is for what we allow and `32-` for what we deny. Perhaps the right move here is to just simply remove the `@anyvm @anyvm ask` line and this should also solve the problem. I'll add some notes about this in a comment above.

deeplow (changes requested):
Thanks for finding this. Generally speaking this does not have security impact since users will see a dialog, and would have to pick a VM. But it's less than ideal.

I did  some digging into why we had the `@anyvm @anyvm ask` policy and I coudn't find a good reason. It could have been a guard to prevent overriding by another policy, but even then it doesn't make sense. So I think we can remove it without a problem. The default is to deny anyways and we're already guarding against impact with the other two denial policies.

For historical context, that was introduced in [this commit](https://github.com/freedomofpress/securedrop-workstation/commit/3703bd827428ce4b519e8be485e841da2864a08b#diff-c3b8900707dbf877b778a2d4222b636663d737f039290adc74585335acba24b8R89-R106) with no clear explanation/. But that commit also provides some clarity: the bug you identified of the `ask` policy being before was actually introduced when the policy format was converted (see [qubes-issue on it](https://github.com/QubesOS/qubes-issues/issues/4370) and the [implementation in the workstation](https://github.com/freedomofpress/securedrop-workstation/commit/b99e43526c53f74230023f05a38ed6d8e6fdf339) that introduced the bug).

deeplow:
@mausamrijall would you have availability to finish this until next week? Otherwise we can wrap it up, so it gets included in an upcoming release.

deeplow:
Closing this in favor of https://github.com/freedomofpress/securedrop-workstation/pull/1825, so I can iterate on the branch without overwriting your main. Thanks again for the contribution. Please do feel free to chime in, if you have other comments :slightly_smiling_face:

PR #1802, opened 2026-07-22
Have `make test-launcher` fully pass locally

legoktm:
* Add missing pytest-mock package in poetry
* Skip tests that require qubesadmin by checking for that instead of a CI=true environment variable

## Test plan

* [ ] Running `poetry sync && make test-launcher` passes
* [ ] OpenQA output shows that these tests are not skipped there
## Checklist

This change accounts for:

### DIFF
--- a/launcher/tests/test_updater.py
+++ b/launcher/tests/test_updater.py
@@ -1,3 +1,4 @@
+import importlib.util
 import json
 import os
 import subprocess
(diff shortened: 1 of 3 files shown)


legoktm:
OpenQA failure is for test_local_key_in_remote_keyring, which I haven't seen flake/fail before.

deeplow:
And it passed on [yesterday's nightly](https://openqa.qubes-os.org/tests/189250#step/test_vm_sd_gpg/3). Today's nightly failed due to something unrelated to the workstation (or so it seems).

deeplow:
> OpenQA failure is for test_local_key_in_remote_keyring, which I haven't seen flake/fail before.

Could be the key changing due to https://github.com/freedomofpress/securedrop/pull/7882/changes

OpenQA actually pulls in this key from the securedrop repo.

legoktm:
No, the key fingerprint didn't change, we just added an extra binding signature (plus that was a few weeks ago...). It should import just fine into GPG. 

I do think it's unrelated to this PR and we can just keep an eye on it failing again.

deeplow:
You're right.

deeplow (approved):
Good catch and a better approach.

Strictly speaking, a the pytest mock is still not needed because dom0 doesn't use poetry. But it doesn't hurt to have it there. As a side-benefit, this can now be run in any regular qubes VM, so long as qubesadmin is installed.

deeplow (approved):

PR #1804, opened 2026-07-22
Deploy locally built RPMs as a server (for updater testing)

deeplow:
Adds an RPM server from the built RPMs. This enables the testing of the
update process. This is only intended to be used for development and in
terms of threat model, it should not change regarding the previous
approach.

It also generates a GPG key to sign the artifacts, as qubes-dom0-update
requires it (see qubes.ReceiveUpdates RPC service). This needs to happen
even in cases where signature checking is disabled.

## Test plan

- [ ] have some old version installed (e.g. 1.8.0)
- [ ] in dom0 run `./scripts/local-rpm-server.sh`
- [ ] run `sdw-updater`
- [ ] Ensure the installed version is the one from your local build

## Checklist

This change accounts for:

### DIFF
--- a/scripts/clone-to-dom0
+++ b/scripts/clone-to-dom0
@@ -6,6 +6,8 @@ set -e
 set -u
 set -o pipefail
 
+source "$(dirname "$0")/common.sh"
+
 # Ensure we're running in dom0, otherwise clone action could destroy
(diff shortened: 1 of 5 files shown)


legoktm (approved):
Fantastic, ty!!

PR #1809, opened 2026-07-24
Don't have any wildcard versions in pyproject.toml

legoktm:
dependabot gets confused and thinks because any version is acceptable, it can do dependency resolution by e.g. downgrading semgrep, which we don't want.

## Test plan

just visual review should be fine


zenmonkeykstop (approved):
LGTM

PR #1825, opened 2026-08-06
fix(policy): prevent qubes.USBAttach ask from overwriting other policies

deeplow:
Replaces #1795. Credit to @mausamrijall for the original PR. This is was opened as a separate PR so we wouldn't overwrite the contributor's main branch.

Quoting from original PR:
> The wildcard rule 'qubes.USBAttach * https://github.com/AnyVM https://github.com/AnyVM ask' in 31-securedrop-workstation.policy was evaluated before the deny rules in 32-securedrop-workstation.policy (Qubes first-match-wins evaluation order). This made the deny rules in file 32 unreachable dead code, meaning a USB attach request targeting any @tag:sd-workstation VM would result in an interactive 'ask' prompt instead of being silently denied.
> 
> Fix: Add explicit deny rules for @tag:sd-workstation targets directly in 31-securedrop-workstation.policy, immediately before the wildcard ask rule, ensuring they are evaluated first within the same file.
> 
> qubes.USBAttach * https://github.com/AnyVM @tag:sd-workstation deny
> qubes.USBAttach * @tag:sd-workstation https://github.com/AnyVM deny
> 
> The sd-export-target allow rule (sys-usb -> @tag:sd-export-target) is intentionally preserved above the new deny rules since that allow is the correct and expected policy for USB export operations.
> 
> Add regression test test_usbattach_policy_denies_sd_workstation() to tests/test_rpc_policy.py to verify that qubes.USBAttach from external VMs (sys-net, sys-firewall, sys-usb) to any non-export sd-workstation VM is denied, and that sd-workstation VMs cannot initiate USB attaches to those external VMs either.

Extra changes (on top of original):
- Removes overly broad @anyvm<->@anyvm ask policy on `qubes.USBAttach (see commit message)
- Modify regression test to be test usb policies instead of just as a regression test

## Test plan

- [ ] visual review
- [ ] tests pass

## Checklist

This change accounts for:

### DIFF
--- a/files/31-securedrop-workstation.policy
+++ b/files/31-securedrop-workstation.policy
@@ -32,6 +32,11 @@ qubes.GpgImportKey      *           @tag:sd-client sd-gpg allow
 qubes.Gpg2              *           @tag:sd-client sd-gpg allow target=sd-gpg
 
 qubes.USBAttach         *           sys-usb @tag:sd-export-target allow  user=root
+# Deny USB attachment to/from all sd-workstation-tagged VMs explicitly before
+# the wildcard ask rule below, preventing the deny rules in
+# 32-securedrop-workstation.policy from being shadowed (first-match-wins).
+qubes.USBAttach         *           @anyvm @tag:sd-workstation deny
+qubes.USBAttach         *           @tag:sd-workstation @anyvm deny
 qubes.USBAttach         *           @anyvm @anyvm ask
 
 qubes.USB               *           @tag:sd-export-target sys-usb allow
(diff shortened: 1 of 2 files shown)


deeplow:
@legoktm logs [failed to upload](https://openqa.qubes-os.org/tests/192025), but we can see it's a qrexec timeout issue:

Screenshot 2026-08-11 at 17-16-20 Qubes OS openQA Video

legoktm (approved):
LGTM; the original commit isn't signed, I can sign it for them.

PR #1826, opened 2026-08-06
Use shellcheck-py package

legoktm:
As we do in the server; this greatly simplifies how we run shellcheck and means we're using the same version everywhere and are no longer relying on a third-party docker image.

----

I noticed this when working on the rpmlint PR and was confused why we were pulling in the docker image.

## Test plan

* [x] CI is happy
## Checklist

This change accounts for:


cfm (approved):
Delightful!

Issue #1822, opened 2026-08-04
Updater ADR


jskinne3:
We should document the SDW updater in an ADR.

Things to cover include:
* Reasons for proposal in 2019: https://github.com/freedomofpress/securedrop-workstation/issues/341
* How the updater handles failed updates and restarting updates
* How we log updates
* Choices tagged with "[updater](https://github.com/freedomofpress/securedrop-workstation/issues?q=is%3Aissue%20updater%20label%3Aupdater)"

Give the context and consequences of these choices, and briefly describe alternatives we considered.

Place the result in [architecture-decisions](https://github.com/freedomofpress/securedrop-workstation/tree/main/architecture-decisions).

legoktm:
In 2023 we spent some time extracting the updater out into a separate package and repository: https://github.com/freedomofpress/securedrop-workstation/issues/808 / https://github.com/freedomofpress/securedrop-updater/. That work was unfinished for reasons. In https://github.com/freedomofpress/securedrop-workstation/issues/966 during the 4.2 push, some of that refactoring was re-absorbed, but some components, like versioned migrations, were abandoned.

I thought there was some sort of proposal doc for the split securedrop-updater but I can't find one, so I suspect the GH issue is the authoritative source.

In 2024, a major milestone was refactoring the template updates to use Qubes' updater under the hood: https://github.com/freedomofpress/securedrop-workstation/issues/899. That issue has a lot of rich comments regarding the design of the updater, e.g. https://github.com/freedomofpress/securedrop-workstation/issues/899#issuecomment-1583560608.

legoktm:
I've turned Erik's comment that I linked above into an ADR that focuses on the purpose of the updater without getting too much into the implementation. I think it covers the first two bullets?

I'm not sure how much there is to say about logging, we try to log everything. 

Looking through the updater-tagged issues (+ my own thoughts), I think it's also worth documenting:

1. Migrations: https://github.com/freedomofpress/securedrop-workstation/issues/567, https://github.com/freedomofpress/securedrop-workstation/issues/966, https://github.com/freedomofpress/securedrop-workstation/issues/673
2. 1.8.0 workarounds + planned 1.8.1 improvements.

jskinne3:
Merged https://github.com/freedomofpress/securedrop-workstation/pull/1832
Keeping the issue open because I think Kunal is planning at least one more ADR

legoktm:
Last ADR is up for review!

PR #1832, opened 2026-08-11
Add retroactive ADR for the updater's purpose

legoktm:
Based off of <https://github.com/freedomofpress/securedrop-workstation/issues/899#issuecomment-1583560608>.

Refs #1822.

### DIFF
--- a/architecture-decisions/002-updater-purpose.md
+++ b/architecture-decisions/002-updater-purpose.md
@@ -1,0 +1,50 @@
+# Updater purpose
+
+_This is a retroactive ADR to document past decisions that determined the purpose of the SecureDrop Workstation Updater._
+
+## Summary
+
+The updater, also known as the launcher, enforces a policy of requiring all updates to be applied before the Inbox can be used. The goal is to ensure users have all available updates, whether security-critical or not, to ensure an attacker cannot exploit a publicly known vulnerability against a journalist.
+
+## Context
+
+The default Qubes updater does not provide any sort of enforcement or requirement to install updates before VMs can be launched.
+
+One of the promises of a Qubes-based workstation as opposed to Tails is that because it does have internet access, we can install more software on it instead of being limited to merely what Tails ships in their offline image. However, that also comes with the need and responsibility to keep all of that software regularly up to date.
+
+## Decision
+
+### Policy
+
+Upon login, the updater checks to see if updates have been applied within the past 8 hours. If not, it
+requires the user run updates first.
+
+The updater downloads and installs updates for dom0, the Debian-based SDW TemplateVMs and the current Fedora TemplateVM for sys-* qubes. It checks for all updates unconditionally, it does not rely on Qubes' knowledge of whether there are pending updates.
+
+If dom0 updates are installed, it requires the user to reboot so they take effect. If there are no dom0 updates, all the affected VMs are restarted so they pick up the updates to the templates.
+
+If the updater fails for whatever reason, the Inbox cannot be opened. We want to fail closed, erring on the side that if we cannot guarantee the system is fully up to date, we should prevent the user from using the system. This is only enforced through the updater, i.e. nothing stops a journalist from manually launching `sd-app`.
+
+### MultiVM upgrades
+
+Changes to VMs from dom0 are often larger than what can be done in the RPM upgrade itself. As a result, the updater is also responsible for calling the relevant tool to ensure those changes take effect.
+
+### State enforcement
+
+A side-effect of how the updater is implemented is that after dom0 updates are applied, the updater runs the dom0 salt states, which serves as some level of state enforcement.
+
+This is not necessarily intentional or desired, but it does provide some level of self-healing.
+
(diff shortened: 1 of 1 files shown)


architecture-decisions/002-updater-purpose.md
```diff
@@ -0,0 +1,24 @@
...
+
+The updater downloads and installs updates for dom0, the Debian-based SDW TemplateVMs and the current Fedora TemplateVM for sys-* qubes. It checks for all updates unconditionally, it does not rely on Qubes' knowledge of whether there are pending updates.
+
+If dom0 updates are installed, it requires the user to reboot so they take effect. If there are no dom0 updates, all the affected VMs are restarted so they pick up the updates to the templates.
+
+If the updater fails for whatever reason, the Inbox cannot be opened. (Through the updater at least, nothing stops a journalist from e.g. manually launching `sd-app` and opening the Inbox via a terminal.)
+
+## State enforcement
+
+A side-effect of how the updater is implemented is that after dom0 updates are applied, the updater unconditionally runs the dom0 salt states, which serves as some level of state enforcement.
```
legoktm: This is no longer correct as of 1.8.0, but since this is supposed to be historical/retroactive, I've described it as it used to be. I'm working on a separate doc for the 1.8.0 changes.

architecture-decisions/002-updater-purpose.md
```diff
@@ -0,0 +1,24 @@
...
+
+The updater downloads and installs updates for dom0, the Debian-based SDW TemplateVMs and the current Fedora TemplateVM for sys-* qubes. It checks for all updates unconditionally, it does not rely on Qubes' knowledge of whether there are pending updates.
+
+If dom0 updates are installed, it requires the user to reboot so they take effect. If there are no dom0 updates, all the affected VMs are restarted so they pick up the updates to the templates.
+
+If the updater fails for whatever reason, the Inbox cannot be opened. We want to fail closed, erring on the side that if we cannot guarantee the system is fully up to date, we should prevent the user from using the system. This is only enforced through the updater, i.e. nothing stops a journalist from manually launching `sd-app`.
+
+## State enforcement
+
+A side-effect of how the updater is implemented is that after dom0 updates are applied, the updater unconditionally runs the dom0 salt states, which serves as some level of state enforcement.
```
deeplow: We're missing here the role of "migrations". Was this before those existed? I see these are "MultiVM upgrades" (better word needed). Please let me know in case you want me to write something on it.

legoktm: See https://github.com/freedomofpress/securedrop-workstation/issues/1822#issuecomment-5255093861; I want to do a separate ADR about migrations. 

IMO migrations are an implementation detail rather than a specific policy choice that the updater must have, which is why I left it out of this.

deeplow (commented):
I guess the purpose is to keep this as an artifact of the past, but it's hard to fight the urge to update it.

architecture-decisions/002-updater-purpose.md
jskinne3: I see what you're trying to do in making this historical, but I don't like the idea of newly introducing now-incorrect statements.


```suggestion
A side-effect of how the updater was implemented is that after dom0 updates were applied, the updater unconditionally ran the dom0 salt states, which served as some level of state enforcement. This is no longer true as of 1.8.0.
```

jskinne3 (changes requested):
I'd like to have the headings `Summary`, `Context`, `Decision`, `Consequences` and `Alternatives considered` as a standard template for ADRs including this one.

architecture-decisions/002-updater-purpose.md
deeplow: By migrations I was referring to the concept of "MultiVM upgrades". The specific implementation through what we called "migrations" is indeed an implementation detail. With that said, the fact that this is called by the updater is something I believe should be mentioned (even if we don't define its implementation).

Something like:

> Upon detection that the system's architecture has changed, the updater should call the appropriate tool to ensure these changes take effect.

Would something like this work for you? My point being: this is a bigger topic and we don't have to get into implementation details, but at the end of the day is is an integral part of the updater and thus should be mentioned.

legoktm: Oh yes, that makes total sense. Let me work that in.

architecture-decisions/002-updater-purpose.md
legoktm: I removed "unconditionally", which I think is sufficient to make it correct today, as the full migration path runs sdw-admin --apply which also runs the dom0 states.

legoktm:
Added the standard headings, filling in Context, Consequences, and Alternatives Considered, plus the MultiVM upgrade stuff.

deeplow (commented):
Approving as far as my feedback is concerned. @jskinne3 feel free to approve if your feedback has been addressed as well.

jskinne3 (approved):
Nice! Thank you for your work Kunal!

Issue #1750, opened 2026-07-06
Add unit tests to `sdw-admin` and other scripts in `files/`


deeplow:
The [scripts under `files/`](https://github.com/freedomofpress/securedrop-workstation/tree/1169010/files) include `sdw-admin.py` and other scripts that are used in production. Most of them are already tested through integration testing, but as they grow more complex it's good to have unit tests.

Completion of this task requires:
- [ ] proper python module packaging for `files/*.py` — otherwise we need workarounds to import them in tests
- [ ] unit tests

deeplow:
For context, this is a follow up to a past conversation:

>> **NOTE:** It'd like to make some tests for this (separately from the release — can go on another PR), but the fact that sdw-admin.py is a standalone python file rather than an proper Debian package has made that a bit awkward.
>>
>> Shall we turn it into a python package (or include it in sdw_updater), for example?
>> — _Originally posted by @deeplow in [#1744](https://github.com/freedomofpress/securedrop-workstation/pull/1744/changes#r3506668897)_
>
> Yes, I think we should. IIRC we were kinda moving in this direction anyways with https://github.com/freedomofpress/securedrop-workstation/issues/1680 (one other script that can no longer be standalone).
> 
> I would suggest for tests though that you just do an importlib hack/workaround for now, and once we have tests then we can rework sdw-admin :) 
>
> — _Originally posted by @legoktm in [#1744](https://github.com/freedomofpress/securedrop-workstation/pull/1744/changes#r3515679196)_

PR #1839, opened 2026-08-27
Test fixtures improvement

deeplow:
Split out of from https://github.com/freedomofpress/securedrop-workstation/pull/1754/ to simplify its review.

1. Mark PROJECT_ROOT fixture as session-scoped
2. Adds generic fixture for import non-compliant py files
    
    We already had a way to import non-compliant python files: RPC services.
    These problematic standalone files are not part of any python module and
    may lack the extension and have invalid file names.
    
    This moves that fixture into conftest as a fixture factory so that
    sdw-admin and other non-compliant modules can be imported until we place
    them in proper python modules (see #1750).

## Checklist

This change accounts for:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -26,7 +26,7 @@ from tests.base import (
 )
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
 def proj_root() -> os.PathLike[Any]:
     return Path(__file__).parent.parent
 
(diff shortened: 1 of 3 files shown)


legoktm (approved):
Thanks, already reviewed in the previous PR, LGTM. Will just wait for OpenQA to sign off as well.

PR #1844, opened 2026-09-01
`local-yum-server`: warn when vm dependencies are missing

deeplow:
Script would fail mid-way and make it not immediately clear the issue was that createrepo-c was missing on the system.

Reported during 1.9.0 QA. 

## Test plan

In dev vm remove some of the dependencies and ensure they are reported as missing.

## Checklist

This change accounts for:

### DIFF
--- a/scripts/local-rpm-server.sh
+++ b/scripts/local-rpm-server.sh
@@ -163,7 +163,7 @@ vm_main() {
     # Entry point in case this runs in a VM
 
     export GNUPGHOME=$SD_DEV_GNUPGHOME
-    check_pkgs createrepo_c gpg qubes-core-agent-dom0-updates
+    check_pkgs createrepo-c gpg qubes-core-agent-dom0-updates
 
     if [[ "$1" == "generate-key" ]]; then
         vm_gen_signing_key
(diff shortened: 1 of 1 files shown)


nathandyer (approved):
Code changes LGTM, and I was able to verify the behavior locally.

scripts/local-rpm-server.sh
```diff
@@ -152,6 +163,7 @@ vm_main() {
     # Entry point in case this runs in a VM
 
     export GNUPGHOME=$SD_DEV_GNUPGHOME
+    check_pkgs createrepo_c gpg qubes-core-agent-dom0-updates
```
nathandyer:
```suggestion
    check_pkgs createrepo-c gpg qubes-core-agent-dom0-updates
```

I realized the check I did was actually potentially a false-positive; I still get an error after reinstalling createrepo-c

Should the package name here have a dash, not an underscore?

nathandyer (changes requested):

nathandyer (approved):
I have confirmed that this works properly now with the corrected package name for `createrepo-c`.

PR #1845, opened 2026-09-01
Disable parallel tests in CI

### LINKED ISSUE #1539: test-created disposable causing side effects other tests when ran in parallel

deeplow:
Fixes #1539

## Checklist

This change accounts for:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,9 +158,24 @@ install-dom0-test-prereqs: assert-dom0 ## Installs pytest dependencies in dom0
 	rpm -q $(DOM0_TEST_PREREQS) || \
 		sudo qubes-dom0-update -y $(DOM0_TEST_PREREQS)
 
+
+# N.B. make sure to use "-vv" for max-length diffs on test failures
+PYTEST_CMD := pytest -vv --durations=5 tests launcher/tests
+
 test: test-prereqs ## Runs all application tests (no integration tests yet)
-	# N.B. make sure to use "-vv" for max-length diffs on test failures
-	pytest -vv tests launcher/tests -n auto --dist=loadfile --durations=5
+ifeq ($(CI),true)
+# In CI we want to:
+# 1. Run tests in series to give tests the best chance at not interfering with eachother
+# 2. Produce a JunitXML report for OpenQA to show the results
+	$(PYTEST_CMD) --cov-report term-missing --cov=sdw_notify --cov=sdw_updater --cov=sdw_util \
+	              --junitxml=test-data.xml
+else
+# Locally we want to prioritise parallelization for faster iterations
+# at the cost of needing to run some tests in sequence through a separate command
+	$(PYTEST_CMD) -m "run_alone"
+	$(PYTEST_CMD) -m "not run_alone"  -n auto --dist=loadfile --durations=5
+endif
+
 
 test-base: test-prereqs ## Runs tests for VMs layout
 	pytest -v tests/test_vms_exist.py
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -143,8 +143,6 @@ disallow_untyped_defs = false
 disallow_incomplete_defs = false
 
 [tool.pytest.ini_options]
-addopts = "--cov-report term-missing --cov=sdw_notify --cov=sdw_updater --cov=sdw_util --junitxml=test-data.xml"
-
 # Registering custom pytest markers for tests. Allows selecting or skipping by marked groups.
 markers = [
   "mime: checks for filetype/application associations; requires shell access",
@@ -152,4 +150,5 @@ markers = [
   "packages: checks that all packages are up to date; quite slow, generally",
   "provisioning: checks that VMs exist and their extra-VM prefs are set correctly",
   "configuration: assumes that VMs exist inspects their intra-VM files and folders",
+  "run_alone: tests that must not run in parallel due to dom0 state mutation",
 ]
--- a/tests/base.py
+++ b/tests/base.py
@@ -94,12 +94,9 @@ def is_managed_qube(qube: QubesVM) -> bool:
 def is_workstation_qube(qube: QubesVM) -> bool:
     """
     Is the qube a managed, workstation-tagged qube?
-
-    NOTE: filter out the "sd-viewer-disposable" qube, which is an ephemeral
-    DispVM, which will exist at certain points of the test suite
     """
 
-    return is_managed_qube(qube) and SD_TAG in qube.tags and qube.name != "sd-viewer-disposable"
+    return is_managed_qube(qube) and SD_TAG in qube.tags
 
 
 class QubeWrapper:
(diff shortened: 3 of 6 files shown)


deeplow:
The only thing [that was failing](https://openqa.qubes-os.org/tests/194847#step/Test_SD_App_Common/7) was my attempt at seeing if the mock_attack test would work now, if things ran in sequence — it didn't.

I removed it and so it's ready for review.

legoktm (approved):
Nice :) one less thing for us to think about...

PR #1784, opened 2026-07-20
Do not run dom0 states independently if full migration is needed

legoktm:
This unnecessarily forces us to stick everything in the dom0 state and adds duplication in that the dom0 states end up getting run twice.

Fixes #1763.

## Test plan

* [x] Manually drop an updater flag in /tmp/sdw-migrations 
* [x] Run `sdw-updater --skip-delta 0`
* [x] Review updater log and see that it did not run dom0 states and only ran full sdw-admin --apply
## Checklist

This change accounts for:
- [ ] any necessary RPM packaging updates (e.g., added/removed files, see `rpm-build/SPECS/securedrop-workstation-dom0-config.spec`)
- [ ] any required documentation

PR #1848, opened 2026-09-04
Add ADR that discusses Debian 13 upgrade issues and fixes

### LINKED ISSUE #1822: Updater ADR

legoktm:
Fixes #1822.

### DIFF
--- a/architecture-decisions/004-trixie-updater-woes.md
+++ b/architecture-decisions/004-trixie-updater-woes.md
@@ -1,0 +1,59 @@
+# Debian 13 upgrade's updater woes
+
+## Summary
+
+During the Debian 12 (bookworm) -> Debian 13 (trixie) migration, we encountered a number of issues in the updater that necessitated various workarounds and then later, long-term solutions.
+
+## Context
+
+The Debian 13 migration had been delayed until the SecureDrop Inbox finished its rollout, so we were on a compressed timeline.
+
+We expected to be able to obsolete the `securedrop-handle-upgrade` script that was previously used, which we did; however we learned that it had bitrotted and was broken since the previous Debian 11 (bullseye) -> Debian 12 migration was done along side a fresh install and was never actually upgraded to.
+
+Part of the reason it broke was because as part of the Debian 12 migration, we stopped shipping an RPM containing a TemplateVM and instead switched to building it on the fly.
+
+### Problems
+
+The largest issue was that the upgrade had to be fully compatible with the old version of the updater (1.7.1), because that's the version of the updater that will actually execute the Debian upgrade.
+
+We encountered the following specific issues:
+
+1. The updater [runs dom0 salt states unconditionally](https://github.com/freedomofpress/securedrop-workstation/issues/1762) before the migration kicks in to run a full `sdw-admin --apply`. So either the upgrade needs to happen fully in the dom0 state or the dom0 state has to be compatible with both Debian 12 + Debian 13.
+2. Preloaded disposables interfered with switching the template for sd-viewer.
+3. The 1.7.1 updater [hardcoded the names of the Debian 12 templates](https://github.com/freedomofpress/securedrop-workstation/issues/1771) to update at the end, which meant they needed to be present on the system.
+4. `sd-log` would start up because it was receiving log entries from the newly created templates.
+
+## Decision
+
+We implemented a series of short-term fixes in the 1.8.0 release, so we could get the Debian 13 release out:
+
+1. In the RPM postinst, we [disabled the dom0 top state](https://github.com/freedomofpress/securedrop-workstation/pull/1769), so the updater's invocation of the dom0 state would be a no-op. Then because the migration flag was set, `sdw-admin --apply` would be executed, which re-enabled the top state and applied the upgrade.
+2. We temporarily [disable preloaded disposables](https://github.com/freedomofpress/securedrop-workstation/pull/1744) during the Debian upgrade.
+3. We [did not delete](https://github.com/freedomofpress/securedrop-workstation/commit/efa08077bf881d1894569b57b4934cc58e2325ea) the hardcoded Debian 12 templates, so the updater would still update them, even though they were no longer in use.
+4. We temporarily [set the `prohibit-start` feature](https://github.com/freedomofpress/securedrop-workstation/pull/1744) on all of our VMs to ensure they don't start up for any reason.
+
+One more issue that surfaced after 1.8.0 was released was that if the Debian upgrade fails mid-way (e.g. network hiccup), then it cannot be [resumed via the updater](https://github.com/freedomofpress/securedrop-workstation/issues/1796).
+
+We implemented the following long-term fixes in the following 1.9.0 release:
+
+1. [Only run the dom0 salt states](https://github.com/freedomofpress/securedrop-workstation/pull/1784) if there is no migration flag instead of unconditionally.
+2. n/a
+3. [Fetch the template names to update dynamically](https://github.com/freedomofpress/securedrop-workstation/pull/1793), so it should just work for the next Debian upgrade.
+4. n/a
+
+Additionally, we [switched](https://github.com/freedomofpress/securedrop-workstation/issues/1815) to using `%triggerun` instead of `%post` for dropping the migration flag to make it work for "skip" updates.
+
+We are in the process of adding an OpenQA scenario that performs the upgrade process, which likely would simplified identifying these bugs initially and then verifying the fixes.
+
+## Consequences
+
+The primary consequence was that the Debian 13 upgrade took longer to roll out and we did barely miss the official end-of-life deadline.
+
+If we didn't make any other changes to provision or upgrading logic, we are probably well set for the next Debian 13 -> Debian 14 upgrade in 2027.
+
+But we likely do want to overhaul the updater to make it easier for us to develop with and have [started brainstorming a redesign](https://github.com/freedomofpress/securedrop-workstation/issues/1766).
+
+## Alternatives considered
+
+* There is an open feature request for Qubes to support deferred template switches, so that you can change templates while the VM is running.
+* We tried to have a way to [inject the updater's version](https://github.com/freedomofpress/securedrop-workstation/pull/1785) through to salt and `sdw-admin --apply` to enable them to work around any future updater bugs, but the implementation ended up being quite hacky that we deferred it to the aforementioned updater redesign.


architecture-decisions/004-trixie-updater-woes.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Summary
+
+During the Debian 12 (bookworm) -> Debian 13 (trixie) migration, we encountered a number of issues in the updater that necessitated various workarounds and then later, long-term solutions.
+
+## Context
+
+The Debian 13 migration had been delayed until the SecureDrop Inbox finished its rollout, so we were on a compressed timeline.
+
+We expected to be able to obsolete the `securedrop-handle-upgrade` script that was previously used, which we did; however we learned that it had bitrotted and was broken since the previous Debian 11 (bullseye) -> Debian 12 migration was done along side a fresh install and was never actually upgraded to.
```
deeplow, line 11: Nit: link

```suggestion
We expected to be able to [obsolete the `securedrop-handle-upgrade` script](https://github.com/freedomofpress/securedrop-workstation/issues/1225) that was previously used, which we did; however we learned that it had bitrotted and was broken since the previous Debian 11 (bullseye) -> Debian 12 migration was done along side a fresh install and was never actually upgraded to.
```

architecture-decisions/004-trixie-updater-woes.md
```diff
@@ -0,0 +1,59 @@
+# Debian 13 upgrade's updater woes
```
deeplow, line 1: This is framed in terms of what caused it. Should we frame it in terms of the decisions instead? Maybe something like:

```suggestion
# Template upgrade process (revisited + workarounds)
```

(also the file name if you agree)

architecture-decisions/004-trixie-updater-woes.md
```diff
@@ -0,0 +1,59 @@
...
+We implemented the following long-term fixes in the following 1.9.0 release:
+
+1. [Only run the dom0 salt states](https://github.com/freedomofpress/securedrop-workstation/pull/1784) if there is no migration flag instead of unconditionally.
+2. n/a
+3. [Fetch the template names to update dynamically](https://github.com/freedomofpress/securedrop-workstation/pull/1793), so it should just work for the next Debian upgrade.
+4. n/a
+
+Additionally, we [switched](https://github.com/freedomofpress/securedrop-workstation/issues/1815) to using `%triggerun` instead of `%post` for dropping the migration flag to make it work for "skip" updates.
+
+We are in the process of adding an OpenQA scenario that performs the upgrade process, which likely would simplified identifying these bugs initially and then verifying the fixes.
```
deeplow, line 46: typo

```suggestion
We are in the process of adding an OpenQA scenario that performs the upgrade process, which likely would have simplified identifying these bugs initially and then verifying the fixes.
```

architecture-decisions/004-trixie-updater-woes.md
```diff
@@ -0,0 +1,59 @@
...
+
+During the Debian 12 (bookworm) -> Debian 13 (trixie) migration, we encountered a number of issues in the updater that necessitated various workarounds and then later, long-term solutions.
+
+## Context
+
+The Debian 13 migration had been delayed until the SecureDrop Inbox finished its rollout, so we were on a compressed timeline.
+
+We expected to be able to obsolete the `securedrop-handle-upgrade` script that was previously used, which we did; however we learned that it had bitrotted and was broken since the previous Debian 11 (bullseye) -> Debian 12 migration was done along side a fresh install and was never actually upgraded to.
+
+Part of the reason it broke was because as part of the Debian 12 migration, we stopped shipping an RPM containing a TemplateVM and instead switched to building it on the fly.
```
deeplow, line 13: (note) This reminds me that we should have an ADR for this switch, especially as there are some implicit reasons why it's needed, but those are not immediately obvious (other than bootstrapping ease).

Let me know if you agree and I'll file an issue for it. I know in theory we could retroactively file an issue for any past decision, but I think this one in particular has some nuances to clarify.

architecture-decisions/004-trixie-updater-woes.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Problems
+
+The largest issue was that the upgrade had to be fully compatible with the old version of the updater (1.7.1), because that's the version of the updater that will actually execute the Debian upgrade.
```
deeplow, line 17: It's kind of implied here but maybe we should make it explicit that another important factor was not to break in case of users jumping to post-1.8.0 releases, at least for a while. This was the motivation behind many of your suggestions. Maybe we can call this issue #5 and have [the fix for it (that you mention later)](https://github.com/freedomofpress/securedrop-workstation/pull/1848/changes#diff-be4231080bd9231f11eec965977dc59606854cee0449edee8982c5f3eb820f0dR44) map to it as well.

architecture-decisions/004-trixie-updater-woes.md
```diff
@@ -0,0 +1,59 @@
...
+4. We temporarily [set the `prohibit-start` feature](https://github.com/freedomofpress/securedrop-workstation/pull/1744) on all of our VMs to ensure they don't start up for any reason.
+
+One more issue that surfaced after 1.8.0 was released was that if the Debian upgrade fails mid-way (e.g. network hiccup), then it cannot be [resumed via the updater](https://github.com/freedomofpress/securedrop-workstation/issues/1796).
+
+We implemented the following long-term fixes in the following 1.9.0 release:
+
+1. [Only run the dom0 salt states](https://github.com/freedomofpress/securedrop-workstation/pull/1784) if there is no migration flag instead of unconditionally.
+2. n/a
+3. [Fetch the template names to update dynamically](https://github.com/freedomofpress/securedrop-workstation/pull/1793), so it should just work for the next Debian upgrade.
+4. n/a
```
deeplow, line 42: [nit] It was not obvious that this mapped to the problems and the short-term fixes were not separate lists. What do you think about turning it into a table?

| Problem | Short-term Fix (1.8.0) | Long-term Fix (1.9.0)|
|:--:|----|----|
|  **1** | In the RPM postinst, we [disabled the dom0 top state](https://github.com/freedomofpress/securedrop-workstation/pull/1769), so the updater's invocation of the dom0 state would be a no-op. Then because the migration flag was set, `sdw-admin --apply` would be executed, which re-enabled the top state and applied the upgrade. |[Only run the dom0 salt states](https://github.com/freedomofpress/securedrop-workstation/pull/1784) if there is no migration flag instead of unconditionally. |
| **2** | We temporarily [disable preloaded disposables](https://github.com/freedomofpress/securedrop-workstation/pull/1744) during the Debian upgrade. |  n/a|
| **3** | We [did not delete](https://github.com/freedomofpress/securedrop-workstation/commit/efa08077bf881d1894569b57b4934cc58e2325ea) the hardcoded Debian 12 templates, so the updater would still update them, even though they were no longer in use. |[Fetch the template names to update dynamically](https://github.com/freedomofpress/securedrop-workstation/pull/1793), so it should just work for the next Debian upgrade. |
| **4** | We temporarily [set the `prohibit-start` feature](https://github.com/freedomofpress/securedrop-workstation/pull/1744) on all of our VMs to ensure they don't start up for any reason. | n/a|

architecture-decisions/004-trixie-updater-woes.md
```diff
@@ -0,0 +1,59 @@
...
+### Problems
+
+The largest issue was that the upgrade had to be fully compatible with the old version of the updater (1.7.1), because that's the version of the updater that will actually execute the Debian upgrade.
+
+We encountered the following specific issues:
+
+1. The updater [runs dom0 salt states unconditionally](https://github.com/freedomofpress/securedrop-workstation/issues/1762) before the migration kicks in to run a full `sdw-admin --apply`. So either the upgrade needs to happen fully in the dom0 state or the dom0 state has to be compatible with both Debian 12 + Debian 13.
+2. Preloaded disposables interfered with switching the template for sd-viewer.
+3. The 1.7.1 updater [hardcoded the names of the Debian 12 templates](https://github.com/freedomofpress/securedrop-workstation/issues/1771) to update at the end, which meant they needed to be present on the system.
+4. `sd-log` would start up because it was receiving log entries from the newly created templates.
```
deeplow, line 24:
```suggestion
4. `sd-log` would start up because it was receiving log entries from the newly created templates and preloaded disposables
```

Preloaded disposables were a big annoyance here as well. Although this may break your issue mapping 😓. Also, the other day Marek mentioned something else that could have helped with the disposables and I filed an issue for us to try it: https://github.com/freedomofpress/securedrop-client/issues/3744.

architecture-decisions/004-trixie-updater-woes.md
deeplow, line 17: Also another note that we should make explicit, I think, is that the updater as the update-delivery must be resilient to most kind of failures. Otherwise we always need manual intervention or a fallback mechanism.

architecture-decisions/004-trixie-updater-woes.md
```diff
@@ -0,0 +1,59 @@
...
+
+The primary consequence was that the Debian 13 upgrade took longer to roll out and we did barely miss the official end-of-life deadline.
+
+If we didn't make any other changes to provision or upgrading logic, we are probably well set for the next Debian 13 -> Debian 14 upgrade in 2027.
+
+But we likely do want to overhaul the updater to make it easier for us to develop with and have [started brainstorming a redesign](https://github.com/freedomofpress/securedrop-workstation/issues/1766).
+
+## Alternatives considered
+
+* There is an open feature request for Qubes to support deferred template switches, so that you can change templates while the VM is running.
```
deeplow, line 58: (link + reason for not picking it)

```suggestion
* There is an open feature request for Qubes to support [deferred template switches](https://github.com/qubesos/qubes-issues/issues/8070), so that you can change templates while the VM is running. This would have only been available in the upcoming Qubes version.
```

architecture-decisions/004-trixie-updater-woes.md
```diff
@@ -0,0 +1,59 @@
...
+## Consequences
+
+The primary consequence was that the Debian 13 upgrade took longer to roll out and we did barely miss the official end-of-life deadline.
+
+If we didn't make any other changes to provision or upgrading logic, we are probably well set for the next Debian 13 -> Debian 14 upgrade in 2027.
+
+But we likely do want to overhaul the updater to make it easier for us to develop with and have [started brainstorming a redesign](https://github.com/freedomofpress/securedrop-workstation/issues/1766).
+
+## Alternatives considered
+
```
deeplow, line 57:
```suggestion
* For issue 4 we also considered the possibility of temporarily disabling `securedrop.Log`  as that was the reason reason `sd-log` was getting restarted mid-upgrade. This was not chosen due to the other solutions already solving it and the tight deadline. It could have been equally viable.
```

deeplow (changes requested):

architecture-decisions/004-trixie-updater-woes.md
legoktm: Yeah, "Template upgrade process" is a much better name.

architecture-decisions/004-template-upgrade-process.md
```diff
@@ -0,0 +1,59 @@
...
+
+During the Debian 12 (bookworm) -> Debian 13 (trixie) migration, we encountered a number of issues in the updater that necessitated various workarounds and then later, long-term solutions.
+
+## Context
+
+The Debian 13 migration had been delayed until the SecureDrop Inbox finished its rollout, so we were on a compressed timeline.
+
+We expected to be able to obsolete the `securedrop-handle-upgrade` script that was previously used, which we did; however we learned that it had bitrotted and was broken since the previous Debian 11 (bullseye) -> Debian 12 migration was done along side a fresh install and was never actually upgraded to.
+
+Part of the reason it broke was because as part of the Debian 12 migration, we stopped shipping an RPM containing a TemplateVM and instead switched to building it on the fly.
```
legoktm: > Let me know if you agree and I'll file an issue for it.

Yes please.

architecture-decisions/004-trixie-updater-woes.md
legoktm: Added.

legoktm: Good idea, done.

architecture-decisions/004-trixie-updater-woes.md
legoktm: I love this. Done!

architecture-decisions/004-template-upgrade-process.md
deeplow: Done! https://github.com/freedomofpress/securedrop-workstation/issues/1878

deeplow (approved):
Excellent. Thanks for the initiative on this one!

PR #1849, opened 2026-09-09
Use debian-13-minimal template for sd-admin base

### LINKED ISSUE #1788: Base sd-admin template on debian-13-minimal
Currently sd-admin's template is using debian-13 (GNOME), whereas the viewer/inbox templates use debian-13-minimal. (We don't need to be consistent for consistency's sake, just ensure this is an intentional divergence).

zenmonkeykstop:
Fixes #1788 

This PR:
- updates the RPM build process to build both admin and workstation RPMs (for convenience' and code deduplication's sake)
- switches the base `sd-admin-debian-13` template to be cloned from debian-13-minimal and installs a minimal set of addtional Qubes packages
- enables PVH mode with the securedrop-workstation grsec kernel for the `sd-admin` AppVM

(Note that the use of grsec kernels for the workstation is under review - unless and until a comparable alternative hardened kernel is available, it seems prudent to use it for securedrop-* VMs that handle untrusted data or make external network connections. sd-admin does both, hence why it's included here. The template does not have network access and debian packages are trusted, so it is acceptable to not enable the grsec kernel there.)

## Test plan
- checkout this branch in an `sd-dev` VM on a test workstation
- build RPMs from this branch in dom0 with `make clone`
- [x] verify that both admin and workstation RPMs were built.
- install with `sudo dnf install rpm-build/RPMS/noarch/securedrop-admin-dom0-config-<version>.fc41.noarch.rpm`
- [x] run `make sd-admin` to provision the `sd-admin-debian-13` template and the `sd-admin` appvm, and verify that it completes successfully
- [x] verify that sd-admin:
  - [x] boots successfully
  - [x] includes the securedrop-admin package and its dependencies
  - [x] includes the xfce4-terminal package
  - [x] is set to use PVH virtmode with the `pvgrub2-pvh` kernel option
  - [x] uses the latest securedrop workstation grsec kernel.

## Checklist

This change accounts for:


admin_salt/sd-admin-packages.sls
```diff
@@ -11,12 +11,25 @@
...
   cmd.run:
     - name: apt-get update --allow-releaseinfo-change
 
+# Install qubes VM kernel support here, activate PVH in the sd-admin AppVM later
+install-qubes-packages:
+  pkg.installed:
+    - pkgs:
+      - qubes-vm-recommended
+      - linux-image-amd64
+      - linux-headers-amd64
```
legoktm: Why do we need the headers? This will pull in GCC and a whole build toolchain.

admin_salt/sd-admin-packages.sls
```diff
@@ -11,12 +11,25 @@
...
 
 update-apt-cache:
   cmd.run:
     - name: apt-get update --allow-releaseinfo-change
 
+# Install qubes VM kernel support here, activate PVH in the sd-admin AppVM later
+install-qubes-packages:
+  pkg.installed:
+    - pkgs:
+      - qubes-vm-recommended
```
legoktm, line 25:
```suggestion
      # TODO: most of these should be installed via Debian package
      - qubes-vm-recommended
```

I think it's fine for now so we can keep moving but should be refactored later.

scripts/build-rpm.sh
```diff
@@ -17,11 +17,12 @@ trap 'find . -type f -exec touch -m {} \;' EXIT
 
 rpmbuild \
     --build-in-place \
+    --with admin \
```
legoktm: Let's drop requiring `--with admin` altogether?

admin_salt/sd-admin-packages.sls
```diff
@@ -11,12 +11,25 @@
 # independent of the workstation package.
 ##
 
+{% if grains['id'] != 'dom0' %}
```
legoktm: Why do we need this?

legoktm (changes requested):
Just did a quick skim

admin_salt/sd-admin-packages.sls
zenmonkeykstop: aye, probably reflex on my part from the bad old days of that custom qubes module whose name I forget.

admin_salt/sd-admin-packages.sls
zenmonkeykstop: holdover from another iteration of this where I was trying (failing) to install the packages via a nested qubesctl call and then enable grsec in the template. Can delete.

scripts/build-rpm.sh
zenmonkeykstop: If you're down with it, sure! I was keeping the ability to build the workstation rpm separately jsut in case there were scenarios I hadn't accounted for, but it's an easy change.

admin_salt/sd-admin-packages.sls
zenmonkeykstop: I kindof like having these explicit. Also we'd either need an fpf-hosted metapackage that basically just pulls this and other qubes- stuff in, or we'd need to add qubes-only dependencies to securedrop-admin. Maybe that's ok once Tails is out of the picture but now it feels like not so much.

legoktm: We have to create a `securedrop-admin-qubes` package anyways to install other packages too, like TBL. Unless it absolutely has to be in salt, it shouldn't be in salt.

scripts/build-rpm.sh
legoktm: Nah, we shouldn't need that anymore. Any workstation releases we need to make before admin tools go out should be point releases from the previous branch.

admin_salt/sd-admin-packages.sls
zenmonkeykstop: "Why not less Salt" will always win :)

legoktm (approved):
LGTM, provisioning went well. The copilot comment is on a stanza that's going to go away, so OK not fixing it now.

Noticed that this template also suffers from the locale issue when I opened xfce4-terminal, so we'll need to port https://github.com/freedomofpress/securedrop-client/pull/3535 over to securedrop-admin-qubes.

admin_salt/sd-admin-packages.sls
```diff
@@ -17,6 +17,17 @@ update-apt-cache:
   cmd.run:
     - name: apt-get update --allow-releaseinfo-change
 
+# Install qubes VM kernel support here, activate PVH in the sd-admin AppVM later
+# TODO: add an fpf metapackage to manage these dependencies
+install-qubes-packages:
+  pkg.installed:
+    - pkgs:
+      - qubes-vm-recommended
```
legoktm, line 25: as part of moving these to the metapackage:

* don't think we need linux-image-amd64 because that's the distro kernel (probably leftover)
* grsec metapackage already pre-depends qubes-kernel-vm-support so we don't need to specify it here
* I would've expected that grub2 would've been installed by the metapackage, but it isn't. The other templates don't use it, and when I remove it everything is fine, so let's drop it.

zenmonkeykstop: Confirmed, the linux-image package was needed to support PVH in the absence of grsec, but now grsec is back in the picture we don't need it.

admin_salt/sd-admin-packages.sls
deeplow: I'm late for the review, but do we need `linux-image` if we also install the kernel? My understanding is that this is only needed if we expect to boot without grsec, but `phvgrub2-pvh` as the kernel.

PR #1852, opened 2026-09-10
Install packages via securedrop-admin-qubes metapackage

legoktm:
This package pulls in what we need in the template. Depends on https://github.com/freedomofpress/securedrop/pull/7909 which creates the new package.

## Test plan

* [ ] fresh provision of sd-admin works + boots the template
## Checklist

This change accounts for:

### DIFF
--- a/admin_salt/sd-admin-packages.sls
+++ b/admin_salt/sd-admin-packages.sls
@@ -17,17 +17,6 @@ update-apt-cache:
   cmd.run:
     - name: apt-get update --allow-releaseinfo-change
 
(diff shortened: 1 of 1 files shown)


legoktm:
Marking as ready for review since the other PR was merged, but I don't think provisioning will work until we have nightlies first: https://github.com/freedomofpress/securedrop/pull/7914.

zenmonkeykstop (approved):
Approved - one thing that this does introduce is the need for at least a partial/hotfix server release before we push this. But that's likely OK.

PR #1857, opened 2026-09-11
Update all templates with `sd-updater` tag

### LINKED ISSUE #1598: Include `sd-extras` tag in updates
## Description

Include qubes/templates with the `sd-extras` tag in the updates applied by the SDW preflight updater.

In some cases newsrooms have been advised in how to set up additional software on the SecureDrop Workstation. We advise creating a new template with a distinct name, installing additional software in that template, and then creating appropriate application qubes or disposables based on that template. We also advise on setting up RPC policies for these new qubes. A [qubes vm tag](https://dev.qubes-os.org/projects/core-admin-client/en/latest/manpages/qvm-tags.html) is typically applied to these new qubes and used to help manage these RPC policies (e.g. `sd-extras`). 

In a normal SDW installation, the Preflight updater takes care of all updates. It is not necessary for users to run the [Qubes Update tool](https://doc.qubes-os.org/en/latest/user/how-to-guides/how-to-update.html) and they are not instructed to do so after the initial installation. 

We should include a tag that we advise is applied to additional custom software templates and qubes in what's updated by the SDW updater so that regular updates are applied without relying on additional manual runs of the Qubes Update tool. This will help ensure extra software is kept up-to-date without creating an additional maintenance burden for admins or users. 

@deeplow: Currently template update failures result in a failed update process — this preventing the use of the workstation. As we expand the updater use-cases, we may want to consider scenarios where an external template (tagged with `sd-extras`) purposely (or accidentally) fails to cause a denial of service. We may want a more fine-grained best-effort update approach to those.

### LINKED ISSUE #1816: Conditionally update the sd-admin template with the Workstation updater
SDW's preflight updater updates a hardcoded set of templates required by the journalist workflow. This doesn't include `sd-admin-debian-13`. Admins can update `sd-admin-debian-13` manually with the Qubes updater, but will they?

Maybe we should update the preflight updater to also update `sd-admin-debian-13` when it exists on the system. When it does not exist, skip it silently. The journalist-only update workflow will remain unchanged.

If `sd-admin` exists but its expected template is missing, report an actionable configuration error rather than silently treating the admin environment as updated.

## Acceptance criteria

- `sd-admin-debian-13` is updated when present.
- A missing `sd-admin-debian-13` does not cause updates to fail on journalist-only Workstations.
- Running the updater never provisions `sd-admin` as a side effect.
- An inconsistent configuration (`sd-admin` present without its expected template) is reported clearly.
- Automated tests cover admin-enabled, journalist-only, and inconsistent configurations.

legoktm:
This tag is applied to sd-inbox-debian-13, sd-viewer-debian-13, sd-admin-debian-13 and can also be applied to any user-created templates that they want the updater to, well, update.

It is not applied to fedora-XX-xfce templates.

Refs #1158.
Fixes #1598.
Fixes #1816.

## Test plan

* [ ] create a new template; tag it with sd-updater but not sd-workstation and then run the updater and see it gets updated
## Checklist

This change accounts for:

### DIFF
--- a/admin_salt/sd-admin-template.sls
+++ b/admin_salt/sd-admin-template.sls
@@ -20,6 +20,6 @@ sd-admin-debian-13:
         - sd-workstation
         - sd-workstation-trixie
         - sd-admin
+        - sd-updater
     - require:
       - qvm: dom0-install-debian-13-template
-
--- a/launcher/tests/conftest.py
+++ b/launcher/tests/conftest.py
@@ -5,6 +5,7 @@ import tempfile
 import pytest
 
 import sdw_updater.Updater
+from sdw_updater.Updater import SD_UPDATER_TAG
 
 
 @pytest.fixture
@@ -60,8 +61,8 @@ SD_TAG = "sd-workstation"
 MOCK_FEDORA_TEMPLATE = "fedora-XX-xfce"
 
 # SecureDrop-managed TemplateVMs, all tagged `sd-workstation`.
+MOCK_SDW_BASE_TEMPLATE = "sd-base-debian-XX"
 MOCK_SDW_TEMPLATES = [
-    "sd-base-debian-XX",
     "sd-inbox-debian-XX",
     "sd-viewer-debian-XX",
 ]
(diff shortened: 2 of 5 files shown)


zenmonkeykstop (approved):
LGTM, test plan checks out.

Issue #1791, opened 2026-07-21
sd-admin: Figure out how updater works with separate RPM


legoktm:
Currently the plan is to ship admin tools in a separate RPM, independent from our existing securedrop-workstation-dom0-config. The updater ships in the latter, so we need to figure out how a standalone sd-admin-only workstation will get updates. And how a joint sd-admin + journalist workstation would work too.

deeplow:
Technically as it stands, the admin VM [is being tagged](https://github.com/freedomofpress/securedrop-workstation/blob/bfd57d8/admin_salt/sd-admin.sls#L28C1-L29C19) with `sd-admin` as well as `sd-workstation`. So technically with the new proposed [`_get_current_templates()`](https://github.com/freedomofpress/securedrop-workstation/pull/1793/changes#diff-c49796294389d55ca4beb4396a52db9fbaee39b87fbd82f3d3ad2b503e8da417L60-R81) would pick it up, since it uses that tag. This means that the joint scenario should be solved.

Why not simply updater as a separate package, required by both. This means three RPMs, but they have clearly distinct purposes. With https://github.com/freedomofpress/securedrop-workstation/pull/1793/ we're already reducing the hardcoded dependencies and ideally we have a "plugable" approach where more MultiVM applications (or even [individual VMs](https://github.com/freedomofpress/securedrop-workstation/issues/1598)) could be added without needing to change anything in the updater itself.

For the time being tags are a "good enough" way to signal to the updater that these are under its management. Long-term that's a wider discussion.

zenmonkeykstop:
Yup, updater-as-a-separate package is my thinking as well. Seems like it's easy enough to split out but will give it a try and see if there are any gotchas.

legoktm:
I think I'd like to see a bit more of a plan as to how the updater is going to interact with the admin RPM before we refactor it all. (And for the other updater patches to land first)

1) Are we going to implement the same "only run dom0 states unless there's a migration flag" system?
2) Will we have an equivalent `sdw-admin` command for admin tools? Or should `sdw-admin` learn how to deal with both?
3) How will the updater detect admin tools / journalist tools are installed?
4) Currently the updater is a launcher for the Inbox. What does it launch in admin tools? If this is going to intersect w/ #1805 then does the updater code need to support messages for all of those?

deeplow:
[...]

PR #1863, opened 2026-09-15
Add empty securedrop-dom0-manager package

legoktm:
This will contain the new securedrop-manage command (rename of sdw-admin), plus the updater and launcher. For now it just has a README/LICENSE file with dependencies set; we'll start moving over files into it iteratively.

Refs #1791.

----

Name came from Kev's "securedrop-dom0-launcher" name, which I liked, but I used "manager" instead since it also encompasses the updater, and we're naming the CLI "securedrop-manage". Written by hand but pre-reviewed by Claude.

## Test plan

* [x] `make build-rpm` builds the new manager package too
* [ ] try `make dev`; see that it just works, installing both manager + workstation packages
* [x] `sdw-admin --uninstall` removes the manager + workstation packages
## Checklist

This change accounts for:


files/sdw-admin.py
```diff
@@ -387,7 +387,15 @@ def perform_uninstall() -> None:
...
         subprocess.check_call(
-            ["sudo", "dnf", "-y", "-q", "remove", "securedrop-workstation-dom0-config"]
+            [
+                "sudo",
+                "dnf",
+                "-y",
+                "-q",
+                "remove",
+                "securedrop-workstation-dom0-config",
+                "securedrop-dom0-manager",
```
zenmonkeykstop: This also removes securedrop-admin-dom0-config, which is probably not always what we want. Removing securedrop-workstation-dom0-config on its own should also remove securedrop-dom0-manager if no other packages depend on it.

zenmonkeykstop (changes requested):
Erring on the side of conservatism when removing packages unless/until admin stuff is also managed via the same script seems sensible.

files/sdw-admin.py
legoktm: Oh hmmmm.

legoktm: In `make dev` we manually install securedrop-dom0-manager, so it doesn't autoremove when nothing depends on it. So we need an `if not admin: remove.extend(securedrop-dom0-manager)` I think.

legoktm: Implemented it, but it relies on https://github.com/freedomofpress/securedrop-workstation/pull/1866.

zenmonkeykstop: I can probably just grab that and review/merge to move this along.

zenmonkeykstop: (tho that feels like a bit of a hack, IMO we should be building for the prod case and not trying to accommodate dodgy dev modes. The compromise might be to dnf remove the workstation package and then do an explicit dnf autoremove right after)

legoktm: `--uninstall` is dev-only. `dnf autoremove` won't do anything because the manager package is explicitly installed. There's probably ways to work around it, e.g. we can manually adjust the "explicitly installed" flag.

But I'd rather revisit that once we've changed to securedrop-manage since I think it'll look a bit different then.

legoktm: > e.g. we can manually adjust the "explicitly installed" flag.

OK, `sudo dnf mark dependency securedrop-dom0-manager` is the magic invocation, and now dev installs will match prod installs better :)

legoktm:
I also tacked on a commit to ensure we're using matching admin/journalist + manager versions and don't accidentally get out sync

scripts/prep-dev
```diff
@@ -6,42 +6,35 @@ set -euo pipefail
...
-# used to differentiate between the the default RPM and the WIP sd-admin RPM.
-rpm_name="securedrop-workstation-dom0-config"
+RPM_NAME="${RPM_NAME:-securedrop-workstation-dom0-config}"
 
 function find_latest_rpm() {
     # Look up which version of dom0 we're using.
     fedora_version="$(rpm --eval '%{fedora}')"
-    find "${DOM0_DEV_DIR}/rpm-build/RPMS/" -type f -iname "${rpm_name}*fc${fedora_version}.noarch.rpm" -print0 | xargs -r -0 ls -t | head -n 1
+    # TODO: Use proper RPM parsing / sorting to handle this
+    find "${DOM0_DEV_DIR}/rpm-build/RPMS/" -type f -iname "${1}*fc${fedora_version}.noarch.rpm" -print0 | xargs -r -0 ls -t | head -n 1
```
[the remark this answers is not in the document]
legoktm: We don't need to support that anymore.

zenmonkeykstop (approved):
🟢 **Merging recommended**

LGTM, checklist checks out.

Issue #1528, opened 2026-01-05
Drop support for non-disposable `sys-usb`


deeplow:
## Description

Originally `sys-usb` was persistent. However, in Qubes 4.1 it became disposable. During a transition period both options needed to be accounted for (in-place upgrade systems to Qubes 4.1 had sys-usb persistent) — see https://github.com/freedomofpress/securedrop-workstation/pull/764/. However, with the [1.0.0 release](https://securedrop.org/news/securedrop-workstation-1_0_0-released/) a clean 4.2 was required, which means **sys-usb should always be disposable**.

The exception is if the user toggled [the initial Qubes setup option](https://github.com/QubesOS/qubes-anaconda-addon/blob/release4.3/org_qubes_os_initial_setup/tui/spokes/qubes_os.py#L135) to make sys-usb and sys-firewall persistent. If so, this is something we should actively detect and warn admins about.

**Successful implementation includes:**
- [ ] Validate `sys-usb` is disposable at runtime (e.g. on first install or `sdw-admin --apply`)
- [ ] Removing [test case](https://github.com/freedomofpress/securedrop-workstation/pull/764/files#diff-3aca1914d43f77992facb2f4db8103ad4b034c44508c0c66e70ff49bd646717cR19)

> **Note:** We may want to implement this change as in the broader category of system health checks.

## How will this impact [SecureDrop/SecureDrop Workstation users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?

In case sys-usb was made persistent, admins will be informed that this is an unsupported configuration.

## How would this affect the SecureDrop Workstation [threat model](https://github.com/freedomofpress/securedrop-workstation/#threat-model)?

Should not affect.

PR #1868, opened 2026-09-17
Duplicate salt logic for admin workstation to manage sys-* VMs

legoktm:
On a workstation that is admin-only, we still want to manage the sys-* VMs to ensure they're using the latest Fedora template, which we update. Use a nearly identical copy of the existing salt logic, except we don't need to enable the auto-attach handling, nor worry if sys-usb is disposable. (Once #1528 is done the delta between the two would be pretty small and we could just have a test ensure the two are in sync.)

On a combined workstation, have the admin_salt code be a no-op and just let the existing journalist salt code handle it.

Fixes #

## Checklist

This change accounts for:


legoktm:
Per https://github.com/freedomofpress/securedrop-workstation/issues/1881#issuecomment-5799259786.

PR #1869, opened 2026-09-17
Remove obsolete sdlog.conf

legoktm:
This stopped being provisioned in c81c29b (2020) but we missed cleaning it up.

## Test plan

* [ ] grep; find that nothing in `securedrop_salt/` references it (test references are to the copy in securedrop-client)
## Checklist

This change accounts for:


nathandyer (changes requested):
I see three references to sdlog.conf in tests, do those need to be changed as well?

```
$ grep -rnw '/home/nathandyer/Projects/securedrop-workstation/' -e 'sdlog.conf'
grep: /home/nathandyer/Projects/securedrop-workstation/.git/index: binary file matches
/home/nathandyer/Projects/securedrop-workstation/tests/base.py:232:        assert self.fileExists("/etc/rsyslog.d/sdlog.conf")
/home/nathandyer/Projects/securedrop-workstation/tests/test_vm_sd_gpg.py:118:    assert not qube.fileExists("/etc/rsyslog.d/sdlog.conf")
/home/nathandyer/Projects/securedrop-workstation/tests/test_vm_sd_log.py:50:    assert not qube.fileExists("/etc/rsyslog.d/sdlog.conf")
```

legoktm:
> I see three references to sdlog.conf in tests, do those need to be changed as well?

No, those are checking for https://github.com/freedomofpress/securedrop-client/blob/main/log/log_client_files/sdlog.conf - I should've been more precise in my test plan, to grep that nothing in `securedrop_salt/` uses it.

nathandyer (approved):
> No, those are checking for https://github.com/freedomofpress/securedrop-client/blob/main/log/log_client_files/sdlog.conf - I should've been more precise in my test plan, to grep that nothing in `securedrop_salt/` uses it.

Got it! In that case, we're all clear :)

PR #1870, opened 2026-09-17
Fold destroy-vm.py into sdw-admin

legoktm:
In preparation for moving it to the manager package.

The make target was broken at the time it was introduced because the script was in `./files` not `./scripts`.

----

Had Claude do the initial merging and then I cleaned it up.

## Test plan

* [ ] `sdw-admin --uninstall` still works as expected
## Checklist

This change accounts for:


deeplow (approved):
LGTM. I have not visited this script in a while and I can see a few problems, but they are not introduced by this PR, so they are not relevant and this seems to work.

I wonder if there should exist some upstream qube graph functions, since there are so many qube graph traversal code in the Qubes codebase.

legoktm:
> I can see a few problems, but they are not introduced by this PR, so they are not relevant and this seems to work.

Please file and I can at least take a look since all this context is fresh in my mind right now, but yeah, the goal for this was to just get rid of the scripts/ directory :)

PR #1873, opened 2026-09-18
Remove clean-salt; have RPM take over functionality

### LINKED ISSUE #1867: clean-salt is unnecessary
The main part of the script is:
```
# If SDW Salt config dir already exists, delete all SecureDrop Workstation
# related Salt files. In production scenarios, most of these will be provisioned
# by the RPM package, but the top files and configs will not, so we should use a
# common script to ensure all config is removed.

if [[ ! -d "$SDW_SALT_DIR" ]]; then
    sudo rm -rf ${SDW_SALT_DIR}

    # Can be removed in future
    sudo rm -rf ${SALT_DIR}/launcher

    sudo find ${SALT_DIR}/_tops -lname '/srv/salt/securedrop_salt*' -delete

fi
```

But the condition is `! -d`, which means it only runs when the directory *does not* exist, which it always does, so nothing ever happens.

The launcher folder no longer exists, so that's fine. We are leaving the /srv/salt/securedrop_salt directory behind, but I think we should handle that removal via RPM instead of relying on this.

legoktm:
./files/clean-salt had 3 tasks:
    
1) Cleaning /srv/salt/securedrop_salt. We now have RPM remove that directory by correctly marking the files we copy in as %ghost, which gives RPM permission to delete it and the whole directory on uninstall.

2) Delete the launcher directory, which no longer exists.

3) Disable the top file, which we now do in the RPM %preun.
    
Fixes #1867.

----

Claude did the writing after I figured out what the correct solution was.

## Test plan

* [ ] `make dev`, and then `sdw-admin --uninstall`, verify that `/srv/salt/securedrop_salt` is entirely gone

## Checklist

This change accounts for:


rpm-build/SPECS/securedrop-workstation-dom0-config.spec
```diff
@@ -221,6 +224,7 @@ systemctl enable securedrop-logind-override-disable.service ||:
 %preun
 # If we're uninstalling (vs upgrading)
 if [ $1 -eq 0 ]; then
+    qubesctl top.disable securedrop_salt.sd-workstation ||:
     %systemd_preun securedrop-logind-override-disable.service
     %systemd_user_preun securedrop-user-xfce-icon-size.service
     %systemd_user_preun securedrop-user-xfce-settings.service
```
deeplow, line 230: Out of scope for this issue but I wonder if we should move this into the journalist workstation package.

deeplow (approved):
LGTM. I'd like to add some tests to this just so we're not blindly validating uninstallation exits 0.

